Issue #337, opened 2018-02-26
new parameter file release


rgknox:
To help people (and me) plan, we will be issuing a new default parameter file in the near future.  Here is a list of potential changes.  Please comment below, and we can iterate, changing this from potential to actual.

- Now that we have freeze mortality, change freeze tolerance threshold from a dummy val to some nominal value for all PFTs (currently its 1000C, which kills all plants)

- Expand PFTs to include work-in-progress extra-tropical PFTs (@jenniferholm #281 )

- Issue #303, change fates_comp_excln = 3

- Set branch_turnover to non-zero values - setting it to 50 years as default

- Add a field to option storage allometry mode  "fates_allom_stmode"

- logging_dbhmax_infra  (35cm)

- clumping_index = 0.85

- Change "fates_clone_alloc" param name to be "fates_seed_alloc_mature" (#278)

- Check with @lmkueppers and @adamhb to see what new parameters could be needed for recruitment

- Insect parameters? @xuchongang, do you guys have know the set of parameters you want for the insect/pest module?

Tagging @serbinsh and @mdietze, so you know what is coming down the pipe

Issue #153, opened 2016-12-01
Question: Shoot/canopy clumping in ed-clm RT scheme?


serbinsh:
### Summary of Issue:
Question - does the current scheme contain a [canopy] clumping (factor) parameter definable by PFT?  I see many other params such are orientation factor, refl/trans of leaf and stem but not clumping.

### Expected behavior and actual behavior:
Expected a clumping factor parameter

### Steps to reproduce the problem (should include create_newcase or create_test command along with any user_nl or xml changes):
Searched in the clm_params_ed.c160808.nc param file

### What is the changeset ID of the code, and the machine you are using:
NA

### have you modified the code? If so, it must be committed and available for testing:
No

### Screen output or output files showing the error message and context:
No

PR #349, opened 2018-03-08
Some radiation profile updates: clumping, SAI changes, and diagnostics

ckoven:
A first set of changes to the radiation calculations:  include clumping index term, output profiles of PAR, output LAI and SAI across plant size classes, redefine SAI as a function of target leaf biomass rather than stem biomass.

### Description:
Added canopy clumping term that affects the direct and diffuse extinction coefficients
Redefined SAI in terms of the target leaf biomass, because the old bdead scaling meant that SAI for large trees got really large. 
Added diagnostic variables of direct and diffuse PAR profiles across canopy and leaf layers, for comparison to obs
Added diagnostic variables of LAI and SAI as function of size classes.
Also I moved the functions that calculate SAI and LAI to the allometry module because they make more sense there than in the cohort dynamics module where they were before.  That required disaggregating the argument to those functions from the cohort type to its various relevant properties.

### Collaborators:
Discussed with @rgknox, @rosiealice, @serbinsh, others

Yes.  All of these changes should have the effect of increasing light to the understory.

This requires a change to the parameter file, both the clumping index which is already updated and the change to the values and metadata of SAI_scaler (will add that to that thread).

### Checklist:

- [ ] I have updated the in-code documentation AND wiki accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/rgknox/fates/blob/rgknox-new-PR-template/CONTRIBUTING.md) document.
- [] If answers were expected to change, evaluation was performed and provided

### Test Results:
None yet, will test after comment period.  Will post figures of how the light profile changes due to these changes in the discussion of this PR once I make those figures.

FATES-CLM (or) HLM test hash-tag:

FATES-CLM (or) HLM baseline hash-tag:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -148,20 +148,6 @@ contains
          enddo
 
 
-         ! ------------------------------------------------------------------------------
-         ! Check patch area to prevent numerical weirdness
-         ! ------------------------------------------------------------------------------
-         
-         if (currentPatch%area .lt. min_patch_area) then  
-            
-            write(fates_log(),*) 'An incredibly small patch exists that should'
-            write(fates_log(),*) 'had been fused or culled already'
-            write(fates_log(),*) 'currentPatch%area: ',currentPatch%area
-            write(fates_log(),*) 'min_patch_area: ',min_patch_area
-            call endrun(msg=errMsg(sourcefile, __LINE__))
-            
-         end if
-
          ! Does any layer have excess area in it? Keep going until it does not...
          patch_area_counter = 0
          area_not_balanced = .true.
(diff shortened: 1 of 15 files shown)


jenniferholm:
@ckoven - I might have missed this from last year, but is there a reference backing up why we added the canopy clumping index, and if possible why we choose 0.85 as the value? Was this just through different testing and tuning?

I'm revising a manuscript and I would like to include a reference for why we decided to include this clumping index so that it could now affects the direct and diffuse extinction coefficients. Is it from theory from Norman 1979? Or something else in the radiation transfer scheme in ED?

Thanks!

ckoven:
@jenniferholm discussion on the clumping index was here: https://github.com/NGEET/fates/issues/153

and it looks like we chose an initial guess value of 0.85 for the clumping here: https://github.com/NGEET/fates/issues/337#issuecomment-371639001

if you need to revert to the old behavior, setting the clumping index to 1.0 would do that.

PR #414, opened 2018-08-30
Updates to leaf and sapwood allometries

rgknox:
### Description:

1) The power function allometries describing target leaf biomass were given an intercept parameter which should help us match data on saplings moving forward.  

The intercept is now: fates_allom_d2bl1
The slope is now: fates_allom_d2bl2
The exponent is now: fates_allom_d2bl3

Note: The wood density dependent formulation for fates_allom_lmode = 1, is not impacted.

No new parameters for leaf biomass allometry were necessary, however the default values in the file were changed, such that the third option is used (power-function with height capping), such that results match the Saldarriaga method used previously.

2) The allometry describing target sapwood biomass was updated in a few ways.  Firstly, the parameter latosa_int and latosa_slp, were re-named and re-defined to be less ambiguous and more representative of field measurements.  The new parameters fates_allom_la_per_sa_int, and fates_allom_la_per_sa_slp, which should be read "leaf-area per sapwood-area intercept" and "leaf-area per sapwood-area slope".  These renamed parameters are also the inverse of their previous definitions, and now have units of "m2 / cm2".

@bchristo pointed out that the la_per_sa_slp parameter operates on height, and the function has been changed to accomodate that.  Further on this point, the derivative for sapwood biomass, now this functional relationship between height and leaf-area-to-sapwood-area in the derivative as well.

To support hydraulics code, a third output value is now generated by all calls to sapwood allometry, the cross sectional area of sapwood.  

### Collaborators:

@bchristo 
@ckoven 
@jkshuman   (discussions prompted feature 1)

It is expected that these changes should generate nearly the same results given changes in the new parameters, however round-off errors will prevent B4B results compared to master.

### Checklist:

- [x] I have updated the in-code documentation AND wiki accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/rgknox/fates/blob/rgknox-new-PR-template/CONTRIBUTING.md) document.
- [] If answers were expected to change, evaluation was performed and provided

### Test Results:

CTSM Hash:

df9b02d

Test Output:

acre sanity checks suggest that new defaults generate similar results as master

cheyenne, fates regression suite, (no compare) All PASS


ckoven:
I've been chatting abou tthis PR with @rgknox a bit.  I have concerns with the implications of adding an intercept term to the leaf biomass because of how it will interact with other aspects of the model.  In particular the crown area allometry.  Right now, the LAI of a given tree is basically proportional to the SLA times the ratio of the bleaf to the crown area.  The variable SLA that @kovenock has added complicates this slightly but overall the scaling ought to hold.  But if we add an intercept to the biomass allometry but not the crown area allometry, then for very small diameter trees, the LAI will be very large.  The variable SLA in this case makes the problem worse because the mean SLA will increase with LAI of the tree.  So that logic says that if we add a leaf biomass allometry intercept term, then we kind of need to add an intercept to the crown area allometry too.  But this then requires us to figure out what the crown area intercept ought to be.  Its not obvious what the right answer should be — we could hard-code a value that preserves the LAI scaling of the power law for the case where diameter equals zero, but that only works for the cases when d = 0 (and thus bleaf = bleaf_intercept) or bleaf >> bleaf_intercept, not for intermediate values.  We could leave it user-defined, but that might just give us more opportunities to do the wrong thing.  so if anyone has thoughts on how to handle this, please feel free to add here.

rosiealice:
I agree with all of that, seems like it could generate lots of small-tree strangeness. Presumably, the point was to better match up with datasets and/or add flexibility, but I guess the broader point is that one or both of the leaf area and crown diameter allometries don't work for small trees then were going to get into trouble with it?

rosiealice:
I've always quite liked the idea of a leaf-volume density approach.

rgknox:
I think I will remove the intercept parameter stuff from this PR and re-test.  I will open a new issue on the interplay between leaf biomass and crown area.

rgknox:
issueing a new PR as well, this PR might have a stale handle on the branch with the PR (since we went public and the forks disassociated).

PR #429, opened 2018-10-18
Recruitment into size classes

ckoven:
A thing that's been bugging me for a while is our inability to compare prognostic recruitment rates with census data of that, because we don't track the flux of trees into a given size class, only the recruitment into the smallest size class.  I had thought that this could be calculated during postprocessing by continuity, since we have mortality rates from each size class, the change in the population in each size class, and the recruitment into the smallest size class.  Unfortunately, doing that doesn't seem to work in general, I think because there are errors in population conservation within a given size class due to cohort fusion, which are sufficiently large that they mess up the calculation.

So this PR adds a couple variables to actually track this quantity.  I've added two new variables, both on a size-class x PFT dimension, which track the recruitment rates into a given size class, due to (a) normal growth, and (b) cohort fusion.  The real growth-based recruitment flux is positive definite. The virtual recruitment due to fusion can be either positive or negative depending on whether the trees are getting moved into a size bin or out of a size bin when they are fused.

### Collaborators:
Some discussion on this with @rgknox 

Expectation is that these quantities are purely diagnostic and should not change answers of existing variables

### Checklist:

- [ x I have read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/master/CONTRIBUTING.md) document.

### Test Results:
Not yet tested

27f7fa

FATES baseline hash-tag:

Test Output:

cheyenne, intel, fates test suite, alll PASS

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,14 +11,11 @@ For more information on the FATES model, see our wiki:  https://github.com/NGEET
 
 **Most users should not need to directly clone this repository.  FATES needs to be run through a host model, and all supported host-models are in charge of cloning and loading the fates software.**
 
-FATES has support to be run via the Accelerated Climate Model for Energy (ACME) and the Community Earth System Model (CESM).
+FATES has support to be run via the Accelerated Climate Model for Energy (ACME) and the Community Terrestrial Systems Model (CTSM).
 
 https://climatemodeling.science.energy.gov/projects/accelerated-climate-modeling-energy
 
-http://www.cesm.ucar.edu/
+https://github.com/ESCOMP/ctsm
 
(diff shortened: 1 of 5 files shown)


rgknox:
@ckoven , is there a way now that these diagnostics have been added, to test output continuity? ie can we use the growth flux diagnostic, and a mortality flux diagnostic to see if it matches the changes population diagnostics?

rgknox:
@ckoven, it looks like currentSite%growthflux_fusion  is zero'd for every history update, so i don't think we need to remember that array in the restart.  However, I think we do need to add the currentCohort%size_class_lasttimestep to the restart.  Does this sound right to you?

Issue #412, opened 2018-08-14
bug in fire model related to grass leaf fire consumption


ckoven:
All,
In setting up regional FATES runs over California, I have been running into a strange bug where the model would crash when I turned on the spitfire.  The crash showed up as a NaN in the NPP term, as below:

```
42: Plant Growth Integrator could not find
42: a solution in less than          300  tries
42: Aborting
42: carbon_balance                     NaN
42: deltaC                     NaN
42: totalC                     NaN
42: leaf: T                     NaN  7.091519147632314E-002                     NaN
42: froot: T                     NaN  7.091519147632314E-002
42:                     NaN
42: sap: T                     NaN  3.530199739476804E-004  3.530470578426043E-004
42: store: T                     NaN  8.509823315308761E-002
42:                     NaN
42: dead:                     NaN  2.345192039175753E-002  2.345436637946122E-002
42: ----------------------------------------
42:  Dumping Cohort Information             
42: ----------------------------------------
42: co%pft                    =            3
42: co%n                      =    2.71210026657635     
42: co%dbh                    =    1.35295799649237     
42: co%hite                   =    2.84460496513949     
42: co%bdead                  =   2.345436637946122E-002
42: co%bstore                 =                      NaN
42: co%laimemory              =   0.000000000000000E+000
42: co%bsw                    =   3.530470578426043E-004
42: co%bl                     =                      NaN
42: co%br                     =                      NaN
42: co%lai                    =   0.000000000000000E+000
42: co%sai                    =   8.692855374800612E-003
42: co%g_sb_laweight          =   0.000000000000000E+000
42: co%leaf_cost              =   0.000000000000000E+000
42: co%canopy_layer           =            1
42: co%canopy_layer_yesterday =    1.00000000000000     
42: co%nv                     =            1
42: co%status_coh             =            2
42: co%canopy_trim            =   0.832085051966768     
42: co%excl_weight            =    4.39088230981449     
42: co%prom_weight            =   1.074470174389356E-010
42: co%size_class             =            1
42: co%size_by_pft_class      =           27
42: co%gpp_acc_hold           =                      NaN
42: co%gpp_acc                =                      NaN
42: co%gpp_tstep              =                      NaN
42: co%npp_acc_hold           =                      NaN
42: co%npp_tstep              =                      NaN
42: co%npp_acc                =                      NaN
42: co%resp_tstep             =                      NaN
42: co%resp_acc               =                      NaN
42: co%resp_acc_hold          =                      NaN
[...]

Issue #408, opened 2018-07-26
Small typo in EDSurfaceAlbedoMod.F90


jenniferholm:
When double checking the albedo output to the HLMs, I found that there was a small typo for the "albi"  parameter. This is the diffuse surface albedo. 

At line 199, in EDSurfaceAlbedoMod.F90, the 'albd_parb' should be 'albi_parb'. 
This section of the code is setting the initial direct and diffuse albedo when there are no leaf layers in the patch. This will matter at the beginning of bare ground simulations, but could also effect albedo during recovery from clear-cuts, or other potential big disturbances. 

line 198: bc_out(s)%albd_parb(ifp,ib) = bc_in(s)%albgr_dir_rb(ib)
line 199: bc_out(s)%albd_parb(ifp,ib) = bc_in(s)%albgr_dif_rb(ib)

Fix should be minor and quick. :)

rgknox:
See:

https://github.com/NGEET/fates/blob/master/biogeophys/EDSurfaceAlbedoMod.F90#L130

I think we are all set here

PR #431, opened 2018-10-18
Fire mortality diagnostics update

jkshuman:
### Description:
Included in this PR is a fix to issue #408 correcting a typo to set the diffuse surface albedo as identified by J. Holm within EDSurfaceAlbedoMod.
These changes emerged from issue #412 . @ckoven updated the fire mortality diagnostics calculations. Specifically, the rates of fire mortality, rate of cambial mortality (from fire damage) and rate of crown mortality (due to fire damage) are tracked for diagnostics purposes. Site-level fire mortality is further stratified by canopy level. J.K. Shuman reviewed the changes by C. Koven, updated a typo to correct the association between cambial mortality and crown mortality, and removed the hard coding of canopy layers for the site-level stratification. 
This PR was updated to include the changes described in issue #436 by removing the calculated value alpha_FMC from the parameter file and adding the parameter drying ratio to the parameter file. alpha_FMC is calculated using drying_ratio and SAV (both from the parameter file). This update removes the possibility of having SAV and alpha_FMC values that are not related due to user error (forgetting to update both) by placing the calculation within the fuel moisture subroutine of SPITFIRE.

### Collaborators:
JKShuman, CKoven, RKnox, RFisher, JHolm

Answer changes are expected, specifically due to update of diffuse albedo.

### Checklist:

- [x ] I have read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/master/CONTRIBUTING.md) document.
- [x ] FATES PASS/FAIL regression tests were run

### Test Results:
This version compiles and runs on Cheyenne, but fails with a carbon balance error for the file set that I have in year 1 month 4. It failed the basegen tests on Cheyenne. Failure of basegen tests was discussed with @rgknox 

FATES baseline hash-tag:


main/FatesRestartInterfaceMod.F90
```diff
@@ -107,7 +107,6 @@ module FatesRestartInterfaceMod
   integer, private :: ir_bmort_co
   integer, private :: ir_hmort_co
   integer, private :: ir_cmort_co
-  integer, private :: ir_fmort_co
```
rgknox: I'm wondering if we need to add the new fire mortality arrays to the restarts... I suppose we will see if they were necessary after we run the tests.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -2108,6 +2110,36 @@ subroutine update_history_dyn(this,nc,nsites,sites)
...
+               ! fire mortality from the site-level diagnostic rates
+               hio_m5_si_scpf(io_si,i_scpf) = sum(sites(s)%fmort_rate(i_scls, i_pft,:))
+               hio_m5_si_scls(io_si,i_scls) = hio_m5_si_scls(io_si,i_scls) + sum(sites(s)%fmort_rate(i_scls, i_pft,:))
+               !
+               hio_crownfiremort_si_scpf(io_si,i_scpf) = sites(s)%fmort_rate_crown(i_scls, i_pft)
+               hio_cambialfiremort_si_scpf(io_si,i_scpf) = sites(s)%fmort_rate_cambial(i_scls, i_pft)
+               !
+               ! fire components of overall canopy and understory mortality
+               hio_mortality_canopy_si_scpf(io_si,i_scpf) = hio_mortality_canopy_si_scpf(io_si,i_scpf) + &
+                    sites(s)%fmort_rate(i_scls, i_pft,1)
```
rgknox: I think instead of using "1" and "2", we should be using parameters defined in EDTypesMod.F90.
Something like:  canopy_layer_index = 1

Also, for these understory mortality diagnostics.  We need to sum all understory layers, not just the second.

So, put a clause, if nclmax > canopy_layer_index,    sum(canopy_layer_index+1 : nclmax)

jkshuman: Got it, and agreed. I will work through this. I did update other sections where 1 and 2 for canopy layers were hard coded, but missed this one. Thank you for catching it.

rgknox:
Pending @jkshuman 's merging of some changes, tests on cheyenne pass, b4b with master as well:
/glade/scratch/rgknox/clmed-tests/fates.cheyenne.intel.fire-mort-diags-v2-C5dc7f0a-F2b5330b

main/EDTypesMod.F90
```diff
@@ -607,17 +608,25 @@ module EDTypesMod
         
      ! TERMINATION, RECRUITMENT, DEMOTION, and DISTURBANCE
```
rgknox: @ckoven , can you sign-off that changing terminated_nindivs and termination_carbonflux from 2 layers to nclmax is ok with you?  You will see changes in the code that also reflect this change in usage, and are not separated into upper and not-upper.  I changed this because it was confusing to me on first if these arrays were allocated by canopy layer or not, and for clarity and consistency sake, changed it to be by canopy layer since that is our typical convention.  We can revert it if you feel these arrays are better allocated as the binary form, but we need to give them an integer parameter that describes this dimension if so.

main/FatesRestartInterfaceMod.F90
jkshuman: @rgknox Can you provide more details on this comment? I ran tests across South America with a tropical tree and grass with and with out fire. Both runs completed through 36 years successfully.

main/EDTypesMod.F90
rgknox: From email exchange with @ckoven: 
> "great. so long as we do that everywhere, it works for me. "

jkshuman:
@rgknox I am done merging changes and updates. This PR successfully runs with and without fire across South America with a tropical tree and grass. Both runs completed through 42 years successfully with reasonable biomass accumulation for trees and grasses. Let me know if/how you want to address your earlier comment regarding fire mortality arrays in the restarts.

PR #432, opened 2018-10-19
Add a simple scheme of leaf age to improve the seasonal cycles of GPP

xuchongang:
### Description:
This pull request resolve the mismatch between simulated GPP and observed GPP in BCI, where the simulated GPP is too flat after the dry season. This new pull request added a simple scheme that tracks the fraction of emerging, young, old and senescent leaves in the cohorts. It 

### Collaborators:
@serbinsh , @everpassenger , @ckoven , @rosiealice , @rgknox 
The leaf age is turn off by default; however, if turned on, we expected the GPP to be higher right after dryseason and thus higher seasonal variability.

### Checklist:

- [ x] My change requires a change to the documentation.
- [X ] I have updated the in-code documentation .AND. (the [technical note](https://github.com/NGEET/fates-docs) .OR. the wiki) accordingly.

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -1115,8 +1115,13 @@ contains
     
     if (EDPftvarcon_inst%evergreen(ipft) == 1)then
        if(use_leaf_age==itrue)then
-          currentCohort%leaf_md = currentCohort%bl*currentCohort%fracSenLeaves&
-	        / EDPftvarcon_inst%senleaf_long(ipft) 	 
+          if(currentSite%dstatus==1)then
+            currentCohort%leaf_md = currentCohort%bl*currentCohort%fracSenLeaves &
+	        / (EDPftvarcon_inst%senleaf_long(ipft)* EDPftvarcon_inst%senleaf_long_fdrought(ipft))
+          else
+	    currentCohort%leaf_md = currentCohort%bl*currentCohort%fracSenLeaves&
+	        / EDPftvarcon_inst%senleaf_long(ipft)
+	  endif			 
        else
          currentCohort%leaf_md = currentCohort%bl / EDPftvarcon_inst%leaf_long(ipft)
        endif
@@ -1153,7 +1158,11 @@ contains
 	  dOldLeaves_dt=previous_bl *currentCohort%fracOldLeaves &
 	                 / EDPftvarcon_inst%oldleaf_long(ipft)*hlm_freq_day			 			 
 	  dSenLeaves_dt=previous_bl *currentCohort%fracSenLeaves &
-	                 / EDPftvarcon_inst%senleaf_long(ipft)*hlm_freq_day	
+	                 / EDPftvarcon_inst%senleaf_long(ipft)*hlm_freq_day
+	  if(currentSite%dstatus==1)  then !drought periods
+	       dSenLeaves_dt=previous_bl *currentCohort%fracSenLeaves &
+	       /(EDPftvarcon_inst%senleaf_long(ipft)*EDPftvarcon_inst%senleaf_long_fdrought(ipft))*hlm_freq_day	  
+	  endif	
 	  currentCohort%fracExpLeaves = (previous_bl *currentCohort%fracExpLeaves-dExpLeaves_dt)&
 	                /currentCohort%bl
 	  currentCohort%fracYoungLeaves = (previous_bl *currentCohort%fracYoungLeaves+dExpLeaves_dt- &
@@ -1485,7 +1494,7 @@ contains
           currentCohort%bl        = currentCohort%bl + bl_flux
           currentCohort%npp_leaf  = currentCohort%npp_leaf + bl_flux / hlm_freq_day
           if(use_leaf_age==itrue.and.currentCohort%bl>0.0_r8) then
-	     currentCohort%fracExpLeaves = (previous_bl *currentCohort%fracExpLeaves + bl_flux)&
+	     currentCohort%fracExpLeaves = (previous_bl *currentCohort%fracExpLeaves + bl_flux) &
 	                                   /currentCohort%bl
 	     currentCohort%fracYoungLeaves = (previous_bl *currentCohort%fracYoungLeaves)/currentCohort%bl
 	     currentCohort%fracOldLeaves = (previous_bl *currentCohort%fracOldLeaves)/currentCohort%bl
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -284,11 +284,11 @@ contains
                    ! Leaf cost at leaf level z accounting for sla profile
 		   if(use_leaf_age==itrue)then
                       currentCohort%leaf_cost = 1.0_r8/(sla_levleaf* &
-                        (currentCohort%fracExpLeaves*EDPftvarcon_inst%expleaf_long(ipft)+ &
-			 currentCohort%fracYoungLeaves*EDPftvarcon_inst%youngleaf_long(ipft) + &
-			 currentCohort%fracOldLeaves*EDPftvarcon_inst%oldleaf_long(ipft) + &
-			 currentCohort%fracSenLeaves*EDPftvarcon_inst%senleaf_long(ipft) ) & 
-			 *1000.0_r8) !convert from sla in m2g-1 to m2kg-1		   
+                        (currentCohort%fracExpLeaves*EDPftvarcon_inst%expleaf_flong(ipft)+ &
+			 currentCohort%fracYoungLeaves*EDPftvarcon_inst%youngleaf_flong(ipft) + &
+			 currentCohort%fracOldLeaves*EDPftvarcon_inst%oldleaf_flong(ipft) + &
+			 currentCohort%fracSenLeaves*EDPftvarcon_inst%senleaf_flong(ipft) ) & 
+			 *EDPftvarcon_inst%leaf_long(ipft)*1000.0_r8) !convert from sla in m2g-1 to m2kg-1		   
 		   else
                       currentCohort%leaf_cost = 1.0_r8/(sla_levleaf* &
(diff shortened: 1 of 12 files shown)


main/FatesInterfaceMod.F90
```diff
@@ -466,6 +466,9 @@ module FatesInterfaceMod
       ! leaf maintenance respiration rate (umol CO2/m**2/s) 
       ! (NOT CURRENTLY USED, PLACE-HOLDER)
       !real(r8), allocatable :: lmrcanopy_pa(:)
+      
+      ! area weighted leaf water potential (MPa) 
+      real(r8), allocatable :: lwp_pa(:)
```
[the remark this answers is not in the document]
xuchongang: Ryan,
  Thank you very much for pointting this point. I added this to the boundary layer so that I can link this to the leaf moisture for caluclation of transpiration in the host model. I can remove this change!

xuchongang:
@rgknox , per your point, I have removed lwp_pa from bc_out.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -281,8 +282,17 @@ subroutine trim_canopy( currentSite )
...
                 else !evergreen costs
                    ! Leaf cost at leaf level z accounting for sla profile
-                   currentCohort%leaf_cost = 1.0_r8/(sla_levleaf* &
+		   if(use_leaf_age==itrue)then
+                      currentCohort%leaf_cost = 1.0_r8/(sla_levleaf* &
+                        (currentCohort%fracExpLeaves*EDPftvarcon_inst%expleaf_flong(ipft)+ &
+			 currentCohort%fracYoungLeaves*EDPftvarcon_inst%youngleaf_flong(ipft) + &
+			 currentCohort%fracOldLeaves*EDPftvarcon_inst%oldleaf_flong(ipft) + &
+			 currentCohort%fracSenLeaves*EDPftvarcon_inst%senleaf_flong(ipft) ) & 
+			 *EDPftvarcon_inst%leaf_long(ipft)*1000.0_r8) !convert from sla in m2g-1 to m2kg-1
```
rosiealice: @xuchongang could this be written in an extensible way with 'n' pools of leave and the frac_leaf_age as a vector?

xuchongang: @rgknox , I rember you told me that there will be error if you put arrays on the cohort. If it is Ok, I can easily update the codes to do that.

rosiealice:
OK don't worry about it!

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1096,14 +1145,38 @@ subroutine PlantGrowth( currentSite, currentCohort, bc_in )
...
+	  endif	
+	  currentCohort%fracExpLeaves = (previous_bl *currentCohort%fracExpLeaves-dExpLeaves_dt)&
+	                /currentCohort%bl
+	  currentCohort%fracYoungLeaves = (previous_bl *currentCohort%fracYoungLeaves+dExpLeaves_dt- &
+	               dYoungLeaves_dt)/currentCohort%bl
+	  currentCohort%fracOldLeaves = (previous_bl *currentCohort%fracOldLeaves+dYoungLeaves_dt- &
+	               dOldLeaves_dt)/currentCohort%bl
+	  currentCohort%fracSenLeaves = (previous_bl *currentCohort%fracSenLeaves+dOldLeaves_dt- &
+	               dSenLeaves_dt)/currentCohort%bl	
+    endif
```
ckoven: should we add a check here to ensure that fracExpLeaves + fracYoungLeaves + fracOldLeaves + fracSenLeaves = 1?

biogeochem/EDPhysiologyMod.F90
ckoven: shouldn't this leaf cost term be the same logic as the maintenance demand calculated as in https://github.com/NGEET/fates/pull/432/files#diff-b2580903c042bbb143bb267a7b882e3fR1120; i.e. based on the senescent pool and turnover rate?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1079,7 +1117,18 @@ subroutine PlantGrowth( currentSite, currentCohort, bc_in )
     end if
     
     if (EDPftvarcon_inst%evergreen(ipft) == 1)then
-       currentCohort%leaf_md = currentCohort%bl / EDPftvarcon_inst%leaf_long(ipft)
+       if(use_leaf_age==itrue)then
+          if(currentSite%dstatus==1)then
```
ckoven: Is this supposed to apply to the evergreen phenology or the drought-deciduous phenology?  Assuming the former since the parameter file doesn't have that as a thing that's changed. But if so, then won't this have the effect of pushing the leaf further off its target and therefore causing the plant to reallocate more new growth towards leaves?  does this need to be coupled to logic that reduces the target leaf biomass by some factor during the drought period?

parameter_files/fates_params_14pfts.cdl
```diff
@@ -308,6 +308,24 @@ variables:
...
+	float fates_oldleaf_flong(fates_pft) ;
+		fates_oldleaf_flong:units = "fraction" ;
+		fates_oldleaf_flong:long_name = "Fraction of leaf longevity for old leaves(ie turnover timescale)" ;	
+	float fates_senleaf_flong(fates_pft) ;
+		fates_senleaf_flong:units = "fraction" ;
+		fates_senleaf_flong:long_name = "Fraction of leaf longevity for senescent leaves(ie turnover timescale)" ;	
+	float fates_senleaf_long_fdrought(fates_pft) ;
+		fates_senleaf_long_fdrought:units = "unitless[0-1]" ;
+		fates_senleaf_long_fdrought:long_name = "multiplication factor for leaf longevity of senescent leaves during drought" ;
+	float fates_senleaf_long_fdrought(fates_pft) ;
```
rgknox: duplicate entry?

biogeochem/EDPhysiologyMod.F90
rgknox: I think up to this point, we have been using dstatus in the phenology routine to flag if leaves are on or off, with instantaneous effects.   It seems that this is a way to make leaves drop over a period of time when drought status is triggered?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -686,7 +713,12 @@ subroutine phenology_leafonoff(currentSite)
                    ! add falling leaves to litter pools . convert to KgC/m2                    
                    currentCohort%leaf_litter = currentCohort%bl  
                    currentCohort%bl          = 0.0_r8
```
rgknox: @xuchongang , see here, if dstatus is 1, then leaves are dropped, so bl is zero, and there wont be any drought amplified maintenance turnover...  Am I missing something?

biogeochem/EDPhysiologyMod.F90
rgknox: Also, isn't this rate supposed to represent the average, or time-integrated, turnover rate?  We are using the instantaneous fractions of leaf-ages for this calculation that is performed once per year.  Although, I suppose that if a plant is keeping up with its allocation and is continually on allometry, then this is a steady state problem?

xuchongang: @rgknox agreed and keep the turnover based on the mean leaf longevity will resolve this issue.

biogeochem/EDPhysiologyMod.F90
xuchongang: This is for evergreen and this is based on the observations of more leaf turnover during the drought period. I think it may be reasonable to ajdust the target biomass during the drought period if we have the data.

biogeochem/EDPhysiologyMod.F90
xuchongang: dry amplified maintenace turnover is for evergreen trees.

main/EDPftvarcon.F90
```diff
@@ -54,6 +54,11 @@ module EDPftvarcon
      real(r8), allocatable :: slamax(:)
      real(r8), allocatable :: slatop(:)
      real(r8), allocatable :: leaf_long(:)
```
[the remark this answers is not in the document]
xuchongang: I just did all new variables.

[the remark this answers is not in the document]
xuchongang: added with new commit.

biogeochem/EDPhysiologyMod.F90
xuchongang: added the check with new commit.

xuchongang:
@rgknox , I have resolved all the issues raised. Let  me know if you need further modification before your merge.

xuchongang:
@rgknox

rgknox:
update:  @xuchongang and I have chatted offline and we are going to add a new dimension to the parameter file for leaf age class.
vcmax and leaf longevity (and others if desired) will be dimensioned by leaf age class as well as PFT.  If users don't wish to stratify leaf age classes, they simply use a dimension of 1 for this and should maintain their original behavior.
Anyone have any issues with this?
Also, because the conflicts with PARTEH are somewhat messy, we will be issueing a new PR.

rgknox:
One complication that I think we need to be aware of:

When we perform photosynthesis, we perform calculations for each  leaf-layer, in each canopy layer, for each PFT  ( i.e. PFT x leaf-layer x canopy layer).   This is because cohorts that share leaf layers of the same PFT, have exactly the same photosynthesis parameters, so we don't have to loop every single cohort's leaf layer.

We have provisions for hydraulics that will bypass this simplification because the leaf hydraulic traits of cohorts of the same PFT and same leaf layer are not necessarily the same.  This is necessary, but it slows down the model.

If someone is using more than 1 leaf age class, then this will force the effective vcmax, jmax and tpu limitations on each leaf layer to now be different from one cohort to the next.  So, we will have to also bypass the simplification in the same way that hydro does.

rgknox:
@xuchongang and I are working through a replacement branch that is PARTEH compliant, closing this.

PR #433, opened 2018-10-19
Add the option to initialize forest with canopy closure

xuchongang:
### Description:
This pull request add an option to initialize the forest with closed canopy within EDTypesMod. Currently, the site_spead that specified the canopy openness is set to 1.0, which indicate a wide open canopy. This can cause issues when initialzed from forest inventory, many small cohorts are excluded as the maximum value of crown area to dbh are used for two canopy layers.  Users now can turn on this option so that the model will use the mimimum value of crown area to dbh for initialization. The use can also choose to increase the maximum number of layers in addition to this option.

### Collaborators:
@rgknox , @everpassenger , @ckoven , @rosiealice 

If this option is turned on, the model will use mimimum value of crown area to dbh for initialization and thus accomodate more trees to avoid small trees to be excluded in the simulation.

### Checklist:

- [ X] My change requires a change to the documentation.
- [X ] I have updated the in-code documentation .AND. (the [technical note](https://github.com/NGEET/fates-docs) .OR. the wiki) accordingly.

### Test Results:

FATES baseline hash-tag:

Test Output:

ALL PASS, b4b, with master, cheyenne, intel, fates test suite


ckoven:
Hi @xuchongang  -- Do we actually need this logic?  If you specify in your parameter file that fates_allom_d2ca_coefficient_max = fates_allom_d2ca_coefficient_min, then there is no elasticity to the crown allometry and so the canopy closure logic becomes irrelevant.  Is there a case where that isn't a sufficient solution?

xuchongang:
Charlie,

  Yes, you can make that optional change in the parameter file.  I thought this could be very useful as we want to keep the parameter file untouched. The  case that this could be useful will be when forest die off happens and create an open canopy and thus we want to keep the maximum value of fates_allom_d2ca_coefficient_max to make the simulation valid. 

Yours
Chonggang

rgknox:
@xuchongang , this all looks fine to me.  For foresting inventory initialization, I had set it to 0.5 spread as the default. I wonder if we should set the default there to 1, when not using your dense forest initialzation clause.
@ckoven , as the the one who refactored spread, thoughts on this?

https://github.com/NGEET/fates/blob/master/main/FatesInventoryInitMod.F90#L931

ckoven:
I guess I think that maybe for inventory initialization, maybe setting spread to 0 as the default is the safest option, since it won't lead to termination mortality from pushing things into the nlevcan+1th layer? In the event that you init to a sparse forest, it would then increase the spread, but that way is safer than the opposite direction.

xuchongang:
@rgknox , @ckoven ,  I think the best way is to set this parameter in the input parameter file so that the user can choose what type of forest to initialize (e.g., savana will set spread =1.0 and dense forest as 0.0).

xuchongang:
I agree with @ckoven  on setting the default as 0 for the forest inventory.

xuchongang:
I personally feel that we can first make the pull request and then add  this parameter in the parmater file later for an advanced option.

rosiealice:
Unless all the tall and thin trees respite themselves to death in the
meantime?

rgknox:
@xuchongang , how do you feel about the idea of adding a parameter initial_spread to EDTypesMod instead of init_dense_flag?  It would be a real parameter ranging from 0-1.  The default would be 1.0.  That way we can specify this initial_spread to both the bare-ground and the inventory initialization.  
We could also, like you mention, move this to the parameter file if desired.

rgknox:
running tests

PR #435, opened 2018-10-19
coastal vegetation updates for FATES

xuchongang:
### Description:
This pull request make the following five key changes for coastal salt marsh. 
1) update FATES so that, when maximum height/dbh reached, all the carbons goes to the bstore for grasses;
2) when plant dies, bstore goes to reproduction for grasses;
3) recruitment only happens at the growing seasons so that the newly devloped cohorts are viable
4) add a new capability to simulate the impact of salinity and sulphide through btran
5) new parameter file for Spartina
### Collaborators:
Joel Rowland, @rosiealice , @ckoven , @rgknox 
We expect the model to produce better seasonal cycle for grasses within FATES and add the potential for simulating salinity impact

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


biogeochem/EDPhysiologyMod.F90
```diff
@@ -751,8 +751,14 @@ subroutine seeds_in( currentSite, cp_pnt )
        do while (associated(currentCohort))
           do p = 1, numpft
              if (pft_present(p)) then
-                currentPatch%seeds_in(p) = currentPatch%seeds_in(p) +  currentCohort%seed_prod * currentCohort%n / &
+	        if(EDPftvarcon_inst%woody(p)==itrue)then
+                  currentPatch%seeds_in(p) = currentPatch%seeds_in(p) +  currentCohort%seed_prod * currentCohort%n / &
                      (currentPatch%area * npfts_present)
+		else
+		  !for grasses, the bstore goes to the reproduction through colonial reproduction when grass dies
+                  currentPatch%seeds_in(p) = currentPatch%seeds_in(p) + (currentCohort%seed_prod * currentCohort%n - &
```
rosiealice: do we want this to always happen, or only happen in salty mode? (I can see why we might want it to always happen...)

biogeochem/EDPhysiologyMod.F90
```diff
@@ -763,8 +769,15 @@ subroutine seeds_in( currentSite, cp_pnt )
     currentCohort => currentPatch%tallest
     do while (associated(currentCohort))
        p = currentCohort%pft
-       currentPatch%seeds_in(p) = currentPatch%seeds_in(p) +  &
+       if(EDPftvarcon_inst%woody(p)==itrue)then
+         currentPatch%seeds_in(p) = currentPatch%seeds_in(p) +  &
              currentCohort%seed_prod * currentCohort%n/currentPatch%area
+       else
+         !for grasses, the bstore goes to the reproduction through colonial reproduction when grass dies
```
rosiealice:
```suggestion
         !for grasses, the bstore goes to the reproduction through clonal reproduction when grass dies
```

biogeochem/EDPhysiologyMod.F90
```diff
@@ -840,7 +853,14 @@ subroutine seed_germination( currentSite, currentPatch )
 
     do p = 1,numpft
        currentPatch%seed_germination(p) =  min(currentSite%seed_bank(p) * &
-             EDPftvarcon_inst%germination_timescale(p),max_germination)
+             EDPftvarcon_inst%germination_timescale(p),max_germination)     
+       !set the germination only under the growing season...c.xu
+       if (EDPftvarcon_inst%season_decid(p) == 1.and.currentSite%status == 1)then
```
rosiealice: good idea!

biogeochem/EDPhysiologyMod.F90
xuchongang: @rosiealice , good point and I can add a parameter in the input file so that only a fraction of storage goes to reproduction. If the fraction set to zero by default, then no storage goes to reproduction.

xuchongang: @rosiealice , I have added a new parameter in the fates parameter file so that a specified  proportion of bstore goes to reproduction. If needed, this parameter can be used for woody plants for resprout. The default is set to zero so that the code behaviors should be same as before.

xuchongang:
@rosiealice , good point and I can add a parameter in the input parameter file so that only a fraction of storage goes to reproduction. If the fraction set to zero by default, then no storage goes to reproduction.

biogeochem/EDPhysiologyMod.F90
xuchongang: This updates resolves some seasonal cycle component for grass at Delaware but I feel this could be good improvement for temperate/boreal forests....but will not impact tropical forest.

rgknox:
testing on cheyenne

rgknox:
all tests pass on cheyenne. Expect for the LAI output variables fixed by @xuchongang , everything else is B4B.

Issue #438, opened 2018-11-02
need to revamp promotion/demotion in "strict PPA" so that it is PFT-order-independent


ckoven:
A thing that we've been aware of for a while is that, when the fates_comp_excln parameter is set to -1, triggering the strict or deterministic PPA method of cohort promotion/demotion, is that if two cohorts from different PFTs that have identical size are at the edge of the canopy layer boundaries, the way they get promoted and demoted is order dependent.  In general this isn't an issue, but we've noticed in in the past when we start two identical PFTs, one will end up taking over because of this.  I've just encountered a second, more important, case where this matters though: if we compete two non-identical PFTs using inventory initialization with identical starting size distributions of the PFTs.  The initial bit of PPA sorting leads to one PFT starting off in a stronger position than the other because of this.

So we need to fix this by adding logic to the strict PPA sorting where it checks to see if there are one or more cohorts with the same DBH, and then if so, splits each of them instead of just splitting the first cohort that it encounters at the edge of the canopy layers.

This doesn't apply to @rosiealice's original stochastic PPA sorting method (fates_comp_excln > 0), which doesn't suffer from this issue.

ckoven:
note, that this has some conceptual overlap with https://github.com/NGEET/fates/issues/302.  not sure if it makes sense to tackle both at once, or separately.

ckoven:
ok.  the nice thing about this issue is that solving it also solves the issue of equal-height trees due to height-capped allometries that complicated #302.  but right now this one is blocking an experiment that I'm trying to work through, so I will try to solve this first and then #302 shortly afterwards.

ckoven:
initial attempt to fix this with commit ee4344f. if anybody is up for checking whether the logic makes sense, please feel free.

ckoven:
fixed by #441, closing.

Issue #302, opened 2017-11-21
Make height sorting based on height rather than DBH


ckoven:
Currently, the height sorting in FATES is actually diameter sorting rather than height sorting, which was okay when the allometry was baked in, but isn't ok anymore.  We need to change it to be based on height instead.  The issue applies in both of our sorting algorithms, though in slightly different ways.

The issue comes up in two places using the "stochastic PPA" sorting algorithm. Here:
https://github.com/NGEET/fates/blob/master/biogeochem/EDCanopyStructureMod.F90#L324
and here:
https://github.com/NGEET/fates/blob/master/biogeochem/EDCanopyStructureMod.F90#L646

In the "deterministic PPA" algorithm, it uses only the linked-list cohort%shorter and cohort%taller sorting order, which again is based on diameter rather than height.  See here:
https://github.com/NGEET/fates/blob/master/biogeochem/EDCohortDynamicsMod.F90#L1083
and here:
https://github.com/NGEET/fates/blob/master/biogeochem/EDCohortDynamicsMod.F90#L1091

The main complication here will be that for some allometries (e.g. in #293), height is capped, whereas DBH isn't and is therefore a more reliable way of distinguishing rank order at the large size.  So we will need to add logic in the linked list sorting algorithm to handle the case that heights are identical (on account of having hit a cap) while diameters are still different.

I don't _think_ that anyone has been competing trees with different height-diameter allometries against each other, but if they have, then this would obviously have affected results.  Otherwise it shouldn't affect results, but still needs to be fixed.

ckoven:
this was fixed by #441.

PR #441, opened 2018-11-07
sorting logic changed to better handle equal-sized cohorts and to sort based on height

ckoven:
### Description:
This PR changes the sorting so that: 

- if using the deterministic PPA sorting algorithm, handle the special condition where two cohorts of different PFTs but the same size get demoted/promoted in proportion to their area rather than based on sorting order.
- sorting is switched to be a function of height rather than DBH.

This PR is meant to solve #438 and #302.

This should change answers in a few ways:

1. if 2 identical PFTs starting from bare ground are run using the strict PPA sorting algorithm, they should remain equal over time, rather than diverging.
2. if 2 non-identical PFTs are initialized to identical initial conditions using the inventory initialization, then they should actually start with the same structure rather than having some PFT-order effect thrown in.
3. if 2 PFTs with non-identical height-to-dbh allometry are competing, then their competitive dynamics will be different (and hopefully more correct)
4. in all cases when using stochastic PPA sorting algorithm, the answers should change because the promotion term is now height ** comp_excln parameter rather than dbh ** comp_excln.  To go back to the same behavior as before, it ought to work to change the comp_excln parameter by the exponent used in the height-to-dbh allometry.

### Collaborators:
discussed with @rgknox and @rosiealice.  also pinging @jenniferholm as this will effect her simulations due to point 3 above

### Checklist:

### Test Results:
not yet tested.  hold until both formal and more informal tests have been done.

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -396,7 +396,7 @@ contains
 
                   if ( tied_size_with_neighbor ) then
                      ! now we need to go through and figure out how many equal-size cohorts there are.
-                     ! then we need to go through, add up the collective corwn areas of all equal-sized and equal-canopy-layer cohorts,
+                     ! then we need to go through, add up the collective crown areas of all equal-sized and equal-canopy-layer cohorts,
                      ! and then demote from each as if they were a single group
                      !
                      total_crownarea_of_tied_cohorts = currentCohort%c_area
@@ -454,7 +454,7 @@ contains
                     has_taller_equalsized_neighbor) then
                   sumweights_equalsizebuffer = sumweights_equalsizebuffer + currentCohort%excl_weight
                else if (ED_val_comp_excln .lt. 0.0_r8) .and. tied_size_with_neighbor) then
-                  sumweights = sumweights + sumweights_equalsizebuffer
+                  sumweights = sumweights + currentCohort%excl_weight + sumweights_equalsizebuffer
                   sumweights_equalsizebuffer = 0._r8
                else
                   sumweights = sumweights + currentCohort%excl_weight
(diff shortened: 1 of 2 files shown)


ckoven:
I've updated the tech note to reflect the change in this PR: https://github.com/NGEET/fates-docs/pull/14

rgknox:
@ckoven and I have had a good amount of back and forth on this PR, see discussion and analysis here:

https://github.com/ckoven/fates/pull/3

Will start testing

ckoven:
Just re-discussed this with @rgknox.  We're going to hold off on merging this for a bit, as there's one thing in the change that we don't fully understand yet, which is that up until now, if you run two identical PFTs with identical initial conditions and using probabilistic sorting, they stay identical with respect to each other over time.  Something about changing from DBH-based sorting to height-based sorting is causing the model to lost that feature: with probabilistic sorting the PFTs remain very close, but no longer identical. (With deterministic sorting the ability to preserve identicalness of identical PFTs is now just as good as with probabilistic sorting, which is a good new feature and part of the motivation for this PR, but also is not perfect). We're going to stare at this a little longer and see if we can't figure out why.  If anyone else wants to take a look, please feel free.

rgknox:
In our modeling call, @rosiealice suggested we try a test simulation where we use a very large diameter where height capping occurs, which allows us to test a hypothesis.  The premise is that height based promotion and demotion has the potential that many cohorts are at the same height, and thus a potentially large group must all be demoted together.  Height based promotion/demotion has a harder time maintaining precision and exactness between clones when it is forced to promote many of the same height, perhaps because small number densities have to be divided up more times (making very very small numbers that run into precision issues).

In this test, I set the capping height to 150000, and thus removing the chance that cohorts of different diameters would have different heights.

https://drive.google.com/file/d/1Isqo4hLiKDomEyZfo2rGxnPH4hyUz7-i/view?usp=sharing

The test shows that the large cap has no discernible improvement on maintaining parity between two cohorts with the same PFT definition.

ckoven:
really?  I feel like that looks much closer to the first figure in https://github.com/ckoven/fates/pull/3#issuecomment-454278571 than to the other ones.  they both have the one pixel in southern africa that is not equal but otherwise significantly less than a percent error.

rgknox:
After looking at this again, I'm realizing that the master version was not as pristine as the I remember.  And, in fact, it looks like the break down in parity between the two clones is at a smaller magnitude that what we saw in master.  
I'm going to run final tests on this, If anyone has any objections to integrating this, make your peace.

rgknox:
See #469 regarding limited testing capabilities right now. 

Expected tests pass:
/gpfs/fs1/scratch/rgknox/clmed-tests/fates.cheyenne.intel.height-sorting-approved-v3-wrtm-Cd909f57-F2e73ba8

PR #448, opened 2018-12-17
Increase maxlayer to 3, reduce max cohorts to 40

rgknox:
### Description:

This is a simple set of changes that changes to constants in the code. 

 By increasing the maximum number of canopy layers (nclmax) allowed to 3, FATES will allow a multi-layer canopy instead of artificially culling understory plants that exceed the second layer.  Thus far, dynamic simulations have not had a viable 3rd story in any known context, so this has not been much of an issue. However, inventory initialized simulations such as those at BCI will generate a 3rd story, so this change will enable them to be conducted correctly, particularly when running in static stand structure mode.

The maximum number of cohorts per patch has been a very large number as a default (~150). This number is somewhat of a legacy, in that in the past we set this number very high as it was necessary for memory allocation purposes in restart files.  Since this is not true anymore, the maximum number of cohorts can be reduced.  The ED2 model for instance, very often uses less than 50 cohorts per patch, and typically lower than that.  

A sensitivity analysis of going from 2 to 3 maximum number of canopy layers, and 150 to 40 maximum number of cohorts, at BCI was conducted and is provided in discussion below. The results showed a very small sensitivity to the changes.  I will follow up with timing information.

Yes, answers should not be B4B, but they should also not be different with significance.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDCohortDynamicsMod.F90
+++ b/biogeochem/EDCohortDynamicsMod.F90
@@ -603,7 +603,7 @@ contains
             endif
          endif
 
-         ! In the third canopy layer
+         ! Outside the maximum canopy layer
          if (currentCohort%canopy_layer > nclmax ) then 
            terminate = 1
            if ( debug ) then
--- a/main/EDTypesMod.F90
+++ b/main/EDTypesMod.F90
@@ -16,13 +16,13 @@ module EDTypesMod
   save
 
   integer, parameter :: maxPatchesPerSite  = 10   ! maximum number of patches to live on a site
-  integer, parameter :: maxCohortsPerPatch = 80  ! maximum number of cohorts per patch
+  integer, parameter :: maxCohortsPerPatch = 100  ! maximum number of cohorts per patch
   
   integer, parameter :: nclmax = 3                ! Maximum number of canopy layers
   integer, parameter :: ican_upper = 1            ! Nominal index for the upper canopy
   integer, parameter :: ican_ustory = 2           ! Nominal index for understory in two-canopy system
 
-  integer, parameter :: nlevleaf = 40             ! number of leaf layers in canopy layer
+  integer, parameter :: nlevleaf = 20             ! number of leaf layers in canopy layer
   integer, parameter :: maxpft = 15               ! maximum number of PFTs allowed
                                                   ! the parameter file may determine that fewer
                                                   ! are used, but this helps allocate scratch
--- a/main/EDTypesMod.F90
+++ b/main/EDTypesMod.F90
@@ -16,9 +16,9 @@ module EDTypesMod
   save
 
   integer, parameter :: maxPatchesPerSite  = 10   ! maximum number of patches to live on a site
-  integer, parameter :: maxCohortsPerPatch = 160  ! maximum number of cohorts per patch
+  integer, parameter :: maxCohortsPerPatch = 40  ! maximum number of cohorts per patch
   
-  integer, parameter :: nclmax = 2                ! Maximum number of canopy layers
+  integer, parameter :: nclmax = 3                ! Maximum number of canopy layers
   integer, parameter :: ican_upper = 1            ! Nominal index for the upper canopy
   integer, parameter :: ican_ustory = 2           ! Nominal index for understory in two-canopy system
 
--- a/main/EDTypesMod.F90
+++ b/main/EDTypesMod.F90
@@ -16,7 +16,7 @@ module EDTypesMod
   save
 
   integer, parameter :: maxPatchesPerSite  = 10   ! maximum number of patches to live on a site
-  integer, parameter :: maxCohortsPerPatch = 40  ! maximum number of cohorts per patch
+  integer, parameter :: maxCohortsPerPatch = 80  ! maximum number of cohorts per patch
   
   integer, parameter :: nclmax = 3                ! Maximum number of canopy layers
   integer, parameter :: ican_upper = 1            ! Nominal index for the upper canopy
--- a/main/FatesRestartInterfaceMod.F90
+++ b/main/FatesRestartInterfaceMod.F90
@@ -677,8 +677,8 @@ contains
     ! 1D cohort Variables
     ! -----------------------------------------------------------------------------------
 
-    call this%set_restart_var(vname='fates_canopy_layer', vtype=cohort_r8, &
-         long_name='ed cohort - canopy_layer', units='unitless', flushval = flushzero, &
+    call this%set_restart_var(vname='fates_canopy_layer', vtype=cohort_int, &
+         long_name='ed cohort - canopy_layer', units='unitless', flushval = flushinvalid, &
          hlms='CLM:ALM', initialize=initialize_variables, ivar=ivar, index = ir_canopy_layer_co )
 
     call this%set_restart_var(vname='fates_canopy_layer_yesterday', vtype=cohort_r8, &
@@ -1448,7 +1448,7 @@ contains
            rio_ncohort_pa              => this%rvars(ir_ncohort_pa)%int1d, &
            rio_solar_zenith_flag_pa    => this%rvars(ir_solar_zenith_flag_pa)%int1d, &
            rio_solar_zenith_angle_pa   => this%rvars(ir_solar_zenith_angle_pa)%r81d, &
-           rio_canopy_layer_co         => this%rvars(ir_canopy_layer_co)%r81d, &
+           rio_canopy_layer_co         => this%rvars(ir_canopy_layer_co)%int1d, &
            rio_canopy_layer_yesterday_co    => this%rvars(ir_canopy_layer_yesterday_co)%r81d, &
            rio_canopy_trim_co          => this%rvars(ir_canopy_trim_co)%r81d, &
            rio_size_class_lasttimestep => this%rvars(ir_size_class_lasttimestep_co)%int1d, &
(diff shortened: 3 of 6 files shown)


rgknox:
80 Year regression test with master at BCI, run through the ACRE test suite:
https://drive.google.com/file/d/1KeEDrZeV3aNFXFgir7f8Se_b1Zm6yRev/view?usp=sharing

ckoven:
I'm fine with changing nclmax to 3.

Re the maxCohortsPerPatch issue though: by my back-of-envelope calculations, changing maxCohortsPerPatch will end up having a big change in the cohort error tolerances.  If our smallest cohorts recruit at around 0.3 cm DBH, and we can expect trees to possibly grow to about 150 cm, and our cohort fusion criteria is that we want their DBH to not differ by more than 5%, this means that for each PFT we should roughly end up with a geometric series of sizes.  So, if we want the tolerances and the max # of cohorts to be consistent with each other, the maximum size should satisfy the equation: largest dbh = smallest_dbh * (1+fusion tolerance) ** max cohorts

In this case, to arrive at about a 150cm maximum size, we'd need to get: 0.3cm * 1.05**128 = 155 cm.  So if we go smaller than maxCohortsPerPatch = 128 then we'll start to degrade the cohort fusion tolerances beyond our criteria.  By the same logic, I'd estimate that setting maxCohortsPerPatch to 40 would end up giving us cohort fusion error tolerances of about 17% DBH.  

All that is for 1 PFT.  If we have multiple PFTs, it applies even more so.  I.e. for 2 coexisting PFTs, maxCohortsPerPatch=40 means that they'd each differ by about 36% at steady state.  

Anyway I'm perfectly willing to accept that 5% fusion tolerances might be overkill, and I don't really know the costs and benefits here, but I guess I'd advocate that we should probably do more sensitivity here before changing the maxCohortsPerPatch parameter?

rosiealice:
It occurs to me that these calculations of the corresponding error
tolerances and maximum cohort number should perhaps somehow be in the code,
and one emerge from the other?

And yes, it's much better if we can have more than 2 canopy layers. That
was just for sanity's sake for fixing the radiation routines.


On Dec 17, 2018 11:36 PM, "Charlie Koven" <notifications@github.com> wrote:

I'm fine with changing nclmax to 3.

Re the maxCohortsPerPatch issue though: by my back-of-envelope
calculations, changing maxCohortsPerPatch will end up having a big change
in the cohort error tolerances. If our smallest cohorts recruit at around
0.3 cm DBH, and we can expect trees to possibly grow to about 150 cm, and
our cohort fusion criteria is that we want their DBH to not differ by more
than 5%, this means that for each PFT we should roughly end up with a
geometric series of sizes. So, if we want the tolerances and the max # of
cohorts to be consistent with each other, the maximum size should satisfy
the equation: largest dbh = smallest_dbh * (1+fusion tolerance) ** max
cohorts

In this case, to arrive at about a 150cm maximum size, we'd need to get:
0.3cm * 1.05**128 = 155 cm. So if we go smaller than maxCohortsPerPatch =
128 then we'll start to degrade the cohort fusion tolerances beyond our
criteria. By the same logic, I'd estimate that setting maxCohortsPerPatch
to 40 would end up giving us cohort fusion error tolerances of about 17%
DBH.

All that is for 1 PFT. If we have multiple PFTs, it applies even more so.
I.e. for 2 coexisting PFTs, maxCohortsPerPatch=40 means that they'd each
differ by about 36% at steady state.

Anyway I'm perfectly willing to accept that 5% fusion tolerances might be
overkill, and I don't really know the costs and benefits here, but I guess
I'd advocate that we should probably do more sensitivity here before
changing the maxCohortsPerPatch parameter?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/NGEET/fates/pull/448#issuecomment-448040069>, or mute
the thread

rgknox:
After talking with @ckoven, we decided to be a little more gradual in trimming down our cohort cap, and then in the future dedicate more time to studying the sensitivities of our cohort cap in more detail. I've provided some evidence here of limited sensitivity, but that might not be true for different PFTs in different contexts.
Anyhow, I will change the cohort cap to 80.   After doing some back-of-the-envelope calculations that @ckoven  came up with, we arrived at a fusion tolerance "tol" that should roughly equate to 80 intervals between our smallest cohort "d1 = 0.3 cm" and a hypothetically large cohort "d2 = 150 cm".

d2 = d1 * (1 + tol)^n    
tol =   (d2/d1)^(1/n)  - 1  =     450^(1/n) -1 =  0.079    ~ 8%

jenniferholm:
@rgknox - I think this sounds like a better idea to do a more gradual reduction in max number of cohorts, and 80 sounds good. Thanks for showing that it would be ~8% fusion tolerance, and that seems a lot better than the 17% or 36% that Rosie was referring to. 

About the max canopy layers. I am really interested in seeing what happens with 3 canopy layers. But I'm wondering if we should still try and investigate the top canopy layer (layer 1). This layer always seems to fill in very fast, and come to a near-full canopy closure. And the issue I'm concerned with, is that it stays constant at this "full" canopy. I don't think this is anything new, and we know there has been challenges with getting an understory. (And we've tried to achieve this with relaxing the PPA). 

It would be nice if the top canopy layer 1 would have reductions in the crown_area when a large tree death/disturbance occurred, corresponding with a gap opening. But at least as far as I can tell, that isn't occurring. Have others found a 'breakage' or reduction in the canopy crown area when a large tree mortality occurs? But, maybe this is all solved by the canopy closure threshold being set at 0.8, and allowing light into the understory? Just some thoughts about canopy layer issues. 

Attaching some figures for reference. This is the ZF2 site, with a clear-cutting disturbance applied, after the forest has been spun up to a mature steady-state. The top canopy layer quickly returns to near 1, and then stays constant. Maybe this stable canopy area coverage is too stable. 

[Crown_area_disturbance.pdf](https://github.com/NGEET/fates/files/2728019/Crown_area_disturbance.pdf)

ckoven:
Hi @jenniferholm I think this is a definition issue?  When understory trees get promoted, they occupy the canopy, even if they are small trees; that is the essence of the PPA. The only time that the canopy isn't full is when there is bare ground under some fraction of it.  So my interpretation of the figure you show is that there are probably a lot of tiny recruits filling the canopy after the disturbance.  The subsequent stability is just a reflection of the non-stochastic behavior of the ED and PPA equations -- as long as there is something in the understory, it will fill in any ground that remains in an old patch when a canopy tree dies, and the small difference between 1.0 and the canopy closure index reflects the relatively steady-state disturbance rate and subsequent canopy closure in young patches.  Maybe what you could look at is the change to the canopy height distribution after logging?

rgknox:
In running 14 pft global tests ...
It seems that the combination of reducing our cohort cap, and increasing the maximum number of layers is creating trouble for fusion.  Very early in the simulation the dynamic fusion tolerance is growing to above 100% for small plants, and we have not fused the number below the new cap of 80.

Note that we have four conditions for fusion:
1) Are the differences in diameters within the dynamic fusion tolerance
2) Are the cohorts in the same layer?  (now there are 3 not 2)
3) Are the cohorts the same PFT?
4) Are the cohorts new? (ie experienced at least 1 day?)

For a 14 pft case: this means that we can have up to  14*4 classifications alone.  Since new cohorts will exist in the lowest layer only, that accounts for 14 classes, then the non-new cohorts can take up another 14*3 classifications.  Thats 56 classifications, taking up a potential 80 cohorts total, not even counting size differences.

One thing that might help quite a bit, is to allow fusion across canopy layers.  I really don't see any reason this can't happen.  A decision just has to be made as to which layer the fused cohort would take.. the higher or lower.  The end result shouldn't be effected, but it may have an impact on how quickly it converges.

rgknox:
We could pass a flag that signals where the fusion routine is called from.  If fusion is not called from the subroutine that handles the canopy layering, we could signal that cross-layer fusion is allowed.  This may help keep the number of cohorts down..

ckoven:
In general, I don't think we want to allow things to fuse between layers, doing so would mix their identities in ways that I don't think we want.  maybe its not actually that bad to run into the dynamic fusion tolerances early on in the spinup process. presumably we aren't having all these PFTs coexisting beyond the initial period?  Also, we aren't tracking the dynamic_fusion_tolerance as a history variable are we -- maybe we should?

rgknox:
> In general, I don't think we want to allow things to fuse between layers, doing so would mix their identities in ways that I don't think we want. 

I don't feel as strongly on this one, or worried about it.  I think its just as weird that we allow these artificial boundaries between two similar plants to be preserved.  It might actually be beneficial by reducing the amount of artificial bifurcation.

rgknox:
Another thing that I am noticing, is that due to strong seed rain, we have new cohorts, that have not even experiencing a day, populating all three layers.  Therefore, for each PFT, I'm seeing six different classifications for plants that are approximately the same size  (ie new/not-new  x layer ).  I think at the very least, we should only apply the same-layer condition to plants that are not new.

rgknox:
I think I'm coming full circle on this cross-layer fusion thing.  The more I think of it, it seems like we wouldn't gain much.  We must only allow within-layer fusion during the process of organizing the cohorts into discrete layers canopy layers (subroutine canopy_structure()), so I think that any cohort number reductions that are gained outside of that process, will be lost immediately afterwards when those fused cohorts have to split again during re-structuring... Maybe there are other benefits to cross-layer fusion, but this thought exercise was originally oriented around reducing cohort counts.
Although, I think there is still a need to address why, and how can we fix, a scenario where we have newly recruited plants in the first day, spanning 3 canopy layers... that seems dubious.
Perhaps another thing to try is allowing new cohorts to fuse with other similar cohorts, but the new ones will adopt the information from the non-new cohorts instead of fusing data that has not been initialized in them yet...

For now, perhaps we are stuck with large cohort caps when dealing with large numbers of pfts and canopy layers, however raising the fusion tolerance to 8% may help.

rosiealice:
@rgknox, @lmkueppers  and @adamhb 's calibration of the recruitment model indicated a very large decrease in recruitment rates was necessary to match the observations. I suspect we could get away with decreasing seed rain a lot, if this is the case...

rgknox:
I bumped up the maxcohort cap to 100 to save people headaches.  Also, reduced the max number of leaf layers from 40 to 20, which should allow an LAI of 20 in each crown.  Final testing on cheyenne, will integrate if all goes well.

rgknox:
expected PASS:
/gpfs/fs1/scratch/rgknox/clmed-tests/fates.cheyenne.intel.maxohort-merged-v2-nc3-2trop-Cd041226-F699b181

PR #455, opened 2018-12-20
Remove Lasslop wind effect on fire ROS

jkshuman:
Remove reducing effect of high winds based on Lasslop et al 2014

### Description:
Within SFMain remove the conditional clause that reduces fire rate of spread for high wind speeds. Within parameter files removed the parameter for fire_wind_max that triggers the conditional.
This showed an improvement in mode simulated fire behavior in past work (Lasslop et al 2014), but current research demonstrates there is not a clear link between high winds and a reducing effect on rate of spread. Andela et al 2018 ESSDD https://doi.org/10.5194/essd-2018-89 
This PR acts to remove this Lasslop et al (2014) reducing effect, and revert back to original Thonicke ROS wind effect.

### Collaborators:
JKShuman, RKnox

Fire behavior rate of spread will be higher at high winds within simulation, dependent on previous fire_wind_max parameter and forcing data.

### Checklist:

### Test Results:
Code compiles and runs on Hobart.

### DIFF
--- a/fire/SFMainMod.F90
+++ b/fire/SFMainMod.F90
@@ -408,8 +408,8 @@ contains
     use SFParamsMod, only  : SF_val_miner_total, &
                              SF_val_part_dens,   &
                              SF_val_miner_damp,  &
-                             SF_val_fuel_energy, &
-                             SF_val_wind_max
+                             SF_val_fuel_energy
+    
     use FatesInterfaceMod, only : hlm_current_day, hlm_current_month
 
     type(ed_site_type), intent(in), target :: currentSite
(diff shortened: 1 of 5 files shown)


rosiealice:
Looks good to me. Do you think there should be a comment to the effect that this option exists in the literature but we've chosen to not use it for the reasons you describe!

jkshuman:
Thanks for looking this over. I think the information in the PR as to the removal is enough.

rgknox:
running tests

rgknox:
All pass:
/gpfs/fs1/scratch/rgknox/clmed-tests/fates.cheyenne.intel.rem-laslop-C5dc7f0a-F34f580c

PR #468, opened 2019-01-27
phenology status timers

rgknox:
### Description:

This changeset contains refactors to the phenology timing routine, additions to the restart for counting cold days, and new history diagnostics to help track site level status for both cold and drought phenology.  The vector contiaining the last 10 days of vegetation temperature was given a more discriptive name.  

Additionally, since status diagnostics are relevant only to FATES columns, averaging that would ignore non-fates land columns instead of treating them as zeros was desired. This can be achieved by simply using the clm_spval as the flushval.  

Specifics on phenology timing:

For both cold and drought deciduous status timers, the "date of leaves on" and "date of leaves off" is set by the integer model day, and thus the time elapsed since those days is the difference between the model day and the saved values.

For cold phenology: 

The GDD counting was correctly calculating GDD days, however we were not zero-ing the GDD counter after leaves had flushed, and thus Northern Hemisphere sites would have ever-increasing GDDs late into the the year, and would immediately re-flush in the fall immediately after they dropped their leaves.

The "C" parameter for the GDD threshold was changed to -0.01 to match the Botta et al. 2000 paper.

An additional state of "0" was added, to accompany 1 (leaves off) and 2 (leaves on). 0 status indicates the plants are in a "leaf off" status due to time exceedance without a cold-day. Functionally, a 0 status site operates the same as a 1-status site, but the differentiation is to help identify those sites that have leaves off because they are cold and those sites that have leaves off because they are not classified as cold.

For drought phenology: 

The drought phenology algorithm is more or less unchanged in terms of how the algorithm works, the timers have been refactored to use absolute counting days instead of the day-of-year for simplicity, but the results should be the same.

### Collaborators:

@jkshuman , @ckoven, Bill Riley

See above, cold deciduous should be affected, drought deciduous should not.  Not B4B with base.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:


rgknox:
Here is a video that shows the daily progression of the cold-deciduous flags over the course of 2 years.  Note that the first 300+ days were removed because the first year has some expectedly strange behavior due to fact that the time-since-last leaf drop or leaf flush has no precedent.

https://drive.google.com/file/d/13zB1LMJOe0W1G8be9p4w9IRVMnOa1KiO/view?usp=sharing

jkshuman:
The video is great. Thanks for including that. @rgknox  I am happy to look the code over, but will take me a bit to find time.

rgknox:
I've made some updates.  And a new video here:  https://drive.google.com/file/d/1gRqFRlUS1kCZ55E6hubZEQMrLhBxAP1q/view?usp=sharing

A note about the status flags:
```
cold status          ! are leaves in this pixel on or off for cold deciduous
                           ! 0 = this site has not experienced a cold period over at least
                           !       400 days, leaves are dropped and flagged as non-cold region
                           ! 1 = this site is in a cold-state where leaves should have fallen
                           ! 2 = this site is in a warm-state where leaves are allowed to flush

drought status    ! are leaves in this pixel on or off for drought decid
                           ! 0 = leaves off due to time exceedance
                           ! 1 = leaves off due to moisture avail
                           ! 2 = leaves on due to moisture avail
                           ! 3 = leaves on due to time exceedance
```

Also, for reference, here is the GDD threshold that GDD is compared against:
https://drive.google.com/file/d/1YdNz_TOXss788Kuf6cKQE3zQSO3oVq1V/view?usp=sharing

rgknox:
FYI: I'm holding on this integration until I can perform long-term global tests with the default 14 pft testing file.  Bug fixing related to issue #494 is required before I can do this.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -408,135 +408,185 @@ subroutine phenology( currentSite, bc_in )
...
-       endif !status
+    !3) There should have been at least one chilling day in the counting period.  
+    !   this prevents tropical or warm climate plants that are "cold-deciduous"
+    !   from ever re-flushing after they have reached their maximum age (thus
+    !   preventing them from competing
+
+    if ( (currentSite%cstatus == 1 .or. currentSite%cstatus == 0) .and. &
+         (currentSite%grow_deg_days > gdd_threshold) .and. &
+         (currentSite%nchilldays >= 1)                    .and. &
+         (dayssincecleafoff > ED_val_phen_mindayson))  then
```
glemieux: @rgknox, why does the logic need to check the number of days leaves have been off against the minimum days on parameter?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -408,135 +408,185 @@ subroutine phenology( currentSite, bc_in )
...
+         (model_day_int > num_vegtemp_mem)      .and. &
+         (ncolddays > ED_val_phen_ncolddayslim) .and. &
+         (dayssincecleafon > ED_val_phen_mindayson) )then
+       
+       currentSite%grow_deg_days  = 0._r8          ! The equations for Botta et al
+                                                 ! are for calculations of 
+                                                 ! first flush, but if we dont
+                                                 ! clear this value, it will cause
+                                                 ! leaves to flush later in the year
+       currentSite%cstatus       = 1             ! alter status of site to 'leaves on'
```
glemieux: This is a comment typo correct?  Should it read `! alter status of site to 'leaves off'`?

biogeochem/EDPhysiologyMod.F90
rgknox: This is to prevent fickle plants from resprouting leaves if the local climate is in the process of getting cold, but has a spurious warm day.  However, since we do zero the growing degree days when leaves are dropped, this really doesn't seem necessary...

biogeochem/EDPhysiologyMod.F90
rgknox: yes! it should alter status to 'leaves off'

good catch, that is confusing

biogeochem/EDPhysiologyMod.F90
```diff
@@ -953,11 +1030,13 @@ subroutine seed_germination( currentSite, currentPatch )
        currentPatch%seed_germination(p) =  min(currentSite%seed_bank(p) * &
              EDPftvarcon_inst%germination_timescale(p),max_germination)     
        !set the germination only under the growing season...c.xu
-       if (EDPftvarcon_inst%season_decid(p) == itrue .and. currentSite%is_cold)then 
-             currentPatch%seed_germination(p) = 0.0_r8
+       if ( (EDPftvarcon_inst%season_decid(p) == 1) .and. &
```
glemieux: Why was this and new line 1037 change from `itrue`?

main/EDInitMod.F90
```diff
@@ -185,66 +188,61 @@ subroutine set_site_properties( nsites, sites)
...
+    integer  :: cstat      ! cold status phenology flag
     real(r8) :: GDD
-    logical  :: dstat
+    integer  :: dstat      ! drought status phenology flag
     real(r8) :: acc_NI
-    real(r8) :: watermem
-    integer  :: dleafoff
-    integer  :: dleafon
+    real(r8) :: watermem 
+    real(r8) :: cleafon    ! DOY for cold-decid leaf-on, initial guess
```
glemieux: Why are the `cleafoff` and `cleafon` values set as `real` whereas the `dleaf` states stayed integer?  It looks like based on the `cleafondate` there will be a conversion to integer down the line, correct?

main/FatesHistoryInterfaceMod.F90
```diff
@@ -1145,10 +1160,12 @@ subroutine set_history_var(this, vname, units, long, use_default, avgflag, vtype
                                            ! explict name (for fast reference during update)
                                            ! A zero is passed back when the variable is
                                            ! not used
+    
 
     ! locals
     integer :: ub1, lb1, ub2, lb2    ! Bounds for allocating the var
     integer :: ityp
+    integer :: nonfates              ! non-optional
```
glemieux: It looks like this doesn't currently have any specified use, correct?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -548,97 +598,129 @@ subroutine phenology( currentSite, bc_in )
...
-             currentSite%dleafoffdate = t       !record leaf on date           
-          endif
-       endif
+    ! LEAF OFF: DROUGHT DECIDUOUS DRYNESS - if the soil gets too dry, 
+    ! and the leaves have already been on a while... 
+
+    if ( (currentSite%dstatus == 2 .or. currentSite%dstatus == 3 ) .and. &
+         (model_day_int > numWaterMem) .and. &
+         (mean_10day_liqvol <= ED_val_phen_drought_threshold) .and. &
+         (dayssincedleafon > 100 ) ) then
```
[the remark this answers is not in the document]
rgknox: My take:
I think we should at least change 100 to be a constant parameter, with a descriptive name at the top of the file.
After that, if someone with more phenology background in a later pull request (or now if someone feels strongly) decides this should be a parameter in the user parameter file, then we can accommodate that.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -688,8 +770,8 @@ subroutine phenology_leafonoff(currentSite)
           ! for leaves. Time to signal flushing
 
           if (EDPftvarcon_inst%season_decid(ipft) == 1)then
-             if ( .not.currentSite%is_cold )then            !we have just moved to leaves being on . 
-                if (currentCohort%status_coh == leaves_off)then !Are the leaves currently off?        
+             if ( currentSite%cstatus > 1  )then                ! we have just moved to leaves being on .
```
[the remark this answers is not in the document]
rgknox: yes, I agree

biogeochem/EDPhysiologyMod.F90
```diff
@@ -709,12 +791,10 @@ subroutine phenology_leafonoff(currentSite)
                 endif !pft phenology
              endif ! growing season 
 
-             ! COLD LEAF OFF
-             ! The site level flag signifies that it is now too cold for 
-             ! deciduous leaves. Time to drop if they have not already
+             !COLD LEAF OFF
+             if (currentSite%cstatus == 1 .or. currentSite%cstatus == 0)then !past leaf drop day? Leaves still on tree?  
+                if (currentCohort%status_coh == 2)then ! leaves have not dropped
```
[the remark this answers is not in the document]
rgknox: hmm, this might be an artifact of a merge, I agree this should use leaves_on

biogeochem/EDPhysiologyMod.F90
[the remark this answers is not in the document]
glemieux: That sounds good.

biogeochem/EDPhysiologyMod.F90
rgknox: probably a merge resolution conflict, I probably updated this in a parallel pull request, but did not update this to use itrue in this PR.  it should use itrue.  And like you said earlier, should use cstatus == to different applicable conditions to avoid problems with newly added status.
perhaps we can use the "any" intrinsic function.

biogeochem/EDPhysiologyMod.F90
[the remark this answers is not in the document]
glemieux: Gotcha.  I'm guessing most of my down thread questions are related.

main/EDInitMod.F90
rgknox: correct, so the local is defined as the wrong type

main/FatesHistoryInterfaceMod.F90
rgknox: yes, I think that is an artifact and should be removed

main/FatesHistoryInterfaceMod.F90
```diff
@@ -3333,6 +3386,66 @@ subroutine define_history_vars(this, initialize_variables)
          avgflag='A', vtype=patch_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,    &
          ivar=ivar, initialize=initialize_variables, index = ih_area_treespread_pa)
 
+    call this%set_history_var(vname='SITE_COLD_STATUS', units='1,2', &
```
rgknox: this metadata should reflect the true range of values, same as drought status

rgknox:
testing: /gpfs/fs1/scratch/rgknox/clmed-tests/phentests-0519-v3-C27cdc37-Fe1d278f.fates.cheyenne.intel

rgknox:
resolving merge conflict, last test is now outdated, will also run a 50 year f10 smoke test

rgknox:
I think this one is pretty close to being ready.  I ran a multi-decade test, but then accidentally over-wrote everything with another test (which I forgot to re-name), but that simulation was well beyond 30 years at that point (on the f10 grid).  Hopefully the new simulations will be done by tomorrow.

rgknox:
multi-decade tests pass, running FATES test suite.  @glemieux are there any changes left on resolved in your review?

glemieux:
This looks all good.  Approving the changes.

rgknox:
Expected tests PASS: 
/gpfs/fs1/scratch/rgknox/clmed-tests/phentests-0519-v4-C27cdc37-F4b7293d.fates.cheyenne.intel

PR #484, opened 2019-02-21
added some comments in FatesHistoryInterfaceMod.F90

### LINKED ISSUE #479: Comment needed to explain 'scpf'
I was trying to figure out which of the new variables I needed to print out to get GPP through the vertical profile, wrt issue #472. Looking through the HistoryInterface, I came across this reference to 'scpf' at the beginning of the list of scpf variables, which doesn't have any explanation of what it means. 

https://github.com/NGEET/fates/blob/08bb1406bf4de99252cc14faa08a7d2eb651f140/main/FatesHistoryInterfaceMod.F90#L162

Obviously this is documented elsewhere in the code, but I thought that perhaps it is the case that I'm not the only person ever who will try to figure this out in this way. So, can we add a comment to explain what scpf means here?  The same goes for 'scls' further down. 

Apologies for the dunce-hat wearing nature of this comment.

ckoven:
Just some additional comments on how to add new history variables; no testing should be needed.

fixes #479.

### DIFF
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -45,8 +45,65 @@ module FatesHistoryInterfaceMod
   ! These variables hold the index of the history output structure so we don't
   ! have to constantly do name lookup when we want to populate the dataset
   ! These indices are set during "define_history_vars()" call to "set_history_var()"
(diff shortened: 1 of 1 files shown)


rgknox:
Did a build test, looks good, merging

PR #487, opened 2019-02-25
Increase nlevleaf to 30

rgknox:
Increased the number of maximum leaf layers to 30.  LAI was exceeding the maximum "binnable" LAI in a regional simulation.  Note while LAI of 20 is strangely large... this maximum value is the maximum possible LAI in a crown, and also needs to accommodate unusual circumstances that occur from things like parameter sensitivity analysis.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/main/EDTypesMod.F90
+++ b/main/EDTypesMod.F90
@@ -24,7 +24,7 @@ module EDTypesMod
                                                   ! to understory layers (all layers that
                                                   ! are not the top canopy layer)
 
-  integer, parameter :: nlevleaf = 20             ! number of leaf layers in canopy layer
+  integer, parameter :: nlevleaf = 30             ! number of leaf layers in canopy layer
   integer, parameter :: maxpft = 15               ! maximum number of PFTs allowed
                                                   ! the parameter file may determine that fewer
                                                   ! are used, but this helps allocate scratch


rgknox:
Tests, cheyenne intel, all pass:

/gpfs/fs1/scratch/rgknox/clmed-tests/fates.cheyenne.intel.maxlev30-Ce862304-Fa6305b1

walkeranthonyp:
One possible alternative to this if you want to save memory space could be to divorce the number of layers from LAI. e.g. you could always calculate photosynthesis in ten leaf layers whatever the LAI. If LAI is 2 then there is 0.2 m2 leaf per layer, or if LAI is 20 then there are 2 m2 per layer. 

Apologies if I misunderstood what this PR is doing.

rgknox:
@walkeranthonyp , I like where you are going with this.  Basically, sticking with the current method, instead of having a variable bin-size, has been a kick-the-can-down-the-road, punt-the-football, type thing.  But I like where you are going with this.  Perhaps we could use a minimum bin size as well, so that we would not be using lots of bins for small crown lai's.

walkeranthonyp:
@rgknox yeah that sounds like a good idea, adjust the number of calculated loops down if LAI is too low. If we wanted to get fancy I guess you could have variable leaf area per layer, starting small and getting larger as you get deeper in the canopy. Analogous to soil layering.

rgknox:
A 14 pft test maxed out the array size at 30 layers as well. I will bump this up to 40.  This happened in the 4th year. In the absence of a massive emergent, I'm wondering if these are recruits causing such high LAI...  Note that the testing parameter file may have some strange/poor choices for allometry and other parameters.

rgknox:
```
The leaf and stem are predicted for a cohort, maxed out the array size
502: lai:    30.0924666036124     
502: sai:   0.220300887876172     
502: lai+sai:    30.3127674914885     
502: nlevleaf,dinc_ed,nlevleaf*dinc_ed :          30   1.00000000000000     
502:   30.0000000000000     
502: ENDRUN:
502: ERROR in FatesAllometryMod.F90 at line 749  
```

ckoven:
do we just have weird parameters for one of the PFTs?  do you know which PFT it is that is crashing, and what are the slamax, slatop, allom_d2bl1, allom_d2ca_coefficient_min, and allom_d2ca_coefficient_max traits for that PFT? Maybe we should write these by default to this error message...

rgknox:
The culprit seems to be the C4 grass!  

dun dun dun

```
502: The leaf and stem are predicted for a cohort, maxed out the array size
502: lai:    30.0924666036124     
502: sai:   0.220300887876172     
502: lai+sai:    30.3127674914885     
502: nlevleaf,dinc_ed,nlevleaf*dinc_ed :          30   1.00000000000000     
502:   30.0000000000000     
502: pft:           14
502: n:    363.433244624115     
502: c_area:    23.5800202673186     
502: dbh:   0.168522618791017     
502: canopy_trim:   0.838410801344931     
502: target_bleaf:   4.764469143986919E-003
502: canopy layer:            1
502: ENDRUN:
502: ERROR in FatesAllometryMod.F90 at line 756         
```

I'm going to push its parameters through the allometry unit testing mechanism and see what I can find out

ckoven:
I went through and did some allometry calculations on the parameters for the C4 grass PFT and it looks to me like it shouldn't be having this problem.  But I think there is a clue in the error message, which is the reported ratio of SAI to LAI — which is far less than 0.1.  Since the SAI is calculated to equal (target LAI) * 0.1, that means that the actual bleaf is much larger than the target bleaf.  So something in the history of this cohort must have led it to have much more bleaf than its target value?

rosiealice:
One thing we did in CLM5 was massively increasing the turnover of C4 leaves
so their lifespan was something like 2 months. That helped bring the LAI
down a lot, although probably orthogonal to the target va actual LAI issue.

rgknox:
I added some diagnostics, do see if perhaps the vcmax scaling was having a weird effect on things, but that looks ok.

```
502: lai:    30.0924666036124     
502: sai:   0.220300887876172     
502: target_lai:    2.20300884593433     
502: lai+sai:    30.3127674914885     
502: nlevleaf,dinc_ed,nlevleaf*dinc_ed :          30   1.00000000000000     
502:   30.0000000000000     
502: pft:           14
502: n:    363.433244624115     
502: c_area:    23.5800202673186     
502: dbh:   0.168522618791017     
502: canopy_trim:   0.838410801344931     
502: target_bleaf:   4.764469143986919E-003
502: canopy layer:            1
502: canopy_tlai:   0.564248966310990       0.000000000000000E+000
502:  0.000000000000000E+000
502: vcmax25top:    78.0000000000000     
```

The target LAI is much below the actual.  So there is some disconnect happening.  My guess is either that fusion or growth are creating this problem.  If its fusion, I wondering if we have the following series of events:
1) Plants fuse, conserve crown area, but dbh is decreased in the process.
2) With a smaller DBH, next time crown area is called, it shrinks, but since leaf mass is purely prognostic, it does not shrink, thus large LAI

or if it is a growth thing, it could be imprecision

rgknox:
I want to note, that in these tests I am using the crown conservation that @ckoven has been working on. I'm running a test now that turns that back off.

rgknox:
After some thought, I think hypothesis 1 is not the problem.  @ckoven has coded up his new fusion routine to make dbh match the newly formed crown area, the crown area should not jump on subsequent calls.  I think what I will do is add more checks to report if LAI exceeds target LAI (by a non-trivial margin) and find where that first occurs.

rgknox:
closing this because the change to leaf layers will come in a different PR

Issue #500, opened 2019-03-07
Summary of Model Crashes - March 2019


rgknox:
Description of Crashes:

Crash 1: A cohort with a very large LAI of >30 is created, but this is larger than the array space we have to hold the leaf layers, so the simulation exits.  See issue #495

Crash 2: During the canopy promotion/demotion phase, the model is unable to shuffle cohorts around such that the canopy coverage areas have good agreement with the ground area.  See issue #494

----------------------------------------

Hypothesis of Factors Leading to Recent Errors:  There are no signs that the model has regressed or that new malformed code has been introduced.  Most likely, changes have been recently introduced that have stopped covering-up issues/errors that have been in the code.  

    These new changes are:

    Expanding the maximum canopy layers from 2 to 3
    Reducing the maximum number of leaf layers 
    Increasing the fusion tolerance and decreasing the max number of cohorts

---------------------------------------------

A set of fixes has been introduced that is undergoing testing that seems to be generating stable results:

Fix 1: A genuine bug has been found that was sending plants in the lowest canopy layer to the litter pool (killing them) when they should had been ignored

Fix 2: Charlie’s crown-area fusion conservation algorithm will make the promotion/demotion scheme “smoother”. Promotion/Demotion uses fusion, but if the total crown area changes during fusion, promotion/demotion will constantly be re-shuffling after being affected by this

Fix 3: Charlie and Ryan found that it was important to keep one of the grass biomass pools tied to allometry (like structure is with woody plants). With woody plants, if the structural biomass pool exceeds the maximum structural biomass associated with the tree's stature (dbh), we reset the dbh of the plant so that its maximum structural biomass (aka target biomass) matches the existing structural biomass.  The scenario where actual biomass is higher than the target is an artifact, and one that occurs due to numerical integration error, or because of non-linearity during fusion.  It is important to keep the biomass pools at, or below, the target biomass pool, so that the growth algorithm works correctly.  Grasses, don't have structural biomass, and up to now, we had not been forcing grass plants to have diameters that are tied to a biomass pool (like with woody plants).  In this fix, we tie grasses dbh to root biomass; thus if the root biomass exceeds the carrying capacity of the plant, we increase the dbh to match the root biomass.  As a result, now when fusion generates grass plants with abnormally high leaf biomass relative to its size, the plant's stature is adjusted (increased) and the the leaf area to crown area ratios will gradually reduce until they are within the expected range.  

[...]

PR #447, opened 2018-12-05
change cohort fusion to conserve crown area

ckoven:
This PR slightly revises the cohort fusion routine, so that it conserves crown area during the fusion timestep instead of conserving total stem diameter.

### Description:
Currently, when cohorts fuse, the resulting cohort tries to conserve total diameter of the two fusing cohorts.  I ran into an issue where, when cohorts get really big, the numerical errors in crown area conservation get big enough to crash things.  This got me thinking that the way we do cohort fusion might be leading to more numerical errors than needed, because of the lack of crown area conservation during cohort fusion, which is itself typically a part of canopy structuring.  What happens in a typical canopy structure loop is that we split the cohorts between the canopy and the understory, which in general this causes some cohort to get moved between strata.  But upon arrival in the new strata, the small cohort will typically find a similarly sized cohort already there, and will get fused.  But because we don't conserve crown are during fusion, the total post-fusion crown area has changed, and thus the canopy structure logic isn't what it thought it was.

So what this PR does is, when fusing cohorts, to calculate the DBH of the newly fused cohort as being that DBH which conserves total crown area and the dbh-crown area allometry.  further, in the event that dbh is to be incremented based on the biomass allometry to put it back on the biomass allometry, instead of immediately recalculating the dbh, it waits until the next growth step to do that.  that way, if the fusion happens during the canopy structure logic, it conserves crown area during that timestep, even if doing so causes it to be above its biomass allometry, and then resets the dbh to go back onto allometry at the start of the next growth timestep.

Fixes #442.

### Collaborators:
Discussions with @rgknox and @rosiealice 

### Expectation of Answer Changes:
This should be answer changing but at the level of noise, except for the weird edge cases that gave rise to #442.  Possibly by better conserving crown area during canopy structure it might reduce radiation numerical error.

### Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have updated the in-code documentation .AND. (the [technical note](https://github.com/NGEET/fates-docs) .OR. the wiki) accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/master/CONTRIBUTING.md) document.
- [ ] FATES PASS/FAIL regression tests were run
- [x] If answers were expected to change, evaluation was performed and provided

### Test Results:
Not formally tested yet.  hold until tests done.

PR #501, opened 2019-03-08
fixes addressing leaf area and demotion/promotion

rgknox:
### Description:

This pull request contains a suite of fixes aimed towards increasing model stability, particularly addressing instabilities related to (1) overflowing within crown leaf-area, and (2) demotion/promotion logic.

See issue #500 for details on the crashes.  Following that issue, here is a list of fixes. Fix 5 and fix 6 are new.

Fix 1: A genuine bug has been found that was sending plants in the lowest canopy layer to the litter pool (killing them) when they should had been ignored

https://github.com/NGEET/fates/pull/501/files#diff-14f02c3d54c6afd6502e04e81b2917a6R724

Also: the error checking tolerances for demotion/promotion precision were expanded to include absolute and relative error checks.

Fix 2: Charlie’s crown-area fusion conservation algorithm will make the promotion/demotion scheme “smoother”. Promotion/Demotion uses fusion, but if the total crown area changes during fusion, promotion/demotion will constantly be re-shuffling after being affected by this

Fix 3: Charlie and Ryan found that it was important to keep one of the grass biomass pools tied to allometry (like structure is with woody plants). With woody plants, if the structural biomass pool exceeds the maximum structural biomass associated with the tree's stature (dbh), we reset the dbh of the plant so that its maximum structural biomass (aka target biomass) matches the existing structural biomass. The scenario where actual biomass is higher than the target is an artifact, and one that occurs due to numerical integration error, or because of non-linearity during fusion. It is important to keep the biomass pools at, or below, the target biomass pool, so that the growth algorithm works correctly. Grasses, don't have structural biomass, and up to now, we had not been forcing grass plants to have diameters that are tied to a biomass pool (like with woody plants). In this fix, we tie grasses dbh to root biomass; thus if the root biomass exceeds the carrying capacity of the plant, we increase the dbh to match the root biomass. As a result, now when fusion generates grass plants with abnormally high leaf biomass relative to its size, the plant's stature is adjusted (increased) and the the leaf area to crown area ratios will gradually reduce until they are within the expected range.

This is mediated by calls through "ForceDBH()":

https://github.com/NGEET/fates/pull/501/files#diff-6403a5d68498c9a46a99461bedeb157aR2203
https://github.com/NGEET/fates/pull/501/files#diff-4bbd4c8a44000d218ae88b6b78485536R1003
https://github.com/NGEET/fates/pull/501/files#diff-4bbd4c8a44000d218ae88b6b78485536R1040
https://github.com/NGEET/fates/pull/501/files#diff-d180d6c64267eff0250516fea633a2e3R469
https://github.com/NGEET/fates/pull/501/files#diff-d180d6c64267eff0250516fea633a2e3R497

Fix 4: A modification of the demotion and promotion logic was introduced to avoid using an iterative solve, and thus reducing number of math operations. The iterative solve is necessary because the probabalistic demotion/promotion algorithm may impose that 1 or more cohorts demote >100% of their area on the first pass. The algorithm that re-scales these demotion/promotion fractions so they are all <=100% is the complicated part.

https://drive.google.com/file/d/1cyylwhtBzoRSPGIwS_Hr4Ma0vgwisLWz/view

https://github.com/NGEET/fates/pull/501/files#diff-14f02c3d54c6afd6502e04e81b2917a6R408

Fix 5: The patch termination algorithm was modified to at all costs fuse patches with areas less than 0.01 m2.  Excessively small patches were surviving that algorithm, and were creating numerical havoc in demotion/promotion and potentially radiation transfer.

https://github.com/rgknox/fates/blob/rgknox-demprob-merge-carea/biogeochem/EDPatchDynamicsMod.F90#L1865

Fix 6: The maximum number of canopy layers was reduced back to 2.  Errors in the radiation transfer algorithm were being generated, reducing the layering back to 2 removed this errors. Understanding these erorrs is beyond the scope of these fixes and should be targeted as a more comprehensive analysis.

### Collaborators:

@ckoven @jkshuman @peterhor 

Yes, answers should change.  The crown area conservation in the fusion routine, as well as the change in the number of canopy layers should have non-trivial impacts on the simulation.  The rest of the fixes should have at least round-off changes.

### Checklist:

### Test Results:

Long-term tests using the 14 pft default file on the f45 grid are on-going and showing promise.

Standard FATES test suite: TBD

FATES baseline hash-tag:

Test Output:


rgknox:
long term tests are on-going. A good amount of new code has temporary diagnostics I was using to track down bugs, which still need to be removed.

rgknox:
A 120 year test with 14 pfts on the 4x5 grid, and a 100 year base-comparison test has completed. Will follow up with some diagnostics.

rgknox:
Here is a comparison with tag 1.21 (which is called "base"):

https://drive.google.com/file/d/1YiMMxn5bTP5OEIyoOqrR1xXj2rLyA4vc/view?usp=sharing

Both are spin-up type simulations at BCI, both use the default parameter set.  Note the difference in accumulated biomass is significant.  I ran similar tests on @ckoven 's PR #447, which had gotten side-tracked, but preliminary results showed significant differences in biomass accumulation as well.

I'm going to run a few more tests to compare with tag 1.21.  I'd like to see how this branch compares when I flip the switch back to basal-area (diameter) conservation during fusion.  I'd also like re-run the comparison of Charlie's branch from #447 with tag 1.21.

ckoven:
@rgknox it would be good to look at the mean `GROWTHFLUX_FUSION_SCPF` field for the last 20 years or so between these simulations.  Do you have that in your output?

rgknox:
Ahh, I do not.  I will add it to the new runs though. Any other diagnostics requested?

ckoven:
It seems like we need to understand why the biomass is larger, and to exclude the hypothesis that it is something other than that the cohorts end up having a larger DBH after fusion in the new regime than the old one.  So I guess I'd want to look at:
```
'DDBH_CANOPY_SCPF','DDBH_UNDERSTORY_SCPF','GROWTHFLUX_SCPF','GROWTHFLUX_FUSION_SCPF','NPLANT_CANOPY_SCPF','NPLANT_UNDERSTORY_SCPF','MORTALITY_CANOPY_SCPF','MORTALITY_UNDERSTORY_SCPF'
```

rosiealice:
This is still a fix that rolls in the fix which doesn't send so many cohorts erroneously to the litter pool, right? That seems like it might have been a cause of the higher biomass?

rgknox:
@rosiealice , that is true. 
Here is a comparison of the branch in this PR (test) against Charlie's branch (base) which contains only the crown fusion feature merged in with master.

https://drive.google.com/file/d/1b4qY15kmDCuY3X5KTkBIdvPgFS9C4L7d/view?usp=sharing

Most of the differences in biomass with tag 1.21 seem to be associated with the crown fusion.

The differences in this comparison could be from a few things: 1) reverting back to a maximum of 2 canopy layers, 2) changes in the canopy demotion/promotion logic, 3) the fix @rosiealice mentioned that was sending cohorts to litter

I will take a look at the growth flux diagnostics and post that next

rgknox:
Here is a comparison of the Charlie's branch with crown area conservation during fusion (merged with master) (test), against tag 1.21 (base).

https://drive.google.com/file/d/1tRZJF9KFlkJna4b4SF7cEmmcXAOf_p3u/view?usp=sharing

ckoven:
@rgknox and I just looked at this in greater detail.  there's clearly a bug in this PR (which I introduced) somewhere, which we need to sort out before merging.  So, just to say don't stare too hard at these figures trying to make sense of them, or not yet anyway.

rgknox:
I think we are back on track. Looks like we were missing a call to re-calculate crown areas prior to fusing them together.  The following link is a comparison of the crown fusion feature branch with tag 1.21, the differences now in terms of biomass and leaf area trajectories are pretty similar. :)

https://drive.google.com/file/d/1QwL-B1C-7Qq-sz_5TAodMCzzQdjJ24HF/view?usp=sharing

ckoven:
@rgknox do we need to do the call to StructureResetOfDH here?: https://github.com/NGEET/fates/pull/501/commits/668e3de6f4d1c43d9b0a677f4a12d9bc9b0e8f14#diff-4bbd4c8a44000d218ae88b6b78485536R984  My concern is that in doing so, we don't actually conserve crown area once we do that. Or was that the source of the bug?

rgknox:
My thinking was, if at that point we have aborted our attempts to conserve crown area and are now conserving DBH, we might as well be consistent with how we do things in the alternative (DBH conserving) case.  What are your thoughts? Are you concerned that maybe it will shift things around during fusion calls to promotion/demotion?

ckoven:
@rgknox sorry, I misread your code.  I see your logic now and agree with it.

rgknox:
Here is a comparison of the current branch with tag 1.21 at bci:
https://drive.google.com/file/d/1w1itcOdGJAAdXD0F9GxVfkVC14Kn01ch/view?usp=sharing

rgknox:
Here is a summary of the global biomass partitions after 46 years, using the 14 pft case.  Notes:  (1) This simulation has spit-fire turned on.  (2) I put almost no emphasis on parameterization, I just grabbed the default file, and (3), as you will see the c4 grasses are so win, and ultimately seem to suppress tree grown in areas they shouldn't

https://drive.google.com/file/d/1ZYoFykcD73BvY1pioDEhQlKLJSm4A4oB/view?usp=sharing

I'd like to try some simple simulations that either remove c4 grasses, or alternatively have some combination of making their recruit sizes smaller and enabling a small background seed-rain

rgknox:
Here is a new comparison. Global simulations, 16 years into simulation.  

Base: master branch with 3 small tweaks (or else it won't complete), 2 canopy layers, 40 leaf layers, and the bug fix to erroneously terminating cohorts in the lowest layer during fusion.

Test: the current branch, after merging in the master (thus including the new photosynthesis parameters)

https://drive.google.com/file/d/1_VzK8sjFbTSmCn4Wu_OkLkbmUiF2XQUr/view?usp=sharing

My take-aways:  
1) Current master in this test configuration shows similar-ish patterns to leaf and biomass, as the test branch.
2) Although, re 1), the test branch does include answer changing code (ie demotion logic, and crown fusion), which features added specifically to overcome crashes and were an important part of these changes
3) I think the default parameterization, or the set of PFTs used needs to be tweaked.  It would be nice to generate a global demographic of biomass and leaf area that is more representative of reality, which would really help testing

ckoven:
thanks @rgknox.  glad that the code changes here don't seem too have much of a qualitative effect on the outcomes, and that the default-pft-file C4 grass world takeover is a separate problem.  could you assign code reviewers to this PR so that we can get it merged soon?

rgknox:
All expected PASS:
/glade/u/home/rgknox/clmed-tests/fates.cheyenne.intel.demprom-carea-fixes-Ccb5be8e-Fd6c81e0

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -469,98 +485,150 @@ subroutine DemoteFromLayer(currentSite,currentPatch,i_lyr)
...
                if(currentCohort%canopy_layer  ==  i_lyr) then 
-                  currentCohort%excl_weight = demote_area*currentCohort%excl_weight/sumweights
-                  if( currentCohort%excl_weight > currentCohort%c_area ) then
-                     remainder_area = remainder_area + currentCohort%excl_weight-currentCohort%c_area
-                     currentCohort%excl_weight = currentCohort%c_area
-                  else
-                     sumweights_old = sumweights_old + currentCohort%excl_weight
-                  end if
+
+                  currentCohort%excl_weight = currentCohort%excl_weight/sumweights
```
glemieux: Do I understand correctly here that the code will still run through the normal (stochastic) case first?  Otherwise ```sumweights``` would start off as ```0.0_r8``` and result in a divide-by-zero?

glemieux: I think I answered my own question: this has to be true given that the ```currentCohort%excl_weight``` is also computed in that section as well.

glemieux: I see; we need to run through the existing weights first to get their sum and then run back through them again to normalize.

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -469,98 +485,150 @@ subroutine DemoteFromLayer(currentSite,currentPatch,i_lyr)
...
+
+               ! Trivial case, all of the demotion fractions are less than 1.
+
                currentCohort => currentPatch%tallest
-               do while (associated(currentCohort))      
-                  if(currentCohort%canopy_layer == i_lyr)then
-                     if ( currentCohort%excl_weight < (currentCohort%c_area-nearzero) ) then
+               do while (associated(currentCohort))
+                  if(currentCohort%canopy_layer  ==  i_lyr) then 
+                     currentCohort%excl_weight = currentCohort%c_area * currentCohort%excl_weight * scale_factor
```
glemieux: Note to self: ```scale_factor``` includes ```c_area``` so needs to be cancelled out.

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -469,98 +485,150 @@ subroutine DemoteFromLayer(currentSite,currentPatch,i_lyr)
...
 
-                        ! Update this cohorts exclusion
-                        currentCohort%excl_weight = currentCohort%excl_weight + cc_loss
-                        
-                        ! Update the sum of weights for cohorts where exceedance can
-                        ! still be spread to
-                        if( currentCohort%excl_weight < (currentCohort%c_area-nearzero) ) then
-                           sumweights = sumweights + currentCohort%excl_weight
+                     currentCohort%excl_weight = currentCohort%c_area * &
+                           (currentCohort%excl_weight * scale_factor_min + &
```
glemieux: @rgknox I don't think you need to multiply by the area here since ```scale_factor_min``` doesn't include the area like ```scale_factor``` correct?

glemieux: I assumed that this should match the technical paper exactly in this case, but I'm thinking that you meant to do this since new line 610 assumes the area is incorporated?

biogeochem/EDCanopyStructureMod.F90
glemieux: Update: See comment on line 574 below.  The intent appears to be to compute the excluded area immediately here, which makes sense looking at the debug below and the ```ed_cohort_type``` comments.

biogeochem/EDCanopyStructureMod.F90
rgknox: ok, I'm looking through this now as well and double checking too

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -906,110 +984,164 @@ subroutine PromoteIntoLayer(currentSite,currentPatch,i_lyr)
...
-                  exceedance_counter = exceedance_counter + 1
-                  if( exceedance_counter > 100 ) then
-                     write(fates_log(),*) 'It is taking a long time to distribute promotion exceedance'
-                     write(fates_log(),*) '(ie greater than c_area) to neighbors... exiting'
-                     call endrun(msg=errMsg(sourcefile, __LINE__))
-                  end if
+               else
                   
-               end do
+                  ! Non-trivial case, at least 1 cohort's demotion
```
glemieux: Comment typo: should read promotion, not demotion, correct?

rgknox: yup, good catch

glemieux:
Finished reviewing the promotion/demotion subroutines; looks good.  Per our Gchat conversation, my brain itched to make the repeated calls to:
```
do while (associated(currentCohort))
               if(currentCohort%canopy_layer  ==  i_lyr) then 
```
into a callable subroutine.  @rgknox noted that;
> Typically things in-line are more efficient than subroutines in fortran, but this code is not called every second, so improvements for clarity that slows code is an acceptable trade-off here

I'm assuming that temporarily containing the whole of a cohort exclusion or promotion area in a vector  wouldn't be (memory) efficient either, correct?  My thought was to allow for using `min()` instead of an 'if' loop to find the `scale_factor_min`.

For posterity and potential future reference [here's the link the the Jupyter notebook](https://github.com/glemieux/fates-scratch/blob/master/Demotion/ProbablisticDemotionPromotion.ipynb) where I worked through the demotion technical paper by hand to get a feel for it.

rgknox:
running final test, results permitting, will integrate

rgknox:
all expected PASS:
/gpfs/fs1/scratch/rgknox/clmed-tests/fates.cheyenne.intel.demprom-vfinal-Cbc17668-F459f09e

PR #515, opened 2019-03-28
changed ncvarsort to use scipy netcdf

ckoven:
A really minor PR to update the ncvarsort script.  Only changes python code so not formally tested. Based on request from @rgknox that we consolidate our python netcdf dependencies to only use the scipy library.


rgknox:
thanks @ckoven , merging

PR #521, opened 2019-04-12
Moved structure reset code out of parteh and into edcohortdynamics

rgknox:
### Description:

The growth algorithm performs a potential "reset" of dbh if a plant's structural biomass is greater than the maximum structural biomass for its current stature (DBH).  For grasses, it uses leaf biomass instead of structure.  This is necessary for the plant to stay on the allometric curve.   

This changeset is a refactor of the process, that moves this code outside of the specific allocation hypothesis, and is called in its own subroutine, immediately before the growth (parteh) call.  By doing this we 1) avoid calling the same exact code for different parteh hypotheses (which is a no no) and 2) allows us to track the changes in diameter (%dbhdt) that are a result of the reset and a veritable result of accumulating biomass.

### Collaborators:

@ckoven 

This should only change any diagnostics that use the cohort rate of change of diameter (%dbhdt).

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


rgknox:
expected PASS (b4b): 

/gpfs/fs1/scratch/rgknox/clmed-tests/growth-resets.fates.cheyenne.intel

main/EDMainMod.F90
```diff
@@ -270,7 +271,9 @@ subroutine ed_integrate_state_variables(currentSite, bc_in )
     real(r8) :: dbh_old               ! dbh of plant before daily PRT [cm]
     real(r8) :: hite_old              ! height of plant before daily PRT [m]
     real(r8) :: leaf_c
-    
+    real(r8) :: delta_dbh             ! correction for dbh
```
ckoven: should we do anything with the info in delta_dbh and delta_hite, (i.e. send one of them to a history variable) or is that overkill?

rgknox: I just left them there in case someone wanted to use them for diagnostics in the future, as a convenience.  I have no plans of adding them to the history file personally.

ckoven: ok, works for me.  it's nice to have that there in case we ever suspect something weird is going on with the structure resets.

Issue #489, opened 2019-02-25
bole harvest litter logic


rgknox:
We have 4 coarse woody debris pools.  The last pool is for the trunk.  In the code that handles distribution of tree mortality into the different litter pools, we have some logic that handles harvesting differently for trunk and non-trunk components:

https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L1229

However, it looks like the logic is backwards.  When c==ncwd, we should be making provisions for exporting harvested wood, and when c/=ncwd, we should be operating normaly.

@huangmy @ckoven

rgknox:
I think the litter generation is fine, its just the trunk product and the exported carbon should be moved to the first condition.

huangmy:
@rgknox @ckoven 
Just to make sure you capture my comments here. Ryan is correct that the trunk product and the exported carbon should be moved to the first condition. In addition, Michael (Keller) suggested some numbers to use to apply a discount to the actually exported carbon and add the leftover to the large CWD pool ranging from 1% to 3.8% for RIL and CL respectively as a result of limitation in length of transportation vehicles and/or poor planning (trees could be felled but not found). So feel free to let me know if you want to submit a pull request on this, or I can submit it as part of an new issue.

rgknox:
@huangmy , if you would like to submit a PR, that would be great.  Otherwise, I will include this in my nutrient refactors, but the time-table on that is not short term.

huangmy:
@rgknox sure, I will include these in a PR then, likely sometime next week. Thanks.

rosiealice:
Just concurring that the litter_in logic is indeed backwards to my mind.  The way that SPITFIRE handle the trunk index (which I had thought would avoid having the hardwire the C==NSC condition) is a bit diffiult/confusion, in that while we have four NSC classes, spitfire has six fuel classes, and so the indexing in SPITFIRE is not the same as for when we are thinking aboutthe NSC pools (as in, the trunk index in SPITFIRE is 5, not 4.

https://github.com/NGEET/fates/blob/64b5affb69b681c923a6575457eda2785f5b5789/main/EDTypesMod.F90#L134

I can't think of a good way out of this right now, other than to maybe re-order the fiel catagories so that we have the NSC's first and then the live and dead grasses as indices 5 and 6. (right now they are 1 and 6). That way, at least the NSC indicies in SPITFIRE would align with those in the logging and litter code.

[...]

PR #523, opened 2019-04-23
Sf fuels refactor + bole harvest litter logic fix/export frac

huangmy:
### Description:

This pull request aims to refactor the SPITFIRE and wood harvest logic as follows:

Reorder the fuel classes in SPITFIRE to be consistent with that in other parts of FATES;
Fix a reversed logic in treating CWD contributed by directly logged trees
Add an export fraction to account for transportation losses when shipping the boles offsite

### Collaborators:

@jkshuman 
@rgknox 
@rosiealice 
@ckoven 

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -59,6 +59,9 @@ module EDLoggingMortalityMod
 
    character(len=*), parameter, private :: sourcefile = &
          __FILE__
+
+
+   real(r8), public, parameter :: logging_export_frac = 0.8_r8
    
    public :: LoggingMortality_frac
    public :: logging_litter_fluxes
@@ -170,6 +173,8 @@ contains
       real(r8), parameter   :: adjustment = 1.0 ! adjustment for mortality rates
  
       if (logging_time) then 
+
+         
          if(EDPftvarcon_inst%woody(pft_i) == 1)then ! only set logging rates for trees
 
             ! Pass logging rates to cohort level 
@@ -194,6 +199,9 @@ contains
             ! will be applied via "understory_death" via the disturbance algorithm
             if (canopy_layer .eq. 1) then
                lmort_collateral = logging_collateral_frac * adjustment
+            else
+               lmort_collateral = 0._r8
+               l_degrad         = l_degrad + logging_collateral_frac * adjustment
             endif
 
          else
@@ -202,6 +210,7 @@ contains
             lmort_infra      = 0.0_r8
             l_degrad         = 0.0_r8
          end if
+
       else 
          lmort_direct    = 0.0_r8
          lmort_collateral = 0.0_r8
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -197,11 +197,14 @@ contains
 
             ! Collateral damage to smaller plants below the canopy layer
             ! will be applied via "understory_death" via the disturbance algorithm
+            ! Important: Degredation rates really only have an impact when
+            ! applied to the canopy layer. So we don't add to degredation
+            ! for collateral damage, even understory collateral damage.
+
             if (canopy_layer .eq. 1) then
                lmort_collateral = logging_collateral_frac * adjustment
             else
                lmort_collateral = 0._r8
-               l_degrad         = l_degrad + logging_collateral_frac * adjustment
             endif
 
          else
@@ -318,10 +321,19 @@ contains
                   (currentCohort%lmort_collateral + currentCohort%lmort_infra)
 
          else
+
+            ! This routine is only called during disturbance.  The litter
+            ! fluxes from non-disturbance generating mortality are 
+            ! handled in EDPhysiology.  Disturbance generating mortality
+            ! are those cohorts in the top canopy layer, or those
+            ! plants that were impacted. Thus, no direct dead can occur
+            ! here, and indirect are impacts.
+
             if(EDPftvarcon_inst%woody(currentCohort%pft) == 1)then
                direct_dead   = 0.0_r8
-               indirect_dead = logging_coll_under_frac * currentCohort%n * &
-                     (patch_site_areadis/currentPatch%area)  !kgC/site/day
+               indirect_dead = logging_coll_under_frac * &
+                    (1._r8-currentPatch%fract_ldist_not_harvested) * currentCohort%n * &
+                    (patch_site_areadis/currentPatch%area)   !kgC/site/day
             else
                ! If the cohort of interest is grass, it will not experience
                ! any mortality associated with the logging disturbance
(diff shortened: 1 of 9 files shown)


huangmy:
@rgknox, @ckoven, @rosiealice  : this is a pull request to address https://github.com/NGEET/fates/issues/489. Please take a look and let Jackie and me know your comments?

rgknox:
@huangmy and @jkshuman, does the index in this burned_frac_litter need to be changed?

https://github.com/NGEET/fates/blob/master/biogeochem/EDPatchDynamicsMod.F90#L1088

jkshuman:
good catch @rgknox . Yes, that should be updated to just "c" as the refactor makes the "+1" unnecessary.

rgknox:
will submit a PR to @huangmy 's branch with some small changes. 
1) will fix that (c+1) index
2) will move the export_fraction to be a local constant instead of parameter from file, and will stage the additional parameter in the parameter file to be added along with a group of other additions in a later PR

huangmy:
Sounds good. Thanks!

rgknox:
I'm going to run a multi-decade 10x10 test on this branch (with logging events on) to make sure that there are no lingering carbon balance errors.  Will check back in.

rgknox:
@jkshuman and @huangmy : At the time of writing this I am still generating mass conservation errors following logging events.  I see no evidence suggesting that the new changes are causing these imbalances, from what I can tell these have been there for a while.
Would you like to :
a) proceed with integrating these changes (since they contain non-logging updates as well), thus fixing the mass imbalance in a later PR, or
b) wait and see if the mass balance errors can be resolved and pushed through with this set of changes

jkshuman:
@rgknox I think digging in and fixing the mass conservation errors before incorporating these changes is best plan.

rgknox:
Thinking more on this, I agree. Lets get this fixed before we proceed. I will look into this as well.
Labeling as "not ready".

rgknox:
Trying to solve this remaining carbon balance error.  I believe the non-disturbance side of things is ok. I think the problem is a disconnect between litter generation and the number of trees killed during the logging disturbance portion of the code.
Here is a potential disconnect:

In this part of the code we are calculating the number of dead in lower canopy layers due to impacts from falling trees above (from logging), this number is just used to generate mortal litter flux:
https://github.com/NGEET/fates/blob/12a6c7ea35b9cd390051d1590ebb49420d3fc0fd/biogeochem/EDLoggingMortalityMod.F90#L323

Here is where we actually reduce the number density of understory trees due to impacts:
https://github.com/NGEET/fates/blob/12a6c7ea35b9cd390051d1590ebb49420d3fc0fd/biogeochem/EDPatchDynamicsMod.F90#L761

In the prior, we do not seem to be scaling the understory litter generation by this "currentPatch%fract_ldist_not_harvested" scaler.

I also think there is a disconnect in calculating that fract_ldist_not_harvested variable, as we are not filtering out grasses.

rgknox:
Another thing I am grappling with:
currentPatch%fract_ldist_not_harvested is the fraction of the newly disturbed patch that is associated with "degredation", but not actually area opened by falling trees right?  

If this value is zero, then we assume all the area in the newly created patch is associated with falling trees, that would make the understory impact mortality rate increase (to the full value of the impact parameter).  If this value was 1, that means nothing was actually falling on the trees below, and I would assume that the understory impact morality would decrease (to zero).  But, we are actually doing the opposite here:
 https://github.com/NGEET/fates/blob/12a6c7ea35b9cd390051d1590ebb49420d3fc0fd/biogeochem/EDPatchDynamicsMod.F90#L762

I think that the currentPatch%fract_ldist_not_harvested  just needs to be encased with a (1-currentPatch%fract_ldist_not_harvested ).

huangmy:
@rgknox @ckoven @jkshuman : I just merged Ryan's changes to fix the mass balance problem to the branch. Please test and review again.

rgknox:
great!  I think Cheyenne is down for maintenance this week though... It will be at least a few days before we can final tests completed and integrate.

rgknox:
@jkshuman and @huangmy , I'm having trouble getting comparable output on this branch against previous test output.  Aside from changes to the logging module, is this PR answer changing?

huangmy:
@rgknox for the changes in the logging module, the fix to the reversed bole harvest logic should change the truck product only. However, the export fraction (set at 0.9 right now) would allow more boles left on site as CWDs. An easy test would be to set export fraction to 1.0 to see if it could match previous outputs.

I will let @jkshuman comment on the sf fuel refactoring.

jkshuman:
@rgknox it looks like there is a mistake in the parameter ordering for the refactor for max_decomp. It was correct on my commit https://github.com/NGEET/fates/pull/523/commits/f95b44874d23b249d6ca221fd19fb45477d0ff53
checking over the params now.

huangmy:
@jkshuman I might have made a mistake when enter them manually. Thanks for checking.

jkshuman:
@rgknox not sure if I am reading this wrong. If I look at the files changes, it looks like it was updated correctly, but if I look at mid level commits it was correct and then wrong. Please confirm the correct order as:
fates_max_decomp = 0.52, 0.383, 0.383, 0.19, 1, 999 ;

I do not expect it to be answer changing.

jkshuman:
@huangmy I made the same mistake myself, so the reason for my multiple commits to the param file. pesky decomp.

rgknox:
I can confirm that is the order in the parameter file.

huangmy:
@jkshuman I feel much better now :-).

rgknox:
@jkshuman found the problem that was preventing b4b comparisons. I now get b4b comparisons against the baseline.

Note that in the tests,  there is a false positive, where  FUEL_MOISTURE_NFSC reports an error.  As far as I can tell, since this variable is evaluated over the fire fuel dimension, and since those positions have changed, the values in the different positions should now change.

```
FUEL_MOISTURE_NFSC   (lon,lat,fates_levfuel,time)  t_index =      2     2
       5700    19872  (    67,     5,     5,     1) (     1,     1,     1,     1) (    67,     5,     5,     1) (    11,    20,     5,     1)
                8796   2.473892450332642E+00          0.000000000000000E+00 7.2E-01  2.473892450332642E+00 8.1E-02  2.470947265625000E+00
                8796   2.473892450332642E+00          0.000000000000000E+00          1.752644181251526E+00          1.750557541847229E+00
               19872  (    67,     5,     1,     1) (     1,     1,     1,     1)
          avg abs field values:    1.270223498344421E+00    rms diff: 2.7E-01   avg rel diff(npos):  8.1E-02
                                   1.270230531692505E+00                        avg decimal digits(ndif):  1.0 worst:  0.5

```

https://github.com/NGEET/fates/blob/master/main/FatesHistoryInterfaceMod.F90#L2256

Test directory:
/gpfs/fs1/scratch/rgknox/clmed-tests/SF-refactors-merged-1.26.3-v2-C6767148-Fe90ae2f.fates.cheyenne.intel/

rgknox:
@glemieux , I will submit a PR to @huangmy 's branch with the fix, once this is updated in this branch, I recommend we merge

glemieux:
Roger that @rgknox

PR #525, opened 2019-04-25
fixes to hydraulics

rgknox:
### Description:

This changeset includes some important fixes to hydaulics.  These fixes were implemented by @xuchongang . 
(Since these are high-priority fixes for the hydro team, and @xuchongang is traveling, @rgknox is acting as submitter.)

1) initialize the plant with more water (not in equilibrium with very dry soil);
2) reduce conductance for soil-root interface for water loss compared to water uptake;
3) greatly increased the hard-wired cap on maximum stomatal resistance (e.g. decreases minimum conductance)

### Collaborators:

@pollybuotte analyzed, tested and identified errors, @rgknox and @JunyanDing identified errors and provided support.

The change in maximum resistance should change ALL fates simulations.  

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:


rgknox:
It seems we have two slightly different ways of calculating minimum leaf conductance "bbb" in this routine.  

Here we prevent it from going lower than 1.0 umol H2O/m**2/s
https://github.com/NGEET/fates/blob/3b15841abf8b03442d1833cdc0dec8c8d3ca8cd0/biogeophys/FatesPlantRespPhotosynthMod.F90#L400

But in these calculations which are only active at night when btran>0, there is no lower bound:

https://github.com/NGEET/fates/blob/3b15841abf8b03442d1833cdc0dec8c8d3ca8cd0/biogeophys/FatesPlantRespPhotosynthMod.F90#L972

My take is that 1 umol H20/m**2/s is pretty small right?   The defualt cuticular conductances in our parameter file are 10,000 and 40,000.  Is there any reason why we aren't using the BBB calculated in line 400, that is passed as an argument, not to be used in line 972?

@xuchongang @pollybuotte @JunyanDing

pollybuotte:
@rgknox I have not gone through the code yet, but given Chonggang's description of his changes...

"There are also other place where the minimum conductance is capped at 1.0 umol/m2/s, mainly for reasonable photosynthesis; but this can also cause a higher stomata conductance for water loss. I have also changed those to be determined by the cuticular conductance inputs and btran."

I would have expected all the places where bbb is calculated to now be based on btran and cuticular conductance.

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -968,6 +983,7 @@ subroutine LeafLayerPhotosynthesis(f_sun_lsl,         &  ! in
 
         !is there leaf area? - (NV can be larger than 0 with only stem area if deciduous)
         if ( laisun_lsl + laisha_lsl > 0._r8 ) then 
+	  if(bbb > 1.0_r8)then !only if stomata open larger than cuticular conductance
```
rgknox: @xuchongang , I don't follow the connection between the logic and its comment.   Can you help me clear this up?

The code is seeing if the minimum leaf (cuticular) conductance (bbb) is greater than 1.  But the comment implies that its comparing the stomata opening against cuticular conductance.... 

Since 1 umol/m2/s is very very small, it seems that this comparison is really about seeing if btran>0 ? I must be missing something.

rgknox: @xuchongang and I checked in on this, and agreed that we can remove this logic: "I basically try to avoid the cap by 1 umol/m2/s and avoid the cases of unstable solutions of gs. I agree that 1 umol/m2/s is very very small and this part of change is not critical for the bug fix. I understand your confusion and I think we can leave the code as it is to avoid this confusion."

rgknox:
All, please see the following additional changes I'm planning to add:
https://github.com/rgknox/fates/pull/2

rgknox:
We have identified that there are still non-trivial mass balance errors generated during the 1D-solve portion of hydraulics.  I re-labeled this as "not-ready", and we will continue to evaluate this more after the DOE PI meeting.

rgknox:
@JunyanDing , can you take a look at the changes I'm staging to be added in https://github.com/rgknox/fates/pull/2.   Do these changes look like they will fix this inconsistency?

rgknox:
35 year comparison at BCI (the changes to rsmax0 made this non b4b change):

https://drive.google.com/file/d/13A5UJilQ4oP1awBPjrRDnx5o4xl9MAy2/view?usp=sharing

50 year f10 smoke test: 

/gpfs/fs1/scratch/rgknox/fates-clm-tests/hydro-fixes-f10_f10_mg37/run

Regression tests (not b4b, should not be):

/gpfs/fs1/scratch/rgknox/clmed-tests/hydro-fix-C6767148-F5dbff93.fates.cheyenne.intel

PR #526, opened 2019-04-25
added ability to make identical cohorts for multiple PFTs from inventory initialization

ckoven:
For experiments where one wants to compete multiple PFTs against each other, its nice to be able to start from an inventory initialization file with identical starting size distributions for each PFT.  This PR ads that capability.  it is triggered by setting the PFT number to zero in the .css file that is used for the inventory initializations.

### Collaborators:
@rgknox 

### Expectation of Answer Changes:
None

### Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have updated the in-code documentation .AND. (the [technical note](https://github.com/NGEET/fates-docs) .OR. the wiki) accordingly.
- [ ] I have read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/master/CONTRIBUTING.md) document.
- [ ] FATES PASS/FAIL regression tests were run
- [ ] If answers were expected to change, evaluation was performed and provided

### Test Results:
Not yet tested.

CTSM (or) E3SM (specify which) test hash-tag:

CTSM (or) E3SM (specify which) baseline hash-tag:

FATES baseline hash-tag:

Test Output:

PR #532, opened 2019-05-02
inventory output capability

rgknox:
### Description:

This set of changes adds a simple method of writing the core demographic information to a text output file.  Currently, the output format is the same as the inventory initialization format we use "type 1", which is compatible with ED input formats as well.  There are no controls on the file name convention, it will generate 1 file couplet for each site in a simulation, appending a generic name with the lat/long of the site to it.  Turning this functionality on requires toggling a hard-coded switch.  
If there is sufficient interest, I would be happy to add this switch, or potentially, file name prefixes to the input parameter file. 

### Collaborators:

@ckoven 

This should NOT change answers.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:


main/FatesInventoryInitMod.F90
```diff
@@ -69,6 +69,9 @@ module FatesInventoryInitMod
                                                            ! defined in model memory and a physical
                                                            ! site listed in the file
 
+   logical, parameter :: do_inventory_out = .true.
```
ckoven: is the plan to test the PR with this `do_inventory_out` switch set as true to confirm correct behavior, but then change it back to false before merging?

rgknox: It should be changed back to false before pulling.  We you able to use this output, did it work correctly for you @ckoven ?

ckoven: yes it is working correctly for me at present

ckoven:
Hi All,
Just to note here for reference that using this feature is nice for ensembles, because it allows one to use inventory initialization without requiring the step where you read in a census file with possibly many thousands of trees assigning each tree to its own cohort, and then letting the cohort fusion routine condense down to a manageable number of cohorts.  I had been running into memory issues in ensembles when running 1 instance per CPU, and I assumed I'd run into them even more bigly when I tried to initialize to large numbers of PFTs (via #526).  But if I just run a single instance through this code, and then output the cohorts themselves as an inventory file, then I don't run into the same memory problems when I use the resulting inventory file to initialize large numbers of instances and/or PFTs.

main/FatesInventoryInitMod.F90
```diff
@@ -940,4 +955,92 @@ subroutine set_inventory_edcohort_type1(csite,bc_in,css_file_unit,npatches, &
...
+       ! Generate pss/css file name based on the location of the site
+       ilat_int = floor(currentSite%lat)
+       ilat_dec = int(100000*(currentSite%lat - real(ilat_int,r8)))
+       ilon_int = floor(currentSite%lon)
+       ilon_dec = int(100000*(currentSite%lon - real(ilon_int,r8)))
+       
+       write(pss_name_out,'(A8,i3.3,A,i5.5,A2,i3.3,A,i5.5,A5)') &
+             'pss_out_',ilat_int,'.',ilat_dec,'N_',ilon_int,'.',ilon_dec,'E.txt'
+       write(css_name_out,'(A8,i3.3,A,i5.5,A2,i3.3,A,i5.5,A5)') &
+             'css_out_',ilat_int,'.',ilat_dec,'N_',ilon_int,'.',ilon_dec,'E.txt'
```
glemieux: @rgknox is `currenSite` lat and lon range from 0 to 360 or -180 to 180?

rgknox: hmmm, that is a good question.  I suppose -100 and greater will exceed the string size on ilon_int.  I can bump that up to 4.4

glemieux: Also won't `floor` for a negative number result in an erroneous output?  E.g. -63.59 will return ilat_int = -64.  I think we want `int` here.

rgknox: Well, we want to just preserve the number before the decimal.  I'm thinking it will now take the following combination:

ilat_int = floor(abs(currentSite%lat))

This gives it back it's correct sign:
sign( ilat_int, int(currentSite%lat) )

glemieux: Won't `int()` retain the sign as well?  I was thinking something like:
```
ilat_int = int(currentSite%lat)
ilat_dec = abs(int(100000*(currentSite%lat - real(ilat_int,r8))))
```
I thing we need to add `abs()` to the `_dec` too.

rgknox: actually, sorry, your right.  INT() looks like it is a floor towards zero.  I was thinking of NINT(), which rounds up sometimes.

main/FatesInventoryInitMod.F90
```diff
@@ -940,4 +955,92 @@ subroutine set_inventory_edcohort_type1(csite,bc_in,css_file_unit,npatches, &
...
+       
+       write(pss_file_out,*) 'time patch trk age area water fsc stsc stsl ssc psc msn fsn'
+       write(css_file_out,*) 'time patch cohort dbh hite pft nplant bdead alive Avgrg'
+             
+       ipatch=0
+       currentpatch => currentSite%youngest_patch
+       do while(associated(currentpatch))
+           ipatch=ipatch+1
+           
+           write(patch_str,'(A7,i4.4,A)') '<patch_',ipatch,'>'
```
glemieux: Is this a typo?  Should `patch_str` be `pss_file_out`?

rgknox:
@ckoven , there is a basal area check during the first fusion of the inventory data.  Now that we default to crown area conservation during cohort fusion, these checks are not passing.  I see two roads forward:
1) the first fusion on the inventory initialization should optionally use diameter (BA) conservation instead of crown conservation
2) I change this check during the inventory initialization to make sure that crown area is conserved

FWIW: The fusion on the BCI dataset reduced Basal Area from 32 m2/ha to 29 m2/ha with crown-area conservation enabled.

We can make an issue of this going forward too.

glemieux:
This looks good to me. To clarify, wrt the basal area check, you're proposing nuking that section regardless and fixing it with later PR?

rgknox:
After thinking a little bit, I think the way forward is to just remove that check.  In the past, we conserved basal area whenever we did fusion, and so that check should had passed.  Now, we don't, instead we conserve crown area (by default).  So there should be no expectation of preserving area.  There should be a reasonable amount of divergence from the original basal area, but I don't think we need a check there.

glemieux:
@ckoven are you comfortable with Ryan's changes to remove the basal area check?  If so we'll move forward and merge this feature.

ckoven:
yes thats fine with me. thanks both.

glemieux:
@rgknox do you think we should run a test given the removal of the basal code?  I did get the code to build successfully.

rgknox:
The code that has been touched is not covered by our standard regression testing suite.  I did however make sure that the model was building and generating the expected file names, for the four quadrants of the earth, N,S,E, W.  The one test I don't think I have done yet, which would probably be a good idea: is to generate the initialization file for BCI, and then compare a run that uses the output, to a base run.  I could probably get to that this week.

glemieux:
> The one test I don't think I have done yet, which would probably be a good idea: is to generate the initialization file for BCI, and then compare a run that uses the output, to a base run.

Would you mind me looking over your shoulder when you get to that?  Or walking me through it?

rgknox:
sure! will coordinate offline

rgknox:
Here is a test comparing:
1) single site simulation at BCI that is initialized with inventory, then saves a copy of the post-fusion inventory, and then runs for 3 years
2) single site simulation at BCI that initializes with the post-fusion inventory files from run 1, and then runs for 3 years

Results are indistinguishable.

https://drive.google.com/file/d/1T4E8GTtzIGDWipeKwuhUo76f1sbNUnyS/view?usp=sharing

glemieux:
Looks good @rgknox.  Thanks.

PR #537, opened 2019-05-13
Update README.md

rgknox:
This PR changes the front matter of the FATES repo, the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ For more information on the FATES model, see our [wiki](https://github.com/NGEET
 
 [Code of Conduct](https://github.com/NGEET/fates/blob/master/CODE_OF_CONDUCT.md)
 
+[Choosing a Version of Fates](https://github.com/NGEET/fates/wiki/Choosing-a-Version-of-FATES)
+
 
 ## Important Note:
 ------------------------------


rosiealice:
Otherwise this looks great...

rgknox:
@rosiealice , I added a note at the bottom of that page.  Is that what you were looking to see?  Feel free to edit and update

jkshuman:
@rgknox How do classify something as a release vs pre-release? There isn't an obvious connection to the api numbers or the sci numbers.
I think we can make this a bit more clear, but, without experience with svn, if I read this, I would choose master, or look for the last release tag. I like the "latest release" and "pre-release" flags. So, further pre-releases would be red flags, and the latest release will retain that green flag until you designate another release?

rgknox:
@jkshuman , good question.  
To start, this idea of using the pre-release versus release classifications of tags to convey stability... maybe its more confusion that its worth.
Perhaps its worth it to maintain a branch that is oriented towards stability.  We could start a discussion on how to manage that branch (and potentially its own separate tags).  I am a little wary of this, as it will be a non-trivial time-sink for the SE's to maintain (but maybe it won't be too bad ...).

jkshuman:
@rgknox I like the release flag as it orients the user to a specific branch. We are working hard to get things into master, and testing all possibilities is just tough. I think the challenge is how to define a release. It's stable until it isn't stable...Maybe just have a release flag and get rid of the pre-release flags? If we encounter an issue with a release tag, then you can revert a previous branch to be the designated "release"?
I think that maintaining a separate branch for stability sounds like more trouble than good. It is hard enough tracking what is happening with the existing master repo and side branches with testing. No need to add to the SE load.

rgknox:
Thanks for the feedback @jkshuman.
The "pre-release" versus "release" distinction is binary, so I can't specify as neither.  But maybe the way to distinguish a tag that is blessed for use by copious testing and vetting, is simply to only make releases of those tags?

glemieux:
> Perhaps its worth it to maintain a branch that is oriented towards stability. We could start a discussion on how to manage that branch (and potentially its own separate tags). I am a little wary of this, as it will be a non-trivial time-sink for the SE's to maintain (but maybe it won't be too bad ...).

Famous last words @rgknox ;).  That said, I think it might be worthwhile to discuss adopting a multi-branch scheme versus using the release/pre-release tag on the master branch, as it avoids the confusion as @rosiealice pointed out.  Personally, I'm most familiar with having the master branch be the stable branch in which some sort of gating/review mechanism (e.g. a formal code review meeting) is necessary to allow new features on the master.  But that said, while that experience was in a similar environment to this one (i.e. next-generation software being constructed by multiple developers with different focus areas) we did have a launch date for the product (literally :rocket:), which meant we needed to set milestones/release targets to measure progress towards that launch.  So in that sense, the release distinction was based on top-level requirements and the schedule as opposed to level of testing coverage.

I guess that brings me back to @jkshuman 's question: what do we define as release/pre-release?  Whatever method we use seems to hinge on that.  Do we have anything similar to a 'product launch' that we could use to inform the milestones and then work down towards defining a release/pre-release?  @rgknox you mentioned that this was a suggestion of the folks on the CSEG team, yeah?  Are there CESM roadmaps that could inform what must be in a release?  Or how they define stability?

rgknox:
I feel like we are more analogous to a space station than a rocket.  We are already in orbit, and thus our first launch is complete, but we keep adding and fixing modules as we go.

glemieux:
Oooh, I like that @rgknox!  That analogy also suggests having some of the release/pre-release definition set by the host-models that we interface with.  Gotta make sure that the new module to be added meets a certain set of requirements to support (or not muck up at least) the existing whole.

Issue #502, opened 2019-03-11
Suggest making the default for modules private, so you have to designate what's public


ekluzek:
This is a suggestion, that I want to have some discussion on. I have a PR that will come in that does this for one module. Having public data that is accessible everywhere is evil -- although sometimes a necessary evil. But, something that you want to limit. The biggest problem is if it can be changed everywhere, and a good way to manage that is to use "protected" so that if can only be changed in the parent module. That is something I'm seeing used which is great. I also suggest that as part of the standard boilerplate after declaring "implicit none" you declare, "private" so that all data/subroutines/functions are assumed private rather than public. This makes it a pain to declare something public -- but that's the way it should be -- it should be hard to do something that should be done sparingly. An artifact of NOT doing this is that everything "used" by the module is then public as well, so it can be hard to figure out where data actually lives with a lot of modules that are default public. This last point is the main reason I suggest doing this.

There's a little bit of discussion on this in the CLM Dev guide here...

https://wiki.ucar.edu/display/ccsm/Comprehensive+list+of+standards#Comprehensivelistofstandards-Modularity

rgknox:
@ekluzek , I think setting defaults to private, and using "protected" when possible, is a great suggestion, and I support using this approach in FATES.

glemieux:
@rgknox per our discussion with Charlie today, would you or @ekluzek assign the issue to me please?

glemieux:
I did an audit of the modules this morning and it looks like about 50% of the modules are set `private` by default currently. Many of the others are public by default, but explicitly set to private on a variable-by-variable level.

ekluzek:
The reason to flip it to private by default, is so that you have to explicitly make things public, which is the thing you want to be wary about, rather than private. The worst offender is that anything that's in the "use" list from other modules -- is going to be public -- and you can't make it private to just within the current module. So I'm glad you'll start working on these, thanks for looking into it.

glemieux:
@rgknox I see some of the subroutines in `EDTypesMod.F90` below `contain` being used by other modules by not being called.  Those ones are `dump_cohort` and 'val_check_ed_vars'.  Interestingly it looks like `val_check_ed_vars` isn't being called in `EDTypesMod.F90' itself.  Looks like it was added about 2 years ago: https://github.com/NGEET/fates/commit/df47cfa2fb449614ce559a16f268548368f703ab#diff-14f02c3d54c6afd6502e04e81b2917a6

Has this functionality been superseded?

[...]

PR #539, opened 2019-05-28
Conversion of modules from public to private by default

glemieux:
### Description:
This pull request converts all remaining modules that were public by default to `private` by default.  About 50% of the existing code base was already defined as `private` by default.  The majority of the changes were to modules that are part of `main` with only a few other functions in `parteh` and `fire`.

This work was conducted in response to issue #502.    The intent going forward is to have developers explicitly make `public` the procedures, variables, and data necessary to avoid potential 'leaks' in functionality to users.  That said, in this PR all code should retain the same level of privacy as the baseline code it was tested against; i.e. no effort was given to determining what existing code might benefit from a higher level of privacy that currently exists.

*Notes on formatting*:
- All modules now contain a `private` declaration immediately following `implicit none`.
- Variables and types are declared `public` inline.  For example: `integer, parameter, public :: fates_r8 = selected_real_kind(12)` and `type, public :: fates_history_interface_type`
- Procedures (i.e. functions and subroutines) are defined as public typically at the end of the variable and type definitions at the top of the module file like so: `public :: RKF45`.  Note that many procedures are define as public being `contained` within type, which itself may be public.

### Collaborators:
@rgknox @ekluzek 

None

### Checklist:

### Test Results:

CTSM (or) E3SM (specify which) test hash-tag: clm5.0.dev008-105-gbc17668

CTSM (or) E3SM (specify which) baseline hash-tag: clm5.0.dev008-105-gbc17668

FATES baseline hash-tag: sci.1.26.3_api.7.2.0-16-gf1d4bc5

Test Output:
*Test output*: /glade/u/home/glemieux/scratch/clmed-tests/private-public-modules.fates.cheyenne.intel.Cbc17668-Fabb9df6

*Baseline output*: /glade/u/home/glemieux/scratch/clmed-baselines/private-public-modules.fates.cheyenne.intel.Cbc17668-Ff1d4bc5

All tests passing excepting:
```
FAIL ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro COMPARE_base_rest
FAIL ERS_Ld60.f45_f45_mg37.I2000Clm50FatesCruGs.cheyenne_intel.clm-Fates MEMCOMP Error: Memory usage increase > 10% from baseline
```


fire/SFParamsMod.F90
```diff
@@ -7,68 +7,58 @@ module SFParamsMod
...
-   character(len=param_string_length),parameter :: SF_name_SAV = "fates_SAV"
-   character(len=param_string_length),parameter :: SF_name_FBD = "fates_FBD"
-   character(len=param_string_length),parameter :: SF_name_min_moisture = "fates_min_moisture"
-   character(len=param_string_length),parameter :: SF_name_mid_moisture = "fates_mid_moisture"
-   character(len=param_string_length),parameter :: SF_name_low_moisture_Coeff = "fates_low_moisture_Coeff"
-   character(len=param_string_length),parameter :: SF_name_low_moisture_Slope = "fates_low_moisture_Slope"
-   character(len=param_string_length),parameter :: SF_name_mid_moisture_Coeff = "fates_mid_moisture_Coeff"
-   character(len=param_string_length),parameter :: SF_name_mid_moisture_Slope = "fates_mid_moisture_Slope"
-
+   real(r8),public,protected :: SF_val_fdi_a
```
rgknox: @glemieux , if a variable is defined as protected, does it make it implicitly public?  I poked around on the internet and could not find an answer.

main/EDTypesMod.F90
```diff
@@ -14,27 +14,35 @@ module EDTypesMod
   use FatesConstantsMod,     only : n_anthro_disturbance_categories
 
   implicit none
+  private               ! By default everything is private
   save
 
-  integer, parameter :: maxPatchesPerSite  = 14   ! maximum number of patches to live on a site
-  integer, parameter :: maxPatchesPerSite_by_disttype(n_anthro_disturbance_categories)  = &
+  ! Make public necessary subroutines and functions
+  public :: val_check_ed_vars
```
rgknox: In the discussion at the top of the PR, you mentioned that you were moving procedure specifications to the end after the module variables were defined.  I have no strong feelings about order, but wanted to note it here.

main/FatesParameterDerivedMod.F90
```diff
@@ -14,7 +14,10 @@ module FatesParameterDerivedMod
   use FatesConstantsMod,     only : g_per_kg
   use FatesInterfaceMod,     only : nleafage
   
-  type param_derived_type
+  implicit none
```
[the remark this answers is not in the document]
rosiealice: This just made me laugh inappropriately loudly in a very large room full of people.

main/FatesParametersInterface.F90
```diff
@@ -53,19 +54,21 @@ module FatesParametersInterface
      type(parameter_type), private :: parameters(max_params)
 
    contains
```
rgknox: @glemieux , in the following block of changes, I was surprised that previously listed procedures that were explicitly public were switched to undefined (no label).  Were these not used globally?  Also, private functions are listed down below, which are implicitly private now, right?

rgknox:
@glemieux , I looked through the changes and they look good. Thanks!
I left a few comments in there, mostly just questions.  Pending discussion and conflict resolution and re-test I will integrate.

fire/SFParamsMod.F90
glemieux: That's a good question.  Chivers' book isn't explicit but the example code they give to show suggests that it's not the case that it's implicitly public.  I'll write up a quick script and test it out in a bit.

main/EDTypesMod.F90
glemieux: Good catch.  I'll fix that.

main/FatesParametersInterface.F90
rgknox: ok, just saw that those were bound inside a type, which is declared as public, (@glemieux showed me)

fire/SFParamsMod.F90
glemieux: @rgknox based on the [small sample code](https://github.com/glemieux/fates-scratch/blob/master/fortran-scratch/ProtectedTestMain.F90) I worked up it looks like setting something as `protected` does not make it implicitly `public`.

glemieux:
Deconflicted the branch with master and re-ran tests.  All expected pass, b4b:

/glade/u/home/glemieux/scratch/clmed-tests/private-public-modules-mergedeconflict.fates.cheyenne.intel.C08fc948-F8efc24f

PR #542, opened 2019-06-03
Remove unnecessary (and erroneous) calls to treesai and treelai 

rgknox:
### Description:

There are several calls to calculate tree LAI and SAI during the growth portion of the dynamics driver that are unnecessary.  Moreover, without re-calculating crown area, the third call is erroneous and may generated errors.

### Collaborators:

@jkshuman 

There should be no changes in answers.

### Checklist:

### Test Results:

TESTS:

See discussion

### DIFF
--- a/main/EDMainMod.F90
+++ b/main/EDMainMod.F90
@@ -273,7 +273,6 @@ contains
     real(r8) :: dbh_old               ! dbh of plant before daily PRT [cm]
     real(r8) :: hite_old              ! height of plant before daily PRT [m]
     logical  :: is_drought            ! logical for if the plant (site) is in a drought state
-    real(r8) :: leaf_c
     real(r8) :: delta_dbh             ! correction for dbh
     real(r8) :: delta_hite            ! correction for hite
 
--- a/main/EDMainMod.F90
+++ b/main/EDMainMod.F90
@@ -354,15 +354,6 @@ contains
           
           currentSite%flux_in = currentSite%flux_in + currentCohort%npp_acc * currentCohort%n
 
-          leaf_c = currentCohort%prt%GetState(leaf_organ, all_carbon_elements)
-          currentCohort%treelai = tree_lai(leaf_c, currentCohort%pft, currentCohort%c_area, currentCohort%n, &
-               currentCohort%canopy_layer, currentPatch%canopy_layer_tlai, &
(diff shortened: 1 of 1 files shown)


rgknox:
35 year test at BCI: default parameterization, two tropical trees:
https://drive.google.com/file/d/1_BIlH56LmAY55OBFtdOn5zF9hjj_7Mte/view?usp=sharing
Results are indistinguishable from base

rgknox:
All exptected pass, b4b:
/gpfs/fs1/scratch/rgknox/clmed-tests/treelai-redund-C08fc948-F8ff6f9c.fates.cheyenne.intel

rgknox:
From discussion with @ckoven: "yeah I don’t see how cohort%treelai or cohort%treesai need to be used there.  so best to delete…"

glemieux:
I'm seeing that the it looks like this was added during commit https://github.com/NGEET/fates/commit/3204b39ee2e98c6866674b234395809198ffdf0e and https://github.com/NGEET/fates/commit/5ad22a1b636b741a2d87084816f9eb1cf232223d (thank you GitLens/VSCode!) as part of PR #501.  @rgknox, am I understaning correctly that thse are unnecessary due to `treelai` and `treesai` being handled via other submodules (e.g. EDCohortDynamicsMod)?

rgknox:
@glemieux , yes, treelai and treesai are updated during the canopy structure routines, see here:

https://github.com/NGEET/fates/blob/master/biogeochem/EDCanopyStructureMod.F90#L1282
and then again in a subroutine called within the one above:
https://github.com/NGEET/fates/blob/master/biogeochem/EDCanopyStructureMod.F90#L1546

Those routines handle things related to crowns, canopies and leaf area; which makes it a good location to be updating treelai and treesai. These values are not needed before these routines are called.

rgknox:
50 year 10x10 smoke test completed as well:
/gpfs/fs1/scratch/rgknox/fates-clm-tests/remove-redund-f10_f10_mg37/run

rgknox:
removed declaration of leaf_c, per @ckoven, via 87db059

jkshuman:
@rgknox My complementary 2PFT South America runs with and without fire are into years 14 and 16 and still running. Seems very slow, but think I am competing with traffic on Hobart.

PR #543, opened 2019-06-03
functional (scientific) unit tests for allometry functions

rgknox:
### Description:

This set of changes adds in a functional unit testing system for the FATES allometry functions.  This
includes wrapper code (f90) and python code.  The testing system can be pointed to any valid "cdl" type file to generate the allometric trajectories of the PFTs contained in the file.

This uses python code to drive the native FATES fortran allometry routines using the "ctypes" library.  

There is some code that will auto-generate the correct wrapper fotran codes by assessing the list of relevant variables currently in FatesAllometryMod.F90 and EDPftvarconMod.F90.

A script is provided that builds the fortran part of the code as shared libraries, callable by python. However, this scripting is very simple (bash code), and has only been tested with gfortran.

This is accessory code and will not affect FATES simulations.

### Checklist:


main/EDParamsMod.F90
```diff
@@ -2,14 +2,14 @@ module EDParamsMod
...
-  
+
+   use FatesConstantsMod, only : r8 => fates_r8
    use FatesParametersInterface, only : param_string_length
    use FatesGlobals        , only : fates_log
    use FatesGlobals        , only : endrun => fates_endrun
 
    ! CIME Globals
    use shr_log_mod         , only : errMsg => shr_log_errMsg
-   use shr_kind_mod        , only: r8 => shr_kind_r8
```
glemieux: What predicated the change to `fates_r8`?  Or put another way, what is the history behind the use of `shr_kind_mod`?

main/EDTypesMod.F90
```diff
@@ -4,7 +4,6 @@ module EDTypesMod
   use FatesConstantsMod,     only : ifalse
   use FatesConstantsMod,     only : itrue
   use FatesGlobals,          only : fates_log
-  use shr_infnan_mod,        only : nan => shr_infnan_nan, assignment(=)
```
glemieux: Looks like this wasn't actually being used?

rgknox: yup, not used

main/EDParamsMod.F90
rgknox: The unit test pre-processor code will actually define r8's slightly differently than how we normally do for a standard elm/clm-fates run, in that it will define it based on the iso c standard library.  Its important to use that library for the unit tests.  So in order to use the right definition, we must point to fates constants (where it is defined).  I don't think there is much (at all?) difference between our definition of an 8-byte real, and the definition in shr_kind_mod.  But its more consistent at the very least.

rgknox:
/glade/scratch/rgknox/clmed-tests/fates-allom-ufuncs-C08fc9481-F2689a10a.fates.cheyenne.intel

pass

PR #546, opened 2019-06-12
fix the direct-logging caused changes in trunk product, CWD, and litt…

huangmy:
This PR aims to fix the logic in generating truck product, CWD, and litter stock associated with direct logging and transpiration loss. The direct-logging induced perturbation should not affect the non direct-logging calculations

### Description:

Modified biogeochem/EDLoggingMortalityMod.F90 and biogeochem/EDPhysiologyMod.F90

### Collaborators:

@rgknox 

Yes, but only when the logging module is active.

### Checklist:

- [X] I have updated the in-code documentation .AND. (the [technical note](https://github.com/NGEET/fates-docs) .OR. the wiki) accordingly.

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -51,6 +51,7 @@ module EDPhysiologyMod
   use EDParamsMod           , only : fates_mortality_disturbance_fraction
   use EDParamsMod           , only : q10_mr
   use EDParamsMod           , only : q10_froz
+  use EDParamsMod           , only : logging_export_frac
 
   use FatesPlantHydraulicsMod  , only : AccumulateMortalityWaterStorage
   
@@ -1331,14 +1332,34 @@ contains
          currentPatch%cwd_BG_in(c) = currentPatch%cwd_BG_in(c) + (struct_c + sapw_c) * & 
               SF_val_CWD_frac(c) * dead_n * (1.0_r8-EDPftvarcon_inst%allom_agb_frac(currentCohort%pft))
          
-         ! Send AGB component of boles from non direct-logging activities to AGB litter pool
+         ! Send AGB component of boles from both direct and non-direct logging activities to AGB litter pool
          if (c==ncwd) then
             
             ! CWD contributed by indirect damage
             currentPatch%cwd_AG_in(c) = currentPatch%cwd_AG_in(c) + (struct_c + sapw_c) * & 
                  SF_val_CWD_frac(c) * (dead_n_natural+ dead_n_ilogging) * &
                  EDPftvarcon_inst%allom_agb_frac(currentCohort%pft)
+
+            ! CWD contributed by logged boles due to losses in transportation
+            currentPatch%cwd_AG_in(c) = currentPatch%cwd_AG_in(c) + &
+                 (1.0_r8 - logging_export_frac) * (struct_c + sapw_c) * &
+                 SF_val_CWD_frac(c) * dead_n_dlogging * &
+                 EDPftvarcon_inst%allom_agb_frac(currentCohort%pft)
             
+            ! Send AGB component of boles from direct-logging activities to
+            ! export/harvest pool
+            ! Generate trunk product (kgC/day/site)
+            trunk_product =  logging_export_frac * (struct_c + sapw_c) * &
+                 SF_val_CWD_frac(c) * dead_n_dlogging * &
+                 EDPftvarcon_inst%allom_agb_frac(currentCohort%pft) * &
+                 hlm_freq_day * currentPatch%area
+            
+            currentSite%flux_out = currentSite%flux_out + trunk_product
+            
+            ! Update diagnostics that track resource management
+            currentSite%resources_management%trunk_product_site  = &
+                 currentSite%resources_management%trunk_product_site + &
+                 trunk_product            
          else
             
             currentPatch%cwd_AG_in(c) = currentPatch%cwd_AG_in(c) + (struct_c + sapw_c) * & 
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -51,7 +51,6 @@ module EDPhysiologyMod
   use EDParamsMod           , only : fates_mortality_disturbance_fraction
   use EDParamsMod           , only : q10_mr
   use EDParamsMod           , only : q10_froz
-  use EDParamsMod           , only : logging_export_frac
 
   use FatesPlantHydraulicsMod  , only : AccumulateMortalityWaterStorage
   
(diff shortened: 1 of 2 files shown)


huangmy:
@rgknox Could you please help review the changes when you have a chance? If you agree with my modifications, I will perform some tests.

rgknox:
@huangmy , my understanding was that the trunk products calculated in EDPhysiology were still valid.  I agree with the new code added in EDloggingMortalityMod though.  It is possible that a tree meets the harvesting requirements (diameter) and exists in the lower canopy (thus not creating a disturbance when its felled).  Is this still true?

huangmy:
@rgknox I agree with your understanding. Even in that case, only direct logging mortality can create trunk products, therefore should be moved to EDloggingMortalityMod, correct? The calculation in EDPhysiology are all associated with collateral and infrastructure damages so I don't think the loggers would bother to bring them off-site. What do you think?

rgknox:
I will double check, but I thought direct logging is also handled in EDPhysiologyMod.

rgknox:
After double checking, I can confirm that direct logging is associated with both parts of the code. 

 The part in EDPhysiologyMod handles non-disturbance generating mortality and its litter fluxes.
 The part in EDLoggingMortality handles litter fluxes associated with disturbance events. 
  https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L1279

huangmy:
@rgknox agreed. I guess the comment above https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L1335 was confusing.
I have added the lines back to EDPhysiologyMod.F90 and revised the comment. Please take a look again.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1332,33 +1332,34 @@ subroutine CWD_Input( currentSite, currentPatch)
          currentPatch%cwd_BG_in(c) = currentPatch%cwd_BG_in(c) + (struct_c + sapw_c) * & 
               SF_val_CWD_frac(c) * dead_n * (1.0_r8-EDPftvarcon_inst%allom_agb_frac(currentCohort%pft))
          
-         ! Send AGB component of boles from non direct-logging activities to AGB litter pool
+         ! Send AGB component of boles from both direct and non-direct logging activities to AGB litter pool
```
rgknox: How adding some text that indicates non-exported portion of direct harvesting:

```
! Send AGB component of boles from logging activities into the litter.  
! This includes fluxes from indirect modes of death, as well as the 
! non-exported boles due to direct harvesting.
```

Note, my terminology may be wrong.  I'm assuming that a bole is the above ground portion of a trunk, and that term is appropriate for natural vegetation as well as wood products.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1332,33 +1332,34 @@ subroutine CWD_Input( currentSite, currentPatch)
...
          if (c==ncwd) then
             
             ! CWD contributed by indirect damage
             currentPatch%cwd_AG_in(c) = currentPatch%cwd_AG_in(c) + (struct_c + sapw_c) * & 
                  SF_val_CWD_frac(c) * (dead_n_natural+ dead_n_ilogging) * &
                  EDPftvarcon_inst%allom_agb_frac(currentCohort%pft)
-            
+
             ! CWD contributed by logged boles due to losses in transportation
             currentPatch%cwd_AG_in(c) = currentPatch%cwd_AG_in(c) + &
```
rgknox: yay, I love it when people shorten lines

biogeochem/EDPhysiologyMod.F90
huangmy: Done. please check again.

biogeochem/EDPhysiologyMod.F90
huangmy: :-)

rgknox:
This all looks great @huangmy , thanks for the fixes.

rgknox:
testing on cheyenne

rgknox:
regression tests pass on cheyenne (b4b):
/gpfs/fs1/scratch/rgknox/clmed-tests/fates-sci.1.27.4_api.7.3.0-C08fc948-F15489c7.fates.cheyenne.intel

rgknox:
@huangmy, I just noticed this while doing some comparisons with my multi-element test branch, we have two copies of "logging_export_frac" in EDLoggingMortalityMod.F90: 

One defined (correctly) in EDParams, and one defined localy (should be removed) here:

https://github.com/NGEET/fates/blob/49e258f7c5db7273523187dad62dfd977559b9fe/biogeochem/EDLoggingMortalityMod.F90#L64

(I'm also somewhat at a loss as to why the compiler is allowing two copies of the same variable...I feel this is a great place to complain and stop)

huangmy:
@rgknox I was kind of wondering about that too. I thought you were doing it intentionally :-). Do you want me to remove it?

rgknox:
yes, it is an artifact, should be removed

huangmy:
@rgknox just removed it. Please check.

rgknox:
I checked to make sure that the model compiles and runs with the redundant definition removed. I don't think this needs a complete re-test though.  @glemieux , can you be final judiciary and integrate/request changes?

glemieux:
@huangmy did you want us to hold on the merge until you present results to the non-standard testing case at Tapajos?

huangmy:
@glemieux I will working on it right now. Yes, please hold and I will report back tomorrow. Thanks.

huangmy:
@rgknox @glemieux It turns out that I am not exactly sure what success means for the non-standard test given the recent updates. It will take me a while to get scientifically valid results. So if Ryan's tests were successful, please go ahead to integrate the PR.

huangmy:
@rgknox @glemieux after struggling a while with the input file format, I eventually was able to repeat the numerical experiments at Tapajos. See the changes in carbon pools below, which appears to be reasonable to me. Note that the specific parameter set does not warrant coexistence of two PFTs, therefore the empty panels (g) and (h). Please take a look and let me know if you want to see other results. 

CarbonPoolskm83_1

glemieux:
Thanks for the update @huangmy.  We've been unable to complete the regression testing on our side due to down time on cheyenne and problems on lawrencium.  Ryan's on vacation right now, but I'll make sure to get the testing down as soon as possible and tag up with him or @ckoven on your results.

rgknox:
Expected still pass:
/glade/scratch/rgknox/clmed-tests/fates-sci.1.27.4_api.7.3.0-v2-C08fc9481-Fba5209a0.fates.cheyenne.intel
@glemieux , I see no reason not to pull now

PR #547, opened 2019-06-25
Update the PFT swapper tool to use double precision

rgknox:
### Description:

The python tool used for pulling out specific PFTs from the parameter file needs to be converted to use doubles (instead of single precision real numbers).

### Collaborators:

@JessicaNeedham

### Testing:

This tool is not covered in standard regression testing.  I did however use it to manipulate the default file, and check to see if the resulting file had all the variables, and that they were doubles.

### DIFF
--- a/tools/FatesPFTIndexSwapper.py
+++ b/tools/FatesPFTIndexSwapper.py
@@ -25,13 +25,13 @@ pft_dim_name = 'fates_pft'
 prt_dim_name = 'fates_prt_organs'
 
 
-class timetype:   
-    
+class timetype:
+
     # This is time, like the thing that always goes forward and cant be seen
     # or touched, insert creative riddle here
 
     def __init__(self,ntimes):
-    
+
         self.year  = -9*np.ones((ntimes))
         self.month = -9*np.ones((ntimes))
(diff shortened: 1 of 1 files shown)


JessicaNeedham:
Looks good to me. I'd changed 'f' to 'd' and had no problem with it. I had also changed the word 'float' on line 180 to 'double' but I'm a very new python user so maybe that was unnecessary.

rgknox:
@JessicaNeedham , yes, I think you are right, that double should be there, will fix.

JessicaNeedham:
I added 'from numpy import *' to get that to work.

rgknox:
Actually, double checking this, python by default uses double precision for all real types. float() actually returns a double precision variable, so we should be good to go here.

rgknox:
This was helpful, I needed to read-up on python precision: https://docs.python.org/2/tutorial/floatingpoint.html

JessicaNeedham:
Good to know! Thanks - I'll check out that link.

rgknox:
I'm going to go ahead and merge this since the changes are trivial and ancillary to the base code

PR #553, opened 2019-07-16
patch fusion fix to prevent patch heterogeneity collapse

### LINKED ISSUE #323: episodic patch heterogeneity collapses during long runs
All,

I've noticed a weird thing that happens, which I want to post here in hopes that anyone has ideas about possible solutions.  In long runs (> 500 years or so), where I'm trying to evaluate what sort of long-term steady state the model is tending to relax to, instead of it actually reaching a long-term steady state, it is instead doing this thing where every once in a while the patch heterogeneity collapses, dropping from 10 (the max number at present) to only 1-2 patches all at once.  The heterogeneity then slowly builds up again and then eventually collapses again (with a 200-300 years or so timescale).  I've seen this behavior in a lot of different runs now, some with identical PFTs, some with competing PFTs.  To investigate the overall patch dynamics, I did a couple runs with annually-resolved patch age bin resolution.  Both were run with repeating BCI met driver data (so any variation at decadal or slower frequencies is internally-generated).  Attached a couple figures of what his looks like.  On the y axis are the patch age, and the x axis is the time since the start of the runs.  The lines show trajectories of a given patch. The color of the lines is the patch area. When there are jumps it is because patches are getting fused:

screen shot 2018-01-22 at 2 55 34 pm

screen shot 2018-01-22 at 2 56 02 pm

If I plot timeseries of the number of patches in each run, it looks like this. Red line is one of the runs, and black line is the other.

screen shot 2018-01-22 at 2 58 25 pm

The thing that I find slightly troubling is that if you look at biomass or GPP in the runs (below), you can see that when the patch collapses occur, it also shows up as brief excursions in those fields.  So this is something that has macro-scale effects on the model behavior.

screen shot 2018-01-22 at 2 59 52 pm

screen shot 2018-01-22 at 3 01 35 pm

Anyway, I mainly wanted to note this so that others are aware of it.  It seems like we need to make some changes to the patch fusion criteria to prevent this from happening.  One experiment that I did do was to reduce the increment by which the patch fusion criteria (profiletol) loosens when the max number of patches is exceeded from 10% to 2% to see if this was just an artifact of loosening that too much, but that didn't seem to have an effect

ckoven:
This code slightly changes the patch fusion code to prevent an edge case that has been happening fairly frequently, whereby the patches would all get fused, thereby losing all the heterogeneity.

### Description:
basically this just resets the patch fusion tolerance after every fusion event, so that if the code had encountered a situation where it needed to fuse two patches that were dissimilar enough that the worst biomass bin comparison between the two patches was that one patch had some biomass in a given bin and the other didn't, the looping structure would then proceed to fuse all of the patches. now that shouldn't happen.

note this also changes the outcome of inventory initialization.  earlier code was tending to collapse all the quadrats at BCI into a single patch.  This code leaves several patches after the initial round of patch fusion during inventory initialization using the BCI census data.

fixes #323.

### Collaborators:
some email discussion with @rgknox and @rosiealice.

this should change answers, but possibly only for long runs.

### Checklist:

- [SORT OF] If answers were expected to change, evaluation was performed and provided .  I write "sort of" because I actually did the evaluation on e27deac, not db47637.  the current PR (db47637) is the same logic but applied to master (and also with some added documentation and cleaned up documentation earlier in the code block).

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -1701,13 +1701,13 @@ contains
     do i_disttype = 1, n_anthro_disturbance_categories
 
        !---------------------------------------------------------------------!
-       !  We only really care about fusing patches if nopatches > 1           !
+       !  We only really care about fusing patches if nopatches > 1          !
        !---------------------------------------------------------------------!
 
        iterate = 1
 
        !---------------------------------------------------------------------!
-       !  Keep doing this until nopatches >= maxPatchesPerSite                         !
+       !  Keep doing this until nopatches <= maxPatchesPerSite               !
        !---------------------------------------------------------------------!
 
        do while(iterate == 1)
@@ -1720,9 +1720,9 @@ contains
              currentPatch => currentPatch%older
           enddo
 
-          !---------------------------------------------------------------------!
+          !-------------------------------------------------------------------------------!
           ! Loop round current & target (currentPatch,tpp) patches to assess combinations !
-          !---------------------------------------------------------------------!   
+          !-------------------------------------------------------------------------------!   
           currentPatch => currentSite%youngest_patch
           do while(associated(currentPatch))      
              tpp => currentSite%youngest_patch
@@ -1733,8 +1733,10 @@ contains
                 endif
 
                 if(associated(tpp).and.associated(currentPatch))then
-
-                   ! only fuse patches whose anthropogenic disturbance categroy matches taht of the outer loop that we are in
+                   !--------------------------------------------------------------------!
+                   ! only fuse patches whose anthropogenic disturbance category matches !
+                   ! that of the outer loop that we are in                              !
+                   !--------------------------------------------------------------------!
                    if ( tpp%anthro_disturbance_label .eq. i_disttype .and. &
                         currentPatch%anthro_disturbance_label .eq. i_disttype) then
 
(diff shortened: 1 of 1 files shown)


rosiealice:
Looks good to me...

glemieux:
@ckoven, per our conversation, I ran the regression tests.  All expected pass with the exception of some DIFFs:
```
FAIL ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro RUN time=24
    FAIL ERS_Ld60.f45_f45_mg37.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesNoFire MEMCOMP Error: Memory usage increase > 10% from baseline
    FAIL SMS_Lm13.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesColdDef BASELINE ckoven-patch_fusion_fix.fates.cheyenne.intel.C08fc9481-F2bd9cb46: DIFF
    FAIL SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro RUN time=21
    FAIL SMS_Lm6.f45_f45_mg37.I2000Clm50FatesCruGs.cheyenne_intel.clm-Fates BASELINE ckoven-patch_fusion_fix.fates.cheyenne.intel.C08fc9481-F2bd9cb46: DIFF
```
Comparison files can be found here: ```/glade/u/home/glemieux/scratch/clmed-tests/ckoven-patch_fusion_fix.fates.cheyenne.intel.C08fc9481-Fdb476373```

rgknox:
I think those results make sense @glemieux .  The changes to the code should generate answer changes.  Likewise I'm not surprised that it is only generating differences in the longer runs, since they use the fusion algorithms more, and are more likely to have larger differences in members that are being fused.

ckoven:
yeah I agree.  I'm slightly surprised to see changes in a 3 month single poiint run, but I guess that just means the problem may have occurred more often than I thought.

PR #554, opened 2019-07-19
Grass phenology update

xuchongang:
This pull request update the phenology  for  grass. Namely, to drop stems during cold or drought conditions.

### Description:
We made the following key modifications: 
  1) add phen_stem_drop_fraction into the parameter file and EDPhysiologyMod so that the grass can drop their stem during winter or drought;
  2) make the leaf phenology for grass size-dependent.  This modification is made so that it can accommodate the situation of small salt marsh can overwinter due to the protection of tall dead stems. 

### Collaborators:
Lu Zhai, zhailulu@gmail.com

For grass, it will change the phenology if phen_stem_drop_fraction set to values > 0. Otherwise, it should generate the same output as before.

### Checklist:

- [] My change requires a change to the documentation.

### Test Results:
Not yet

FATES baseline hash-tag:

Test Output:


rgknox:
@rosiealice  and I have been having a discussion over the past year or so, about the idea of getting rid of laimemory.   When we re-flush leaves, instead of flushing to the last known leaf biomass, we could calculate the flushing target allometrically based on other terms, like dbh and canopy trim. This result should be different conceptually from laimemory, in that this target value is based on the stature or maximum carrying capacity of the tree, and not the last known quantity the season before.

I see in your pull request that you intend to add sapwood and structure memory, but these targets during flushing could be retrieved from allometry in the same way that leaf flushing would.

I'm curious what everyone here thinks of this idea of dropping the "memory" variables, and using allometry instead.

*Note: When calculating the exact amount to flush on the first bud-burst, we do have provisions in the code to flush a desired fraction of the target.

LuZhai:
@rgknox  **"When we re-flush leaves, instead of flushing to the last known leaf biomass, we could calculate the flushing target allometrically based on other terms, like dbh and canopy trim. "**
Using other terms as a reference should work to tree. But grass species may not have these terms in the beginning of the growing period. So what about only applying the memory to grass species?

rgknox:
> But grass species may not have these terms in the beginning of the growing period.

ok, I will have to dig in more to understand how sapwood and structural tissues in grasses work in your new framework.

xuchongang:
@rgknox,@rosiealice, I like the idea of getting rid of the leaf or sapwood memory.  One potential advantage of leaf or stem memory is that the modeled plant can deviate from the prescribed allometry under stress. If we use the targeted leaf or stem, then the plant can lost the memory of their bad conditions before. So some kind of stress indicator is needed to carry that information.  We can use  canopy trim for leaf. For grass stem turn over, we may not need to consider the stress factor at this stage as the structure will partially or totally lost. For the future, if we want to add the partial stem damage due to hurricane to the stems, then we should carry a crown damage factor. Thus, it is totally fine with me to remove the memory term and we can add some stress factor for leaf or stem to accommodate the stress factors for both trees or grasses.

rgknox:
Question about how this conceptually is supposed to work.

Lets say a grass triggers "stem drop", and then drops a fraction of its stem.  

It is clear from calls to PRTDeciduousTurnover(), that you specify a drop fraction, and that results in a loss of biomass.

But you are saving the amount lost as the memory term, and not the total amount that the plant had before is dropped biomass.   

So when it flushes again, the amount it tries to send from storage to stem, is just replacing the specific loss it suffered during the drop, and not based on a specific target.  Is this right?


*Note: I'm noticing a potential logic flaw now with leaves that may be a legacy issue.  If we do not drop all our leaves, but request to total amount possible to be sent from storage during re-flush, we could over-shoot.

rgknox:
Looks like we responded at the same time.

> If we use the targeted leaf or stem, then the plant can lost the memory of their bad conditions before. So some kind of stress indicator is needed to carry that information. We can use canopy trim for leaf. 

Well, the actual amount of biomass in all pools is always tracked as a state variable.  So it is always remembering its biomass.  

For instance, the actual amount of carbon in a plant's sapwood, at any time is:

sapw_c   = currentCohort%prt%GetState(sapw_organ, carbon12_element)

Also a note on growth.   Regardless of how much stem is flushed during the "leaf-on" event, the stem biomass will be allocated during the normal growth process in an attempt to get the pool back to the target quantity dictated by the stature (dbh) of the plant.  This will always happen before the plant attempts to increase its stature (dbh).

Not that I'm saying a plant wouldn't want to send carbon from storage to stem during a leaf-on flushing event, something needs to hold up  the new leaf tissue.   Just speaking extemporaneously about the algorithms.

xuchongang:
@rgknox , I agreed with you comments on potential overshooting when using leaf memory for partial leaf drops.  For the stem biomass,  I think you make a good point on the flush event. Even if we did not put the exact amount of carbon from previous year, it will grow to the target.

glemieux:
Testing underway.

glemieux:
My local branch of this PR is failing on the exact restart test for the mpi-serial library.  It looks like `sapwmemory` is being set to `NaN` instead of zero or the previous value.  Currently investigating.

glemieux:
@LuZhai @xuchongang I noticed in EDInitMod that the `init_cohorts` subroutine does not include code for `sapwmemory` or `structmemory` for deciduous stress:
https://github.com/NGEET/fates/blob/bfe83fdea80ed031473248520c6b73eb433fe25a/main/EDInitMod.F90#L506-L522

whereas in FatesInventoryInitMod, there is matching code for both stress and seasonality in `set_inventory_edcohort_type1`:
https://github.com/NGEET/fates/blob/bfe83fdea80ed031473248520c6b73eb433fe25a/main/FatesInventoryInitMod.F90#L1045-L1065

Should EDInitMod have the initialization code for `sapwmemory` and `structmemory` for deciduous stress?

xuchongang:
@glemieux , good catch and it we need to have  EDInitMod have the initialization code for sapwmemory and structmemory for deciduous stress.

glemieux:
> @glemieux , good catch and it we need to have EDInitMod have the initialization code for sapwmemory and structmemory for deciduous stress.

Thanks; I'll add it on my local branch.  I fixed the restart error as well just now.  New PR coming to your branch sometime later today.

glemieux:
That was interesting.  I somehow added labels using keyboard shortcuts.  That's what I get for using two screens and not paying attention to both <facepalm>

xuchongang:
@glemieux, thanks and I did not see the pull request on my branch yet.

glemieux:
@xuchongang apologies for the delay.  I spoke too soon; I wanted to wait for the test to complete, which was held up until the weekend in the Cheyenne queue.  The test passed and I've created the PR to your branch.

xuchongang:
@glemieux, thank you and I have made the merge of your pull request.

glemieux:
Testing update:

All standard testing passed B4B with the default parameter file (i.e. `phen_stem_drop_frac = 0`) except for the `SMS_Lm6.f45_f45` test which failed on a DIFF to the baseline.  @rgknox and I took a look at this yesterday and we're still struggling to determine the reason for the difference.  

I also ran a parallel 5 year test at Brazil with and without the stem drop fraction greater than zero (in this case I set all PFTs to a randomly selected value of 0.25).  Output comparison plots run through the ACRE suite can be found here: https://drive.google.com/file/d/1Bj5gBPf7a13bvrG7X01ho2V6jCaiimBL/view?usp=sharing

glemieux:
At @ckoven I plotted out the `SMS_Lm6.f45_f45` differences between the baseline and the PR branch.  

GPP:
GPP

TLAI:
TLAI

From this and the cprnc output it looks like the problem is just with the one grid cell, for this time point at least.  @rgknox suggestion I've got a longer term f45 grid running to see if the errors grow.  Barring an expansion of the diff to other grids and/or the increase in the difference, do you think this is reasonable to merge, @xuchongang ?

xuchongang:
@glemieux , for me, it is fine to merge.....

glemieux:
@rgknox had another thought on what might be causing the non-B4B results even when the stem drop parameter is zero.  The flushing logic looks like it could possibly flush only a portion of the `store_c_transfer_frac` since it's only checking for non-woody plants: https://github.com/NGEET/fates/pull/554/files#diff-b2580903c042bbb143bb267a7b882e3fR990-R1009

At this suggestion, I added a check during the flush to use all of the `store_c_transfer_frac` for leaves if the stem drop fraction is zero:
Drought leaf on section: https://github.com/glemieux/fates/blob/ca148d0ff5d4dd0470d17d9e51da2e2e90a005cd/biogeochem/EDPhysiologyMod.F90#L990-L1020
Cold leaf on section: https://github.com/glemieux/fates/blob/ca148d0ff5d4dd0470d17d9e51da2e2e90a005cd/biogeochem/EDPhysiologyMod.F90#L1103-L1135

This indeed resulted in B4B results.  The question is, should we be using `totalmemory` for leaves when stem drop is zero like I have done here?  Or is this changing the intent of the hypothesis?

xuchongang:
@glemieux , I think "using totalmemory for leaves when stem drop is zero" sounds reasonable to me.

glemieux:
Final testing results.  All expected PASS:
```
/glade/u/home/glemieux/scratch/clmed-tests/pr554-phen-flush-b4b-fix-Cf8f641ab-Fca148d0f.fates.cheyenne.intel
```

PR #564, opened 2019-08-15
Sort patches by biomass during inventory init if no age information present

### LINKED ISSUE #560: when doing inventory initialization, order the patches by the total biomass on the patches?
one of the outcomes of #553 is that we now get multiple patches when we do inventory initializations; prior to that, the size structure was there, but the patch heterogeneity was tending to get collapsed.  an effect of this is that, since all the patches get an age of zero when they are started, they are in kind of a random order in the linked-list structure.  in principle that's fine, as patch order shouldn't affect any of the fates processes, but it makes visualizing from restart files tricky.  I'd like to suggest that we add logic in the inventory initialization that it inserts each new patch into the linked list to sort the patches based on the amount of biomass in the patch, so that we end up with a linked-list order that roughly follows the oldest-to-youngest structure once the model gets going.

ckoven:
During inventory initialization, if we use actual census data where patch ages aren't available, the patches end up getting put into the linked list in whatever random order they are read in.  i principle that's fine, since the patch linked-list structure doesn't really do much, but for visualization purposes using the restart files, its nice to have the patches ordered in some sensible way from the start.  This code sorts the patches by total biomass instead of age, if age isn't present in the census file.

### Description:
fixes #560.

### Collaborators:
none

probably will be roundoff-level if census initialization is used, because the order of all patch operations will be different because the patches will be in a different order.  otherwise should be bit for bit.

### Checklist:

### Test Results:
Not formally tested yet. Informally, the code appears to do what it is supposed to do.

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/main/FatesInventoryInitMod.F90
+++ b/main/FatesInventoryInitMod.F90
@@ -92,6 +92,7 @@ contains
       use EDCohortDynamicsMod, only : fuse_cohorts
       use EDCohortDynamicsMod, only : sort_cohorts
       use EDcohortDynamicsMod, only : count_cohorts
+      use EDPatchDynamicsMod, only  : patch_pft_size_profile
 
       ! Arguments
       integer,            intent(in)               :: nsites
@@ -104,6 +105,8 @@ contains
       type(ed_cohort_type), pointer                :: currentcohort
       type(ed_patch_type), pointer                 :: newpatch
       type(ed_patch_type), pointer                 :: olderpatch
+      type(ed_patch_type), pointer                 :: head_of_unsorted_patch_list
+      type(ed_patch_type), pointer                 :: next_in_unsorted_patch_list
       integer                                      :: sitelist_file_unit   ! fortran file unit for site list
       integer                                      :: pss_file_unit        ! fortran file unit for the pss file
       integer                                      :: css_file_unit        ! fortran file unit for the css file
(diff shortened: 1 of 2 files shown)


main/FatesInventoryInitMod.F90
```diff
@@ -374,6 +377,82 @@ subroutine initialize_sites_by_inventory(nsites,sites,bc_in)
 
          deallocate(patch_pointer_vec,patch_name_vec)
 
+         ! now that we've read in the patch and cohort info, check to see if there is any real age info
+         if ( (sites(s)%youngest_patch%age .eq. sites(s)%oldest_patch%age) .and. &
```
rgknox: This is just a feeling I have, but I never like testing equivalence of real numbers.  I know in this case they are all read in as the same number, but I wonder if certain compilers or optimizations may mess up the last digit or something, and thus this equivalence would not work even if they numbers are the same in file.

I like comparing their absolute difference to nearzero. 

@glemieux @ckoven  am I just being paranoid here?

main/FatesInventoryInitMod.F90
```diff
@@ -374,6 +377,82 @@ subroutine initialize_sites_by_inventory(nsites,sites,bc_in)
...
+              associated(sites(s)%youngest_patch%older) ) then
+
+            ! so there are at least two patches and the oldest and youngest are the same age.
+            ! this means that sorting by age wasn't very useful.  try sorting by total biomass instead
```
rgknox: Code style is rock solid. Love the line lengths, love the descriptions, love the breaks.  Super readable. Gold star.

main/FatesInventoryInitMod.F90
ckoven: @rgknox just because you're paranoid doesn't mean they aren't after you.

main/FatesInventoryInitMod.F90
```diff
@@ -374,6 +377,82 @@ subroutine initialize_sites_by_inventory(nsites,sites,bc_in)
...
+
+            ! first calculate the biomass in each patch.  simplest way is to use the patch fusion criteria
+            currentpatch => sites(s)%youngest_patch
+            do while(associated(currentpatch))
+               call patch_pft_size_profile(currentPatch)
```
rgknox: Not that this has to happen in this PR. But this would be better as a function. It only updates a single array: currentPatch%pft_agb_profile(:,:)
It would be clearer from the calling function what it is changing.

main/FatesInventoryInitMod.F90
glemieux: @rgknox I think that makes sense.

main/FatesInventoryInitMod.F90
```diff
@@ -374,6 +377,82 @@ subroutine initialize_sites_by_inventory(nsites,sites,bc_in)
...
+               
+               ! check the two end-cases
+
+               ! Youngest Patch
+               if(sum(head_of_unsorted_patch_list%pft_agb_profile(:,:)) <= &
+                    sum(sites(s)%youngest_patch%pft_agb_profile(:,:)))then
+                  head_of_unsorted_patch_list%older                  => sites(s)%youngest_patch
+                  head_of_unsorted_patch_list%younger                => null()
+                  sites(s)%youngest_patch%younger => head_of_unsorted_patch_list
+                  sites(s)%youngest_patch         => head_of_unsorted_patch_list
```
glemieux: @ckoven and I chatted about line 417 and 418 in my effort to understand the linked lists better.  Based on some fortran test scripts I ran, it looks like we might not need line 417 here?  Going to test it out directly on my fork of Charlie's branch once Cheyenne is done being flaky.

main/FatesInventoryInitMod.F90
ekluzek: By the way, I do remember being taught in Numerical Analysis to do just what you suggest and NOT test for exact values but for within a tolerance, or using inequalities. So you are recommending a best practice. And experience definitely shows that floating point arithmetic really is out to get you!

glemieux: @ekluzek I'm taking that exact class right now!

glemieux:
I pre-merged master at 393b905468d93786735a340356d48babc3769988 into my checkout of Charlie's PR branch.  There were some changes in the PR branch that needed updating.  All expected PASS.  Only FAILs due to FatesHydro and MEMCOMP violations:

```
FAIL ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro RUN time=36
FAIL ERS_D_Mmpi-serial_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesColdDef TPUTCOMP Error: Computation time increase > 25 pct from baseline
FAIL SMS_Lm13.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesColdDef MEMCOMP Error: Memory usage increase > 10% from baseline
FAIL SMS_Lm13.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesColdDef TPUTCOMP Error: Computation time increase > 25 pct from baseline
FAIL SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro RUN time=21
```

CTSM (or) E3SM (specify which) test hash-tag: 9c12e402f44772f5db476681d23322741c07f759

CTSM (or) E3SM (specify which) baseline hash-tag: 9c12e402f44772f5db476681d23322741c07f759

FATES baseline hash-tag: 393b905468d93786735a340356d48babc3769988

Test Output: 
```
/glade/u/home/glemieux/scratch/clmed-tests/ckoven-sort-patches.fates.cheyenne.intel.C9c12e402-Fdc7f88e9
```

I'm waiting for a bci test run to complete to check out the veracity of removing the lines I noted above.

main/FatesInventoryInitMod.F90
rgknox: who said that?   I know someone is there ...

main/FatesInventoryInitMod.F90
glemieux: Disregard this comment.  After reviewing the output with @rgknox I realized that I was misreading what the algorithm was attempting to do in conjunction with the diagnostic output I was producing.

main/FatesInventoryInitMod.F90
```diff
@@ -374,6 +377,82 @@ subroutine initialize_sites_by_inventory(nsites,sites,bc_in)
...
+                             sum(currentpatch%pft_agb_profile(:,:)) .and. &
+                             sum(head_of_unsorted_patch_list%pft_agb_profile(:,:)) < &
+                             sum(olderpatch%pft_agb_profile(:,:))) then
+                           ! Set the new patches pointers
+                           head_of_unsorted_patch_list%older   => currentpatch%older
+                           head_of_unsorted_patch_list%younger => currentpatch
+                           ! Fix the patch's older pointer
+                           currentpatch%older => head_of_unsorted_patch_list
+                           ! Fix the older patch's younger pointer
+                           olderpatch%younger => head_of_unsorted_patch_list
```
glemieux: @ckoven do you think it's worth it to insert a `exit` statement in this inner loop to speed up the sorting?  I'm thinking of a situation in which you've moved head, but then current/older keeps itereting down the list via this `do while(associated(currentpatch))` loop.  There might come a later current/older pair that that head could fit between which would push the head further down the list, correct?

ckoven: yeah, that makes sense.

glemieux: Adding the `exit` cut the sort time in about half.  Without `exit`:
```
 BEGIN PATCH LOOP
 Number of patches:        1251
 Total Shifts of Head:      775976
```
Here 'Head' is `head_of_unsorted_patch_list`.  This output is here:
```
/home/glemieux/scratch/clmed-cases/ckoven-patch-sort-test-bci-5yr-iter.C137514abc-Fc9f25c1c.2019-10-01/run/logs/lnd.log.191001-165731
```

With `exit`:
```
 BEGIN PATCH LOOP
 Number of patches:        1251
 Total Shifts of Head:      406927
```
```
/home/glemieux/scratch/clmed-cases/ckoven-patch-sort-test-bci-5yr-iter-exit.C137514abc-Fc9f25c1c.2019-10-01/run/logs/lnd.log.191001-223448
```

Ran a compare test suite on Cheyenne on my branch and all expected passed.  Output is here:

```
/glade/u/home/glemieux/scratch/clmed-tests/ckoven-sort-patches-nearzero.fates.cheyenne.intel.C9c12e402-F7bbcd3ca
```

PR #576, opened 2019-09-12
update contributing.md to put link to google group on it.

ckoven:
Put a description of and link to the google group on the contributing.md page, so that people  (a) are aware of it and (b) can apply directly rather than have to email us about it.

No other code changes, so no testing needed.


rgknox:
I'm just going to integrate this. If we want to change again, we can just issue another PR.

Issue #574, opened 2019-09-12
'Nevercold' deciduous plants staging mysterious comeback (without climate triggers being triggered)


rosiealice:
Despite FATES best efforts with the 'nevercold' logic, it seems, sadly, that the cold deciduous plants are coming back to life in the tropics (seemingly without actually triggering any of our climate switches) :( 

At this point, my primary suspicion is to do with how PARTEH handles the status of new recruits, but I admit that I am not expert enough in PARTEH to figure out what happens with the initialization of cohort%status and bleaf (sorry for being behind on your efforts @rgknox)

Anyway, I've put all this in the form of a short story on Jupyter.  I hope you enjoy it...
https://github.com/rosiealice/jupyter_notes/blob/master/python_scripts/FATES_cold_dec.ipynb

(noting that I've been having intermittant success with getting these notebooks to render on git, it seems like it might just be a thing which is flaky/arbitrarily variable in time)

rgknox:
Still looking through code to see if I can understand what is happening.  I see that the nevercold flag seems to be correct, but indeed there should not be leaves.

PARTEH doesn't actually know what the status of a cohort is.  

Carbon should only flow to leaves if it fills these conditions:

1) if the storepriority parameter is non-zero and the plant is evergreen, some carbon to flow to build leaves, even if there has been no GPP.  (this does not seem to be the case, as you don't have evergreens)

https://github.com/NGEET/fates/blob/master/parteh/PRTAllometricCarbonMod.F90#L475



2) After that, the plant needs to have positive carbon balance to build more leaves, which should not be possible unless there was bud-burst.

https://github.com/NGEET/fates/blob/master/parteh/PRTAllometricCarbonMod.F90#L546


So my guess, is there is something weird going on with either the recruitment, or bud-burst. Will look at that next.

rgknox:
Bud burst only happens in two locations:
For cold deciduous here:
https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L947

For drought deciduous here:
https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L1007

I double checked those calls, and it seems that you won't get flushing during the nevercold status.  Seems it should only trigger when the site is in a phen_cstat_notcold state.

https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L933

Let me know if it looks like I'm missing anything.

rgknox:
[...]

PR #579, opened 2019-10-01
leaf status and phenology fixes

rgknox:
### Description:

This set of changes implements some phenology fixes, where 1) plants in nevercold status were surviving their initial leaf drops and persisting, 2) the zeroing of gdd counters was fixed to prevent re-flushing and 3) timer windows were adjusted.

Details are provided in #574 

### Collaborators:
main author: @rosiealice 
@rgknox 
@ckoven 

This will change answers on default parameter sets for simulations lasting orders larger than months.  This answer changes should be considered bug fixes though, and only targetted cold-deciduous plants with incorrect phenology cycles.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/parteh/PRTAllometricCarbonMod.F90
+++ b/parteh/PRTAllometricCarbonMod.F90
@@ -91,7 +91,8 @@ module PRTAllometricCarbonMod
 
   integer, public, parameter :: ac_bc_in_id_pft   = 1   ! Index for the PFT input BC
   integer, public, parameter :: ac_bc_in_id_ctrim = 2   ! Index for the canopy trim function
-  integer, parameter         :: num_bc_in         = 2   ! Number of input boundary condition
+  integer, public, parameter :: ac_bc_in_id_lstat = 3   ! Leaf status (on or off)
+  integer, parameter         :: num_bc_in         = 3   ! Number of input boundary condition
 
   ! THere are no purely output boundary conditions
   integer, parameter         :: num_bc_out        = 0   ! Number of purely output boundary condtions
@@ -353,6 +354,7 @@ contains
     integer  :: i_var                 ! index for iterating state variables
     integer  :: i_age                 ! index for iterating leaf ages
     integer  :: nleafage              ! number of leaf age classifications
+    integer  :: leaf_status           ! are leaves on (2) or off (1) 
     real(r8) :: leaf_age_flux         ! carbon mass flux between leaf age classification pools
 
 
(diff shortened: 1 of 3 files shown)


glemieux:
This all looks good to me.  I noted in the discussion in #574 that it sounds like there are still some uncertainties that might need to be addressed in other issues, correct?

parteh/PRTAllometricCarbonMod.F90
```diff
@@ -456,7 +459,11 @@ subroutine DailyPRTAllometricCarbon(this)
     call bdead_allom( target_agw_c, target_bgw_c, target_sapw_c, ipft, target_struct_c)
     
     ! Target leaf biomass according to allometry and trimming
-    call bleaf(dbh,ipft,canopy_trim,target_leaf_c)
+    if(leaf_status==2) then
```
ckoven: maybe this is for another PR, but it would be nice to give names to these integer phenology flags in https://github.com/NGEET/fates/blob/master/main/FatesConstantsMod.F90 and use them throughout the code.  Whenever I look at phenology code I always find myself having to look up what the integers mean.

rgknox: Hmmm, this is the label we would want instead of "2":
https://github.com/NGEET/fates/blob/master/main/EDTypesMod.F90#L80

The only problem is that this parteh module, with the interest of modularity in mind, does not access EDTypesMod.  It does access FatesConstantsMod.F90, perhaps we could start putting a lot of these parameter constants (labels) in that file...  I actually like that a lot.

Here is the top of the module where we list the usable modules:
https://github.com/NGEET/fates/blob/master/parteh/PRTAllometricCarbonMod.F90#L14

rgknox: I could create an issue suggesting we migrate parameter constants in EDTypesMod.F90 to FatesConstantsMod.  @ckoven what say you?

ckoven: yes agreed.  for some github reason that I don't understand my comments above were marked as pending, but, yes it seems like we both arrived at the same conclusion independently.

glemieux:
Long-term regression tests FAIL as expected:

CTSM (or) E3SM (specify which) test hash-tag: 9c12e402
CTSM (or) E3SM (specify which) baseline hash-tag: 9c12e402
FATES baseline hash-tag: 1ad93c311ed1a5d45df656b9031c8714a7071e64
Test Output: 
```
/glade/u/home/glemieux/scratch/clmed-tests/fix-leaf-status.fates.cheyenne.intel.C9c12e402-F8876d8e5
```

PR #590, opened 2019-10-31
Hydro 2D Solve

rgknox:
### Description:

This set of changes by @pnlfang solves the plant hydraulic fluxes through 1 plant and a portion of the soil media, by allowing all soil layers to by solved simultaneously.  This is labeled 2D becauase the other dimension in each layer is the rhizosphere shell.  The numerical method used in the solution is LU decomposition.

### Collaborators:
This PR is submitted on behalf of @pnlfang.  Discussions with @xuchongang , @bchristo, @rgknox and @ncollier

This should only generate different answers if the 2D method is selected.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:


xuchongang:
Sorry I missed this pull request and will do next week.

rgknox:
One thing to note @xuchongang , @pnlfang, @ckoven and I thought it would be good to use a new source for the LU solver. Our leading candidate right now is to use lapack or blas (I think it can be found there).

rgknox:
Question for community:
The 2D solve for plant hydraulics is currently turned on by setting the new namelist variable:
'use_alt_planthydro' which is passed from the HLM to FATES.

https://github.com/rgknox/fates/blob/pnlfang-newhydrosolve-merged/main/FatesInterfaceMod.F90#L142

There are alternatives to how we can set this feature (and others?)

1) we could modify the existing namelist parameter use_fates_hydro to **not** be true/false, but instead an integer or strings to specify the specific hydraulics hypothesis.  For instance, standard 1D hydro could be 'CX' (names) or 'TFS' (model basis?) or ', 2D hydro could be 'CXF2D' or 'TFS2D' or something.  We would then change the namelist parameter name to fates_hydro_mode.  The default setting would be 'OFF', or 'SBTRAN' (soil btran ie off).

2) or we could add a parameter to the fates parameter file that toggles the 2D mode on or off, but I would argue that since this method is potentially only valid when use_fates_hydro is true, and only valid when using Brad/Chonggang's method, it would be a contextually valid parameter which isn't always so great (parameter bloat).

Tagging: @pnlfang @bchristo @xuchongang @ckoven @rosiealice @glemieux

glemieux:
Option 1 sound like a good plan.  I typically prefer to have strings for option setting as I feel they are more transparent for the user.

bchristo:
Agreed plan 1 sounds best.

rgknox:
For this PR I will make the 1D vs 2D switch an "in-code" parameter. Adding the switch to the namelist requires a major API change, thus, I will add those changes to the nutrient enabling PR (to be issued).  

Here is what changing the namelist switch looks like on the FATES side:

https://github.com/rgknox/fates/compare/parameters-eca-phen-fire...rgknox:parameters-nlist-eca-phen-fire

rgknox:
After some deliberation, my sense is that the refactors I am currently working on will provide a nice clean location for us to switch between the matrix-newton solve presented here, and the sequential solve.  That branch (https://github.com/rgknox/fates/tree/hydro-diagnostics-wmat ) .  There are also many things that are refactored there that should be applied to this scheme (calculation of water potentials, memory handling, calculation of maximum and effective conductances).  We can always re-open this PR if need be.
Note on the other branch, I have been working closely with @pnlfang on understanding and integrating the new work.  That branch also maintains credit of her code changes.

Closing.

Issue #589, opened 2019-10-29
Connect fire events to ignitions


jkshuman:
Within SPITFIRE there is not a connection between a fire events and an ignition event. As it stands, if it can burn, it will burn regardless of ignitions. Going to add in a connection to ignitions. 
(if FI exceeds fire threshold and ignitions are greater than 0 then fire ==true) Can I get a confirmation that I am not missing something here...? @rosiealice @rgknox @adrifoster @pollybuotte @slevisconsulting  

https://github.com/jkshuman/fates/blob/1ad93c311ed1a5d45df656b9031c8714a7071e64/fire/SFMainMod.F90#L687-L688
This is a larger problem because it is then carried through into tree fire-damage and tree mortality, if it can burn, then the trees might be fire-damaged, regardless of the presence of ignitions. Might explain some of the excessive tree death.
https://github.com/jkshuman/fates/blob/1ad93c311ed1a5d45df656b9031c8714a7071e64/fire/SFMainMod.F90#L963
https://github.com/jkshuman/fates/blob/1ad93c311ed1a5d45df656b9031c8714a7071e64/fire/SFMainMod.F90#L1013

rosiealice:
Shouldn't there be a link between the lightening counts (real number) and the fire area? As per:
https://github.com/NGEET/fates/blob/87ce240691a6c1446b847a9faffcb5e450e4596d/fire/SFMainMod.F90#L788
So the ignition counts do always make a fire, but the amount of fire is variable. I guess what you're getting it is some scalar to account for the fact that not every lightening strike actually causes a fire? ('Lightening strike efficiency' or something like that?)

adrifoster:
In Thonicke 2010 AB is then incorporated into the FI calculation 

FI = h * FC/Ab_frac * ROS/60.0

For UVAFME I did not include this area burnt impact on FI, but used ignition rate times FDI as a "probability" of fire ignition. Fire can only occur if that ignition event happens (and then following that FI exceeds the threshold).

jkshuman:
@rosiealice the AB calculation is indeed connected to ignitions as you notice. The calculation for crown damage and tree mortality is connected only to fire==true on the patch. This is set in the Fire Intensity subroutine: https://github.com/NGEET/fates/blob/87ce240691a6c1446b847a9faffcb5e450e4596d/fire/SFMainMod.F90#L687-L688
[...]

PR #593, opened 2019-11-07
CG scaler on lightning, NF>0 for fire event

jkshuman:
Add scaler for cloud to ground lightning strikes, and strikes >0 for fire event

### Description:
Portion of strikes that are cloud to ground (CG) included in calculation of ignitions per Thonicke
et al 2010 and Latham and Williams (2001).
Calculation of ignitions moved to fire_intensity subroutine. Ignitions greater than zero included 
as condition for successful fire event. 
This address issue #589 

### Collaborators:
@rgknox @slevisconsulting @rosiealice @adrifoster 

This will change answers with fire active. Specifically, it will reduce fire area. 

### Checklist:

### Test Results:
Simulations compiled and run across South America with and without fire. Comparison between prior default ignitions and scaled cloud to ground ignitions included.

Test Output:
Difference in Burned Fraction for year 4 (cloud to ground ignitions - all ignitions "default")
Ignitions_CG_ign-all_ign_yr4

### DIFF
--- a/fire/SFMainMod.F90
+++ b/fire/SFMainMod.F90
@@ -676,7 +676,6 @@ contains
     real(r8),parameter :: CG_strikes = .20_r8      !cloud to ground lightning strikes
                                                    !Latham and Williams (2001)
 
-    currentSite%NF = 0.0_r8
     !NF = number of lighting strikes per day per km2
     currentSite%NF = ED_val_nignitions * years_per_day * CG_strikes
 
--- a/fire/SFMainMod.F90
+++ b/fire/SFMainMod.F90
@@ -658,30 +658,30 @@ contains
 
     !currentPatch%FI  avg fire intensity of flaming front during day. Backward ROS plays no role here. kJ/m/s or kW/m.
     !currentSite%FDI  probability that an ignition will start a fire
+    !currentSite%NF   number of lighting strikes per day per km2
     !currentPatch%ROS_front  forward ROS (m/min) 
     !currentPatch%TFC_ROS total fuel consumed by flaming front (kgC/m2)
 
     use FatesInterfaceMod, only : hlm_use_spitfire
     use EDParamsMod,       only : ED_val_nignitions
     use FatesConstantsMod, only : years_per_day
-    use SFParamsMod,  only : SF_val_fdi_alpha,SF_val_fuel_energy, &
+    use SFParamsMod,       only : SF_val_fdi_alpha,SF_val_fuel_energy, &
          SF_val_max_durat, SF_val_durat_slope
 
     type(ed_site_type), intent(inout), target :: currentSite
-
     type(ed_patch_type), pointer :: currentPatch
 
     real(r8) ROS !m/s
     real(r8) W   !kgBiomass/m2
-    real(r8) NF  !number of lighting strikes per day per km2
     real(r8),parameter :: CG_strikes = .20_r8      !cloud to ground lightning strikes
                                                    !Latham and Williams (2001)
 
+    currentSite%NF = 0.0_r8
     !NF = number of lighting strikes per day per km2
-    NF = ED_val_nignitions * years_per_day * CG_strikes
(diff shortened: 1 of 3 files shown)


rgknox:
Glad to see that the calculation of burned area is contingent on having %fire = 1.  I think this was causing problems in the previous method (swapping routine order).

https://github.com/NGEET/fates/blob/20944a616455a032baeccea1bb0da9c00aec9bc5/fire/SFMainMod.F90#L752

fire/SFMainMod.F90
```diff
@@ -657,19 +658,30 @@ subroutine  fire_intensity ( currentSite )
...
     type(ed_site_type), intent(inout), target :: currentSite
-
     type(ed_patch_type), pointer :: currentPatch
 
     real(r8) ROS !m/s
     real(r8) W   !kgBiomass/m2
+    real(r8),parameter :: CG_strikes = .20_r8      !cloud to ground lightning strikes
+                                                   !Latham and Williams (2001)
+
+    currentSite%NF = 0.0_r8
```
rgknox: This zero can be removed.

jkshuman:
@slevisconsulting @pollybuotte link to these changes inside the read-lightning branch
https://github.com/jkshuman/fates/tree/FI-CG-read-lightning-tg130api8
I will confirm that this is also less in a bit.

rgknox:
One minor suggestion about zero'ing. I think this PR looks good.  Note that adding %NF to the site structure will not require an addition to the restarts, since it is only used,  during fire calculations, and is calculated in that scheme as well.

jkshuman:
@rgknox I am going to rework this again with the frac_burnt as part of calculating fire_intensity (FI) in a separate PR.  I will maintain this contingency in there as well. Yeah, that other implementation was a giant rabbit hole of problems...

main/EDInitMod.F90
```diff
@@ -280,6 +281,7 @@ subroutine set_site_properties( nsites, sites )
           sites(s)%dstatus = dstat
           
           sites(s)%acc_NI     = acc_NI
+          sites(s)%NF         = 0.0_r8
```
jkshuman: @rgknox just to confirm: you are saying to remove this line?

rgknox: I think this is fine, I don't think it will have an effect. 
Looks like you already fixed the line I was talking about.

jkshuman: OK. I was a little unclear on what was necessary in EDInit, so happy to take your input on if it needs to be changed.

fire/SFMainMod.F90
```diff
@@ -684,7 +695,7 @@ subroutine  fire_intensity ( currentSite )
           if( hlm_masterproc == itrue ) write(fates_log(),*) 'fire_intensity',currentPatch%fi,W,currentPatch%ROS_front
        endif
        !'decide_fire' subroutine shortened and put in here... 
-       if (currentPatch%FI >= fire_threshold) then  ! 50kW/m is the threshold for a self-sustaining fire
+       if (currentPatch%FI >= fire_threshold .and. currentSite%NF > 0._r8) then !50kW/m threshold for self-sustaining fire
```
slevisconsulting: @jkshuman I see two new lines of code that sum up your changes in this pull request:
1) `currentSite%NF = ED_val_nignitions * years_per_day * CG_strikes` (currently line 680)
2) `if (currentPatch%FI >= fire_threshold .and. currentSite%NF > 0._r8) then` (currently line 698)
The rest is either infrastructure to support these changes or aesthetic improvements.

Assuming that your testing continues to pass, these changes seem fine to me.

rgknox:
We can start testing this after review.  Would anyone else like to chime in and sign off?  @jkshuman, any other stake-holders you would like to check-in here?

jkshuman:
Thanks for checking in @rgknox. I think with your review and @slevisconsulting we are good. This version is a basic update, so I am not worried about the logic.

jkshuman:
these changes included in PR #594 . 
closing

Issue #476, opened 2019-02-13
update alpha_SH to be a set value, not PFT specific


jkshuman:
@pollybuotte and I were talking through fire parameters. The alpha_SH parameter is currently set to be a PFT specific value, so that the flame height varies based on the PFT.
                   
   EDPftvarcon_inst%fire_alpha_SH(currentCohort%pft) * (currentPatch%FI**0.667_r8) 

https://github.com/jkshuman/fates/blob/e9eb4de6cac81be6c4254031538bb61c44861ae4/fire/SFMainMod.F90#L871-L873

The scorch height is calculated based on this PFT parameter and fire intensity. Fire intensity is calculated from fuel energy and rate of spread. It doesn't make sense to have a fire traveling through a mixed species forest where the flame height changes within a patch based on the PFT. The alpha scorch height parameter in Thonicke et al 2010 made some sense to use at the PFT level as they were using PFT tiles without mixing of PFTs. The mixed PFT composition of FATES should not use a PFT dependent scorch height. Rather scorch height should be computed by the 2/3 law of Byram 1959 relating scorch height and fire intensity. This linear relationship can be modified with a coefficient which might vary regionally. Van Wagner tested this for conifer systems. Suggest looking into VanWagner 1973 (has details of Byram 1959), and the application by Peterson and Ryan (1984) or by Shuman et al 2017.

rosiealice:
Makes sens to me...

rosiealice:
So I read this before I read issue #477. The complication there is to do with the weighting of the SH across PFTs, which makes some attempt to deal with this problem, at least mathematically, but might not make sense scientifically.  The question, I guess, is what we actually mean to represent with the scorch height parameter? Is it a feature of the litter, or of something like crown shape, or is it more to do with the overall canopy structure?
Good that you and @pollybuotte have been getting into the nitty-gritty of this!

jkshuman:
Scorch height (SH) is used specifically to capture tree crown damage as it contributes to tree mortality. From the literature there is an established connection between SH and fire intensity (FI) for forest and shrubland systems (Keeley 2009; Alexander and Cruz 2012; Fernandes et al 2000; Wade 1993; Burrows 1995; Johnson 1992;  Andrews and Rothermal 1982; Van Wagner 1973; Byram 1959). 

[...]

PR #594, opened 2019-11-25
SPITFIRE - ignitions, area burn, intensity, scorch height

jkshuman:
### Description:
This set of changes addresses fire behavior and effects updates to equations. A cloud to ground scaler on lightning ignitions is introduced per Latham and Williams (2001) (addresses #593). Ignitions must be greater than 0 to begin calculations of fire behavior (addresses issue #589).  Area burn and fire intensity subroutines are combined to allow for inclusion of fraction of patch burnt in calculation of fire intensity (FI) per Thonicke et al 2010. Ignition probability (FDI), fire duration (FD), area burnt (AB) and frac_burnt now calculated prior to FI to include this dependancy. Scorch height is updated to be at the cohort level and scaled by PFT trait "alpha_SH" using the patch level fire intensity (FI) per Van Wagner (1973) (addresses issue #477 and #476). This update maintains the fire threshold of 50 kW/m for fire events per Pyne (1986). 

This addresses issues #593 #589 #477 and #476

### Collaborators:
Discussions with @adrifoster @pollybuotte @rgknox @rosiealice @glemieux 

Updates will result in less ignitions, less area burn, and lower fire intensity compared to current master. Attached figure with difference between branch with update to ignitions only and branch using all ignitions for burned area (BA) and GPP after 4 years of simulation.
Ignitions_CG_ign-all_ign_yr4

### Checklist:

### Test Results:
TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDCohortDynamicsMod.F90
+++ b/biogeochem/EDCohortDynamicsMod.F90
@@ -580,7 +580,7 @@ contains
 
     currentcohort%year_net_uptake(:) = 999._r8 ! this needs to be 999, or trimming of new cohorts will break. 
     currentcohort%ts_net_uptake(:)   = 0._r8
-    currentcohort%sh                 = 0._r8
+    currentcohort%scorch_ht          = 0._r8
     currentcohort%fraction_crown_burned = 0._r8 
     currentCohort%size_class            = 1
     currentCohort%seed_prod          = 0._r8
@@ -1655,7 +1655,7 @@ contains
     n%ddbhdt          = o%ddbhdt
 
     ! FIRE
-    n%sh                    = o%sh  
+    n%scorch_ht             = o%scorch_ht  
     n%fraction_crown_burned = o%fraction_crown_burned
     n%fire_mort             = o%fire_mort
     n%crownfire_mort        = o%crownfire_mort
--- a/biogeochem/EDCohortDynamicsMod.F90
+++ b/biogeochem/EDCohortDynamicsMod.F90
@@ -534,7 +534,7 @@ contains
     currentCohort%ddbhdt             = nan ! time derivative of dbh 
 
     ! FIRE
-    currentCohort%sh                    = nan ! scorch height affecting crown (m)
+    currentCohort%scorch_ht             = nan ! scorch height affecting crown (m)
     currentCohort%fraction_crown_burned = nan ! proportion of crown affected by fire
     currentCohort%cambial_mort          = nan ! probability that trees dies due to cambial char P&R (1986)
     currentCohort%crownfire_mort        = nan ! probability of tree post-fire mortality due to crown scorch
--- a/biogeochem/EDCohortDynamicsMod.F90
+++ b/biogeochem/EDCohortDynamicsMod.F90
@@ -534,7 +534,7 @@ contains
     currentCohort%ddbhdt             = nan ! time derivative of dbh 
 
     ! FIRE
-    currentCohort%scorch_ht             = nan ! scorch height affecting crown (m)
+    currentCohort%Scorch_ht             = nan ! scorch height affecting crown (m)
     currentCohort%fraction_crown_burned = nan ! proportion of crown affected by fire
     currentCohort%cambial_mort          = nan ! probability that trees dies due to cambial char P&R (1986)
     currentCohort%crownfire_mort        = nan ! probability of tree post-fire mortality due to crown scorch
--- a/biogeochem/EDCohortDynamicsMod.F90
+++ b/biogeochem/EDCohortDynamicsMod.F90
@@ -534,6 +534,7 @@ contains
     currentCohort%ddbhdt             = nan ! time derivative of dbh 
 
     ! FIRE
+    currentCohort%sh                    = nan ! scorch height affecting crown (m)
     currentCohort%fraction_crown_burned = nan ! proportion of crown affected by fire
     currentCohort%cambial_mort          = nan ! probability that trees dies due to cambial char P&R (1986)
     currentCohort%crownfire_mort        = nan ! probability of tree post-fire mortality due to crown scorch
@@ -579,6 +580,7 @@ contains
 
     currentcohort%year_net_uptake(:) = 999._r8 ! this needs to be 999, or trimming of new cohorts will break. 
     currentcohort%ts_net_uptake(:)   = 0._r8
+    currentcohort%sh                 = 0._r8
     currentcohort%fraction_crown_burned = 0._r8 
     currentCohort%size_class            = 1
     currentCohort%seed_prod          = 0._r8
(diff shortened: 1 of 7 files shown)


biogeochem/EDCohortDynamicsMod.F90
```diff
@@ -534,6 +534,7 @@ subroutine nan_cohort(cc_p)
     currentCohort%ddbhdt             = nan ! time derivative of dbh 
 
     ! FIRE
+    currentCohort%sh                    = nan ! scorch height affecting crown (m)
```
rgknox: could we make this variable a little more verbose? When I see sh without context, my mind goes to specific humidity.

jkshuman: yep. will update now.

jkshuman:
@rgknox thank you for the thorough review of these changes, and glad to hear they pass muster.

I agree that the commented out portion of the ellipse fire shape and size feedback isn't great for readability. I do plan on improving this section of the code in the near future, so I am inclined to leave it here for the moment.

The fire_threshold does have actual meaning, and the active line uses a value from the literature (Pyne 1986) that identifies fires that go out on their own (too small to spread). The fire-threshold set close to zero, was commented out, so I removed that. Tracking all fires with energy close to zero needs more evaluation of the impact on area burnt and other values (veg mortality, etc). I will be cleaning that section of the code in the near future. (This will likely overlap with the modification of the ellipse fire shape and size portion of the code.)

glemieux:
@jkshuman some of the current standard test suite call the `SCORCH_HEIGHT history variable`:
https://github.com/ESCOMP/CTSM/blob/b350e06e4615c124d3ef1d8f3f9831a1f8f3f402/cime_config/testdefs/testmods_dirs/clm/Fates/user_nl_clm#L1-L7
As such, the test runs are crashing trying to find it.  Do you want to keep `SCORCH_HEIGHT` available for future updates, or should I create a PR to update the test configuration files to remove it?

jkshuman:
@glemieux with the switch of scorch_ht to the cohort (PFT) level, we don't need this (scorch height) for future updates. You can get scorch height from the fire intensity. So, yes, my recommendation is to remove it from the test configuration files.

glemieux:
> @glemieux with the switch of scorch_ht to the cohort (PFT) level, we don't need this (scorch height) for future updates. You can get scorch height from the fire intensity. So, yes, my recommendation is to remove it from the test configuration files.

@rgknox do we want to remove the commented lines out of FatesHistoryInterfaceMod.F90 given Jackie's recommendation?

jkshuman:
@rgknox @glemieux oops. yeah, meant to check in with folks on that. The switch to cohort level made the previous patch level use invalid. Got hasty in my implementation. Let me know if you want me to do that, or if you want to open up this decision to more input. I can add it back in with cohort level summary in a future PR, but it should not be an active output.

glemieux:
> I can add it back in with cohort level summary in a future PR, but it should not be an active output.

Sorry for the newbie question @jkshuman : how would a cohort level summary be implemented?

jkshuman:
@glemieux not a bad question. I believe this was why I left it out. I ran out of time to investigate the answer to that question for myself in the mad dash prior to AGU...

jkshuman:
@glemieux I also don't think it is absolutely necessary, as you can get this from fire intensity. But, I should have cleaned up the history file interface. I think I left it there so we could have this discussion about it being removed...I wanted to revisit when I wasn't in a panic for time.

ckoven:
@jkshuman lets just delete it the commented history code.  we can always pull it out of old commits if someone finds they have a need for it.

pollybuotte:
For what it's worth @jkshuman @glemieux , I'll want to look at scorch height as we compare the active and passive crown fire implementations. If it's not a history variable, is there a place to hold the calculation from fire intensity that will be readily available (i.e. so we don't have to ask you, Jackie!)?

jkshuman:
@pollybuotte that calculation is in the spitfire code
https://github.com/jkshuman/fates/blob/e39aa5f27f2d88b8bae6ecf1b906b5f4fdd31496/fire/SFMainMod.F90#L839-L840
@ckoven sure thing. I will can remove those lines.

ckoven:
@pollybuotte @jkshuman it looks like this is really fundamentally a patch x PFT property rather  than a cohort property.  in principle we have the machinery to output such variables; e.g. see here: https://github.com/NGEET/fates/blob/e39aa5f27f2d88b8bae6ecf1b906b5f4fdd31496/main/FatesHistoryInterfaceMod.F90#L2201

in the interest of limiting PR scope-creep, maybe let's add that to a near-future PR?

jkshuman:
@ckoven you are correct, and sounds good to me.

glemieux:
Chatting with @ckoven prior to the all hands discussion today, we decided that since the three test definitions that call `SCORCH_HEIGHT` also call another fire variable (i.e. `FIRE_INTENSITY`), we can likely be safe in striking `SCORCH_HEIGHT` from the test definitions even though we may reuse that variable name in the future PR.

jkshuman:
> @pollybuotte @jkshuman it looks like this is really fundamentally a patch x PFT property rather than a cohort property. in principle we have the machinery to output such variables; e.g. see here:
> 
> https://github.com/NGEET/fates/blob/e39aa5f27f2d88b8bae6ecf1b906b5f4fdd31496/main/FatesHistoryInterfaceMod.F90#L2201
> 
> in the interest of limiting PR scope-creep, maybe let's add that to a near-future PR?

@ckoven if this is not too time consuming per your example, I can add this into the history interface with this PR. and then we would not need to update the testing definitions @glemieux ?
though, I have somehow broken my Hobart git environment, and cannot create cases at the moment, so I am in a holding pattern until I can fix that.

glemieux:
@jkshuman that would make things slightly less complex from my standpoint, since I wouldn't have to coordinate a PR with the HLM.  BTW, I did run a test on my local branch with 'SCORCH_HEIGHT` removed from the test definitions and everything passed as expected (only DIFF being in `FIRE_INTENSITY`).

ckoven:
@jkshuman sounds good.  actually if its ok with you, I can just write that history code myself and submit a PR to your branch?  I'm keen to redo an old set of gridded california simulations with your new fire code and so can do this first before I kick off runs...

jkshuman:
@ckoven that would be great, as I need to fix my git environment. let me know how things look in CA. weird cime error.

glemieux:
Pulled in the updates and re-testing.

glemieux:
Test Results:

All expected PASS.  FAILs only on FatesHydro RUN and test cases with DIFFs for `FIRE_INTENSITY`.

CTSM (or) E3SM (specify which) test hash-tag: 5ad8a8d2

CTSM (or) E3SM (specify which) baseline hash-tag: 097f4239

FATES baseline hash-tag: 17aaa879

Test Output:
```
/glade/u/home/glemieux/scratch/clmed-tests/pr594-fire-threshold.fates.cheyenne.intel.C5ad8a8d2-F931824c8
```

Issue #591, opened 2019-11-06
parameters seed_germination_timescale and seed_decay turnover mixup?


evaleriksen:
I'm going through a list of parameters from a google sheet called FATES parameters, and comparing that list to the actual parameters in [fates_params_default.cdl](https://github.com/NGEET/fates/blob/master/parameter_files/fates_params_default.cdl). 

In the (older?) google sheet, two of the parameters look like this:

param|long name
--|--
fates_seed_germination_timescale | turnover time for seeds with respect to germination
fates_seed_decay_turnover | turnover time for seeds with respect to decay

In the code, the long names are switched, so the germination_timescale has long name with "decay" and the decay_turnover har long name with "germination". Like this:

	double fates_seed_decay_turnover(fates_pft) ;
		fates_seed_decay_turnover:units = "1/yr" ;
		fates_seed_decay_turnover:long_name = "turnover time for seeds with respect to germination" ;
	double fates_seed_germination_timescale(fates_pft) ;
		fates_seed_germination_timescale:units = "1/yr" ;
		fates_seed_germination_timescale:long_name = "turnover time for seeds with respect to decay" ;

Which is correct? 
In the sheet they seem to have very similar values (0.5 and 0.51) but if the long name is wrong in the code maybe it's worth checking that the parameter values are not switched as well.

rgknox:
Hi @evaleriksen,

I don't have a sense of what are reasonable values here, but maybe @lmkueppers might push us in the right direction?

We should indeed update the meta data in the file (and in the code where they are defined).  Any suggestions on new definitions?

```
	double fates_seed_decay_turnover(fates_pft) ;
		fates_seed_decay_turnover:units = "1/yr" ;
		fates_seed_decay_turnover:long_name = "turnover time for seeds with respect to decay" ;
	double fates_seed_germination_timescale(fates_pft) ;
		fates_seed_germination_timescale:units = "1/yr" ;
		fates_seed_germination_timescale:long_name = "germination timescale for seeds (fraction germinated per year)" ;
```

evaleriksen:
Hi @rgknox, 
we are going through the list of parameters for a couple of small projects here and aim to contribute something to the documentation in the other end. 

Right now I haven't really understood exactly how these parameters are used in the model or what they represent, so I don't have suggestions for new definitions but thought it might be useful to raise this issue in case there is a mix-up in the model with these two parameters.

evaleriksen:
[...]

Issue #581, opened 2019-10-10
Parameter file definition edits


pollybuotte:
@rgknox Here are a few minor edits needed when the next round of parameter file updates happens:

           double fates_branch_turnover(fates_pft) ;
		fates_branch_turnover:units = "yr-1" ; 
		fates_branch_turnover:long_name = "turnover time of branches" ;

Units should be "yr".  And here,

	double fates_seed_decay_turnover(fates_pft) ;
		fates_seed_decay_turnover:units = "1/yr" ;
		fates_seed_decay_turnover:long_name = "turnover time for seeds with respect to germination" ;
	double fates_seed_germination_timescale(fates_pft) ;
		fates_seed_germination_timescale:units = "1/yr" ;
		fates_seed_germination_timescale:long_name = "turnover time for seeds with respect to decay" ;

The long_names are switched.

rgknox:
I believe these have been addressed, see:
https://github.com/NGEET/fates/blob/master/parameter_files/fates_params_default.cdl#L126
https://github.com/NGEET/fates/blob/master/parameter_files/fates_params_default.cdl#L448
https://github.com/NGEET/fates/blob/master/parameter_files/fates_params_default.cdl#L451

@pollybuotte , feel free to re-open if you see fit

Issue #573, opened 2019-09-05
Active crown fire development


slevisconsulting:

jkshuman:
@slevisconsulting and I spoke at great length about the implementation of active crown fire, and we have a plan.

https://github.com/jkshuman/fates/blob/ae0a5e7196fa9219b37712c140c55fa7a0c2d51b/fire/SFMainMod.F90#L930
Within subroutine crown damage inside the do loop across woody cohorts, we will
1: calculate EQ 12 from Bessie and Johnson (1995) (critical intensity for active crown fire "active_crown_FI")

2: evaluate EQ 14 from Bessie and Johnson (ratio of Patch%FI/cohort%active_crown_FI)

3: if EQ 14 (Patch%FI/cohort%active_crown_FI  )is greater than 1 and (Patch%FI/cohort%passive_crown_FI) greater than 1 
https://github.com/jkshuman/fates/blob/ae0a5e7196fa9219b37712c140c55fa7a0c2d51b/fire/SFMainMod.F90#L944-L945

4: then active crown fire = true and crown foliage is available as fuel (Bessie and Johnson 1995)

5:  loop through all cohorts and update this active crown fire flag

6: update Patch%sum_fuel to include leaf organ biomass (convert leaf organ C to kgbiomass/m2) 
        for all cohorts that have active_crown_fire_flg == true

7:  **recalculate** ROS_front (and ROS_back?) for altered ground fuel consumption, area burnt
       recalculate FI for altered scorch height

8:   **update crown damage** for cohorts where active_crown_fire_flg was false based on these new ROS, FI, SH values

9: end loop

This will only loop through and recalculate the fire behavior and effects once after finding an active crown fire. This is a good place to start for our model at a daily time step. We can re-visit with implementation and further testing. 

@rgknox: @slevisconsulting and I would like your feedback about implementing the recalculation of ROS, FI and SH in terms of where to call the subroutine, we we are in that cohort do loop. I guess we can evaluate the active_crown_fire flag before we do the final mortality probability?

tagging @lmkueppers @xuchongang @pollybuotte @ckoven @rosiealice @acf7m and any other interested parties for comment

slevisconsulting:
Some more notes for steps (1), (2), (3) mainly for my own benefit:

1) Add Bessie & Johnson (1995) eq. (12) right after @jkshuman added eq. (8):
`! Crown fuel ignition potential, EQ 8 Bessie and Johnson 1995`
[...]

PR #595, opened 2019-11-26
new and fixed parameters for nutrients, phenology, seeds, fire and age-mortality

rgknox:
### Description:

This set of changes focuses only on changes to the default parameter set.  
1) New parameters were added for ECA nutrient aquisition. The code will follow, and these parameters are currently unused
2) New parameters for non-woody stem phenology were added, this is coordinated with PR #554 
3) A new parameter that flags if PFTs can experience active crown fires was added, this is coordinated with PR #584 
4) Parameter names for seed germination and decay, as well as the units and metadata were fixed, see #591 and #581

### Collaborators:
@qzhu-lbl, @evaleriksen, @pollybuotte, @slevisconsulting , @jkshuman, @glemieux , @xuchongang, @ckoven @rosiealice , @JessicaNeedham 

These changes should not change answers

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


ckoven:
Hi all,
discussed this with @rgknox and @glemieux.  Main remaining question here is what to do about symbiotic N fixation parameters.  Right now this PR has two dummy parameters for N fixation, i.e.: https://github.com/NGEET/fates/pull/595/files#diff-a236898dbb7848d5cfdf2610b0ba32c0R321

Is this sufficiently flexible?  e.g., @rosiealice does this cover the PFT parameter needs for a FUN-like N fixation scheme?  I'm imagining one PFT parameter would be a scalar variable that specifies in some units how much carbon a plant might allocate to N fixation, with zero meaning the plant isn't an N fixer.  I could see this having a lower and upper bound to accommodate flexibility in representing obligate vs facultative N fixers, so therefore needing two PFT parameters?

Other PFT parameters might be the N return on the C cost, or the constants controlling the temperature sensitivity of N fixation costs, though I would expect those to be global rather than PFT parameters, and if so, maybe we can add those in once we have a fully developed N fixation model?  Does anyone see a need for more than 2 PFT-level N fixation parameters now as we build out the model?

cc @walkeranthonyp @laagee

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1274,14 +1274,14 @@ subroutine SeedGermination( litt, cold_stat, drought_stat )
 
     !----------------------------------------------------------------------
 
-    ! germination_timescale is being pulled to PFT parameter; units are 1/yr
+    ! germination_rate is being pulled to PFT parameter; units are 1/yr
     ! thus the mortality rate of seed -> recruit (in units of carbon) 
-    ! is seed_decay_turnover(p)/germination_timescale(p)
+    ! is seed_decay_turnover(p)/germination_rate(p)
```
pollybuotte: Should this now be seed_decay_rate instead of seed_decay_turnover?

main/EDPftvarcon.F90
```diff
@@ -49,6 +49,7 @@ module EDPftvarcon
                                                      ! that is occupied by crown. For fire model. 
      real(r8), allocatable :: bark_scaler(:)         ! scaler from dbh to bark thickness. For fire model.
      real(r8), allocatable :: crown_kill(:)          ! scaler on fire death. For fire model. 
+     real(r8), allocatable :: active_crown_fire(:)   ! Is plant susceptible to active crown fire?
```
jkshuman: Depending on how active_crown fire is implemented, this may be altered to a site level "flag" rather than a PFT level parameter. @lmkueppers @slevisconsulting @pollybuotte we should discuss on Friday.
@rgknox can I put a hold on this until we have that discussion?

slevisconsulting: @jkshuman I think your question can be summed up as this:
Will we use this flag as an active crown fire on/off switch for all pfts simultaneously?
OR
Will we use it at the pft level to assign varying pft vulnerability to active crown fire?

slevisconsulting: I prefer to keep the current flexibility of pft-level and am open to discussing.

ckoven: what are the arguments pro and con?

slevisconsulting: Pro for keeping pft-level:
The flexibility to say that, for example, ponderosa pine is more vulnerable to active crown fire than sugar maple by assigning different values between 0 and 1 to each pft.

No cons in my opinion, because we can keep using the pft-level flag as a simple on/off switch if we prefer.

rgknox: @jkshuman and @slevisconsulting , thanks for the feedback, @jkshuman does this sound reasonable, or do you feel like we should take the time now to swap to non-pft parameter here?

rgknox: I see no reason we can't wait to here back from your discussion on Friday, take your time.

biogeochem/EDPhysiologyMod.F90
rgknox: yes, nice find @pollybuotte , will fix

main/EDPftvarcon.F90
jkshuman: This is really a result of the transition from cohort-level implementation to patch-level, and the transition to calculating scorch height at the cohort-level.
Cons: how to justify PFT level implementation of a stand level behavior
Pros: ?

This is more a reflection of how crown fire happens in the stand at the patch level (so patch level not site-level). For the patch, you get active crown fire if you meet two conditions: 
1) there is enough fuel in the canopy to achieve passive crown initiation to propagate fire vertically through the canopy,  (fire intensity from canopy fuels >=  surface fire intensity)
and 2) the fire can spread from crown to crown (fire intensity using canopy rate of spread >= surface fire intensity)

Both of these conditions use patch level stand values from the total canopy, so controlling active crown fire at the PFT level may not be valid. The alpha_SH parameter already acts as a PFT level control on crown scorch, so I think working within the parameter space makes sense. This makes more sense with the transition to calculating scorch height at the cohort-level implemented in PR #594 and carried over into #584 and #572
                `                !Equation 16 in Thonicke et al. 2010 !Van Wagner 1973 EQ8 !2/3 Byram (1959)
                currentCohort%Scorch_ht = EDPftvarcon_inst%fire_alpha_SH(currentCohort%pft) * (currentPatch%FI**0.667_r8)` 


This does depend on the implementation of active crown fire and assumptions about patch size. In a model with a fixed patch size of "appropriate" size, when active crown fire ignites assuming 100% crown damage is reasonable. For FATES  this assumption may not make sense, as the area of application can be quite large. With implementation of a feedback where the fire intensity and behavior is recalculated with active fire, this would alter the scorch height and allow for use of the altered (higher) value. With large areas as @adrifoster noted in #573 stand density and wind are factors: >  "Spotting fires could bridge large spaces between tree crowns but this requires very high fire intensities (30,000 - 100,000+ kW/m). "

This comes back to using the existing PFT level alpha_SH value and recalculating fire intensity and behavior for the patch/stand.  Prior to implementing the feedback onto fire intensity and behavior, using 100% crown damage on the patch for active crown fire.

ckoven: Hi All,  we decided on our call today that this "active_crown_fire" flag should not be a PFT-level parameter, but should instead be either a global flag or a namelist variable (whichever is easier to write/maintain).  We just need to be able to turn it on or off for a given model run.

rgknox: thanks for the update.  I will add this to the parameter file.

rgknox:
updates: changed active_crown_fire to a scalar, and fixed the code comment noted by @pollybuotte

jkshuman:
@rgknox the active_crown_fire parameter will be an "on/off" switch to use the active crown fire portion of the code. It should be switched to a binary logical flag. 1=active crown fire 0=no active crown fire

rgknox:
@jkshuman , how do those changes look?

jkshuman:
@rgknox looks good. Thanks.

rgknox:
working with @JessicaNeedham to incorporate some of her new parameters as well, as some near-term integrations from her that use those parameters are forthcoming

parameter_files/fates_params_default.cdl
```diff
@@ -264,6 +301,18 @@ variables:
 	double fates_mort_bmort(fates_pft) ;
 		fates_mort_bmort:units = "1/yr" ;
 		fates_mort_bmort:long_name = "background mortality rate" ;
+	double fates_mort_ip_senescence(fates_pft);
```
ckoven: @rgknox @JessicaNeedham for clarity/readability sake, can we add "_size_" to these two parameter names to better distinguish from the age ones?

parameter_files/fates_params_default.cdl
```diff
@@ -1098,12 +1220,16 @@ data:
 
  fates_cohort_fusion_tol = 0.08 ;
 
+ fates_cohort_age_fusion_tol = - ;
```
ckoven: @rgknox @JessicaNeedham  can this parameter also serve as a switch to turn off cohort age tracking entirely or do we need to add a separate variable to do that?

parameter_files/fates_params_default.cdl
```diff
@@ -1098,12 +1220,16 @@ data:
 
  fates_cohort_fusion_tol = 0.08 ;
```
ckoven: do we need to change the name of this to fates_cohort_size_fusion_tol to distinguish from the new age variable?

parameter_files/fates_params_default.cdl
JessicaNeedham: Yes, that's a good idea.

parameter_files/fates_params_default.cdl
rgknox: yeah, @ckoven , I like that
"fates_cohort_size_fusion_tol"

parameter_files/fates_params_default.cdl
JessicaNeedham: Yes this parameter could work as the switch. It doesn't have that function yet but I can add it in.

rgknox:
updated parameter names and metadata per discussion @ckoven and @JessicaNeedham

jkshuman:
@rgknox I am going to pull another parameter out of the fire code. CG_strikes is a scaler on lightning datasets to identify the percentage of cloud to ground strikes. Will be easier to test and modify this if it lives in the param file as a global scaler parameter. Will take care of it this week.
https://github.com/jkshuman/fates/blob/e5e808ca3f7c4f65e4a5fd3d6dbee61a639ac9ab/fire/SFMainMod.F90#L683-L684

rgknox:
All expected tests pass, b4b.  namelist tests do not pass, as expected.

/glade/scratch/rgknox/clmed-tests/fates-sci-1.31.1_api.8.1.0-Ceac01319-F26ab2875.fates.cheyenne.intel

@jkshuman, can you take a look at the new cg_strikes parameter and sign off?

parameter_files/fates_params_default.cdl
```diff
@@ -273,6 +310,18 @@ variables:
 	double fates_mort_hf_sm_threshold(fates_pft) ;
 		fates_mort_hf_sm_threshold:units = "unitless" ;
 		fates_mort_hf_sm_threshold:long_name = "soil moisture (btran units) at which drought mortality begins for non-hydraulic model" ;
+	double fates_mort_ip_age_senescence(fates_pft) ;
+		fates_mort_ip_age_senescence:units = "years" ;
+		fates_mort_ip_age_senescence:long_name = "Mortality cohort age senescence inflection point" ;
+	double fates_mort_ip_senescence(fates_pft) ;
```
ckoven:
```suggestion
	double fates_mort_ip_size_senescence(fates_pft) ;
```

parameter_files/fates_params_default.cdl
```diff
@@ -273,6 +310,18 @@ variables:
...
+		fates_mort_ip_senescence:units = "dbh cm" ;
+		fates_mort_ip_senescence:long_name = "Mortality dbh senescence inflection point" ;
+	double fates_mort_r_age_senescence(fates_pft) ;
+		fates_mort_r_age_senescence:units = "mortality rate year^-1" ;
+		fates_mort_r_age_senescence:long_name = "Mortality age senescence rate of change" ;
+	double fates_mort_r_senescence(fates_pft) ;
```
ckoven:
```suggestion
	double fates_mort_r_size_senescence(fates_pft) ;
```

ckoven:
I think my earlier comments to add _size_ to a couple parameters got lost during the alphabetization step

parameter_files/fates_params_default.cdl
```diff
@@ -890,6 +993,14 @@ data:
  fates_mort_hf_sm_threshold = 1e-06, 1e-06, 1e-06, 1e-06, 1e-06, 1e-06, 
     1e-06, 1e-06, 1e-06, 1e-06, 1e-06, 1e-06 ;
 
+ fates_mort_ip_age_senescence = _, _, _, _, _, _, _, _, _, _, _, _ ;
+
+ fates_mort_ip_senescence = _, _, _, _, _, _, _, _, _, _, _, _ ;
```
ckoven:
```suggestion
 fates_mort_ip_size_senescence = _, _, _, _, _, _, _, _, _, _, _, _ ;
```

parameter_files/fates_params_default.cdl
```diff
@@ -890,6 +993,14 @@ data:
...
+
+ fates_mort_r_age_senescence = _, _, _, _, _, _, _, _, _, _, _, _ ;
+
+ fates_mort_r_senescence = _, _, _, _, _, _, _, _, _, _, _, _ ;
```
ckoven:
```suggestion
 fates_mort_r_size_senescence = _, _, _, _, _, _, _, _, _, _, _, _ ;
```

parameter_files/fates_params_default.cdl
```diff
@@ -273,6 +310,18 @@ variables:
 	double fates_mort_hf_sm_threshold(fates_pft) ;
 		fates_mort_hf_sm_threshold:units = "unitless" ;
 		fates_mort_hf_sm_threshold:long_name = "soil moisture (btran units) at which drought mortality begins for non-hydraulic model" ;
+	double fates_mort_ip_age_senescence(fates_pft) ;
+		fates_mort_ip_age_senescence:units = "years" ;
+		fates_mort_ip_age_senescence:long_name = "Mortality cohort age senescence inflection point" ;
+	double fates_mort_ip_senescence(fates_pft) ;
+		fates_mort_ip_senescence:units = "dbh cm" ;
```
ckoven:
```suggestion
		fates_mort_ip_size_senescence:units = "dbh cm" ;
```

parameter_files/fates_params_default.cdl
```diff
@@ -273,6 +310,18 @@ variables:
...
+		fates_mort_ip_senescence:long_name = "Mortality dbh senescence inflection point" ;
```
ckoven:
```suggestion
		fates_mort_ip_size_senescence:long_name = "Mortality dbh senescence inflection point" ;
```

parameter_files/fates_params_default.cdl
```diff
@@ -273,6 +310,18 @@ variables:
...
+	double fates_mort_r_age_senescence(fates_pft) ;
+		fates_mort_r_age_senescence:units = "mortality rate year^-1" ;
+		fates_mort_r_age_senescence:long_name = "Mortality age senescence rate of change" ;
+	double fates_mort_r_senescence(fates_pft) ;
+		fates_mort_r_senescence:units = "mortality rate dbh^-1" ;
```
ckoven:
```suggestion
		fates_mort_r_size_senescence:units = "mortality rate dbh^-1" ;
```

parameter_files/fates_params_default.cdl
```diff
@@ -273,6 +310,18 @@ variables:
...
+		fates_mort_r_senescence:long_name = "Mortality dbh senescence rate of change" ;
```
ckoven:
```suggestion
		fates_mort_r_size_senescence:long_name = "Mortality dbh senescence rate of change" ;
```

rgknox:
see my first question in the discussion of PR 590, here: https://github.com/NGEET/fates/pull/590
It may be worth adding the namelist changes to the current PR

rgknox:
based on feedback so far on #590, I'm not going to add any changes related to hydraulics to this PR. 
It looks like we will most likely make changes to the namelist variables, and how we use hlm_use_planthydro.  I've mocked up some changes here:  https://github.com/rgknox/fates/compare/parameters-eca-phen-fire...rgknox:parameters-nlist-eca-phen-fire

I will go ahead and re-test this PR with @ckoven 's updates, this should be read to merge soon.

rgknox:
all expected pass, with NLFAILs (expected fail).
/glade/scratch/rgknox/clmed-tests/fates-sci-1.31.1_api.8.1.0-C195c9b2e-F25ff0f6e.fates.cheyenne.intel

jenniferholm:
Ahh, I just missed this to add in a new parameter. My bad team. There is a fates_leaf level parameter needed for taking into account nutrients effects during photosynthesis. It's "flnr" for fraction of leaf N in Rubisco (gN Rubisco / gN leaf). 
But if it's ok, I can just add this as another pull request (when I get closer to finishing the photosynthesis component), right?

rgknox:
>But if it's ok, I can just add this as another pull request (when I get closer to finishing the photosynthesis component), right?
sure, we are bound to have another round of parameter updates in the near future, we just just try to clump them whenever possible

PR #599, opened 2020-01-09
Size- and age- dependent mortality 

JessicaNeedham:
### Description:
These changes introduce two new mortality terms, M9 and M10 which correspond to size and age dependent mortality respectively. Both gradually increase the rate of mortality with either size or age. They are each controlled by two parameters, an inflection point which determines the size or age at which mortality rate reaches 0.5, and a rate parameter which determines how steep the curve is, i.e. how rapidly mortality increases with size or age. The age dependent mortality function requires cohort age to be tracked. This means that cohorts cannot fuse unless both size and age are within a given tolerance level. The addition of this cohort age dimension requires changes to E3SM and CTSM. All parameters associated with these changes are _ by default. If a user wants to use this functionality they must set these parameters to be sensible values. Cohort age remains 0 otherwise, as do both mortality terms.  

### Collaborators:
@ckoven 
@rgknox 
@glemieux 
@rosiealice 

If parameter values remain _ then there should be no change from master. 
If they are switched on then mortality rates will increase. 

### Checklist:

- [X ] My change requires a change to the documentation.

### Test Results:

FATES baseline hash-tag:

Test Output:


rosiealice:
Looks good (the implementation that is). Philisophically, both of these things (discretizing by age and introducing size and age dependent mortalities) will probably have impacts that we'll need to understand.   I'm sure you've thought this through already, but fwiw, I'd be interested in 
- The extent to which we end up tracking lots of under-storey cohorts of different ages that previously would have been the same size.
- Whether the discretization parameters for size and age have an impact on the overall outcome.
- If it's possible to easily turn off age discretization if that ends up being a second-order effect?
Anyway, onwards and upwards!

JessicaNeedham:
Oh, good points. 

1. We definitely end up with more cohorts when we track age - figure to follow soon - but I think it is around 5x more. This does slow down model runs so I’ve made it that age tracking is off by default. 
2. I’m not entirely sure I understand your question here. I haven’t run FATES with cohort age tracking independent of age-dependent mortality so I’m not sure whether cohort fusion based on size and age alone will change results. I think it would be straightforward to run that test. 
3. The way I have it now (might change once there is more code review) age tracking is off by default. Cohorts are initialised with age 0, but if cohort age tracking is off this age is never increased and has no impact on cohort outcomes. Currently cohort age will only increase if the cohort age fusion tolerance parameter is set.

JessicaNeedham:
CTSM_test_AGB_BA_CRES
CTSM_test_mortality
CTSM_test_nplants_ncohorts

These figures show the functional form of the size and age-dependent mortality terms. I parameterised them somewhat arbitrarily here but you can see the reduction in BA, AGB and C residence time that results. If you wanted a bigger decrease in the number of large stems (@ckoven ) you can decrease the inflection point of the mortality terms. With both size- and age-mortality turned off (No mort in the legend) results match master branch. Last figure shows the large increase in cohort number when cohort age tracking is on.

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -206,14 +248,36 @@ subroutine Mortality_Derivative( currentSite, currentCohort, bc_in)
...
        dndt_logging = (currentCohort%lmort_direct     + &
-                       currentCohort%lmort_collateral + &
-                       currentCohort%lmort_infra)/hlm_freq_day
-       currentCohort%dndt = -1.0_r8 * (cmort+hmort+bmort+frmort+dndt_logging) * currentCohort%n
+            currentCohort%lmort_collateral + &
+            currentCohort%lmort_infra)/hlm_freq_day
+
+       ! this  caps bmort so that daily mortality cannot exceed 0.995
+       if ((cmort+hmort+bmort+frmort+smort+asmort + dndt_logging)*hlm_freq_day &
+            > 0.995_r8)then
```
ckoven: where is the 0.995 coming from?  maybe make a named constant?

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -75,12 +82,41 @@ subroutine mortality_rates( cohort_in,bc_in,cmort,hmort,bmort,frmort )
...
                                                         ! may help to debug carbon imbalances
                                                         ! and the like
+     
+   ! Size Dependent Senescence
+   ! rate (r) and inflection point (ip) define the increase in mortality rate with dbh
+    mort_r_size_senescence = EDPftvarcon_inst%mort_r_size_senescence(cohort_in%pft)
+    mort_ip_size_senescence = EDPftvarcon_inst%mort_ip_size_senescence(cohort_in%pft)
+
+       ! if param values have been set then calculate smort
+    if (mort_ip_size_senescence < 10000 ) then
```
ckoven: change 10000 to 10000.0_r8?  or make some named constant?

JessicaNeedham: does FATES have a named constant for an arbitrary large number?

main/EDInitMod.F90
```diff
@@ -201,6 +202,7 @@ subroutine zero_site( site_in )
 
     ! fusoin-induced growth flux of individuals
     site_in%growthflux_fusion(:,:) = 0._r8
+!    site_in%ageflux_fusion(:,:) = 0._r8
```
ckoven: delete?

main/EDTypesMod.F90
```diff
@@ -14,15 +14,16 @@ module EDTypesMod
...
+  
   implicit none
   private               ! By default everything is private
   save
 
   integer, parameter, public :: maxPatchesPerSite  = 14   ! maximum number of patches to live on a site
   integer, parameter, public :: maxPatchesPerSite_by_disttype(n_anthro_disturbance_categories)  = &
                                                      (/ 10, 4 /)  !!! MUST SUM TO maxPatchesPerSite !!!
-  integer, parameter, public :: maxCohortsPerPatch = 100  ! maximum number of cohorts per patch
+  integer, parameter, public :: maxCohortsPerPatch = 300  ! maximum number of cohorts per patch
```
ckoven: can we revert this to original number if no age tracking is happening?

biogeochem/EDMortalityFunctionsMod.F90
ckoven: there's this, but I'm not sure if it is the right really large number.
https://github.com/NGEET/fates/blob/master/main/FatesConstantsMod.F90#L136

main/EDTypesMod.F90
```diff
@@ -747,9 +754,9 @@ module EDTypesMod
      real(r8), allocatable :: fmort_rate_crown(:,:)              ! rate of individuals killed due to fire mortality 
                                                                  ! from crown damage per year.  on size x pft array
 
-     real(r8), allocatable :: growthflux_fusion(:,:)             ! rate of individuals moving into a given size class bin 
-                                                                 ! due to fusion in a given day. on size x pft array 
-
+     real(r8), allocatable :: growthflux_fusion(:,:)             ! rate of individuals moving into a given size class bin
+     ! due to fusion in a given day. on size x pft array 
+!     real(r8), allocatable :: ageflux_fusion(:,:)                ! rate of individuals movign into a given age class bin due to fusion- age x pft
```
ckoven: delete?

main/FatesHistoryInterfaceMod.F90
```diff
@@ -2573,10 +2710,22 @@ subroutine update_history_dyn(this,nc,nsites,sites)
                ! while in this loop, pass the fusion-induced growth rate flux to history
                hio_growthflux_fusion_si_scpf(io_si,i_scpf) = hio_growthflux_fusion_si_scpf(io_si,i_scpf) + &
                     sites(s)%growthflux_fusion(i_scls, i_pft) * days_per_year
-
             end do
          end do
          !
+
```
ckoven: delete the commented-out block below?

main/FatesRestartInterfaceMod.F90
```diff
@@ -1822,8 +1851,16 @@ subroutine set_restart_vectors(this,nc,nsites,sites)
                 
              io_idx_si_sc = io_idx_si_sc + 1
           end do
-         
 
+
```
ckoven: delete this block below too?

main/FatesRestartInterfaceMod.F90
```diff
@@ -2574,8 +2622,16 @@ subroutine get_restart_vectors(this, nc, nsites, sites)
                 
              io_idx_si_sc = io_idx_si_sc + 1
           end do
-
```
ckoven: and delete this one too?

main/EDTypesMod.F90
JessicaNeedham: Yes, but somewhere it will need to be increased if cohort age tracking is on.

biogeochem/EDMortalityFunctionsMod.F90
JessicaNeedham: I'll change it to 10000.0-r8

main/EDTypesMod.F90
ckoven: ok, seems like there is the potential for circular dependencies here.  maybe the simplest thing to do is, rather than using the slightly cryptic `if (ED_val_cohort_age_fusion_tol < 10000.0_r8)` condition a few times in the code as it is now, it would make more sense to have some logical variable called like `use_cohort_age_tracking`, which you set as true or false during initialization (possibly using the same assessment of ED_val_cohort_age_fusion_tol), and then refer to that here and elsewhere.  but where is the best place to put such a flag to avoid circular dependencies?  @rgknox do you know?

main/EDInitMod.F90
JessicaNeedham: Oh, I forgot about that. I'll delete all that ageflux code since it is not meaningful.

main/EDTypesMod.F90
JessicaNeedham: Deleted.

main/FatesHistoryInterfaceMod.F90
JessicaNeedham: Deleted.

tools/FatesPFTIndexSwapper.py
```diff
@@ -10,12 +10,13 @@
...
 
 import numpy as np
+from numpy import *
 import sys
 import getopt
 import code  # For development: code.interact(local=locals())
 from datetime import datetime
 from scipy.io import netcdf
-import matplotlib.pyplot as plt
+#import matplotlib.pyplot as plt
```
ckoven: I just ran into this same issue -- that tools/FatesPFTIndexSwapper.py imports pyplot but doesn't actually use it anywhere, so its an unneeded dependency that complicates things.  suggest delete this line entirely.

tools/FatesPFTIndexSwapper.py
```diff
@@ -144,15 +145,15 @@ def main(argv):
 
     fp_in  = netcdf.netcdf_file(input_fname, 'r')
 
-    for key, value in sorted(fp_in.dimensions.iteritems()):
+    for key, value in sorted(fp_in.dimensions.items()):
```
ckoven: is this a python 2 vs python 3 thing?

ckoven: I should say that my python brain is stuck firmly in the aughts as regards to python versioning.  if we are making the switch to python 3, that's great and long overdue, but just let me know that that is a line that we've crossed.

JessicaNeedham: Sorry, yes this should have been a separate issue. I think python 3 only uses items now.

ckoven: great, let's make the change. thanks.

biogeochem/EDCohortDynamicsMod.F90
```diff
@@ -994,7 +1020,20 @@ subroutine fuse_cohorts(currentSite, currentPatch, bc_in)
...
                  !Criteria used to divide up the height continuum into different cohorts.
 
-                 if (diff < dynamic_fusion_tolerance) then
+                 if (diff < dynamic_size_fusion_tolerance) then
+
+                    ! Only fuse if the cohorts are within x years of each other 
+                    ! if they are the same age we make diff 0- to avoid errors divding by zero
+                    !NB if cohort age tracking is off then the age of both should be 0
+                    ! and hence the age fusion criterion is met    
+                   if (currentCohort%coage .eq. nextc%coage) then
```
rgknox: might recommend the following logic here:

if( abs(currentCohort%coage-nextc%coage)<nearzero) then

main/EDTypesMod.F90
rgknox: For the time being, perhaps EDParamsMod.F90?

rgknox: Having a spot for these types of flags is part of the wishlist for refactoring FatesInterfaceMod.F90.  That refactor will create a new module that has just datastructures, including the various control flags, but would not have many of the setup procedures, so it would have little dependency issues.

main/FatesInventoryInitMod.F90
```diff
@@ -1112,8 +1112,10 @@ subroutine set_inventory_edcohort_type1(csite,bc_in,css_file_unit,npatches, &
 
          call prt_obj%CheckInitialConditions()
 
+
          ! Since spread is a canopy level calculation, we need to provide an initial guess here.
-         call create_cohort(csite, cpatch, temp_cohort%pft, temp_cohort%n, temp_cohort%hite, temp_cohort%dbh, &
+call create_cohort(csite, cpatch, temp_cohort%pft, temp_cohort%n, temp_cohort%hite, &
```
rgknox: sorry, being very picky, could you fix indentation here

tools/FatesPFTIndexSwapper.py
rgknox: yup, agreed

main/FatesInventoryInitMod.F90
JessicaNeedham: Yikes, sorry. I'll fix that.

tools/FatesPFTIndexSwapper.py
rgknox: thanks, yeah if we make this py2/3 compliant that is win win

biogeochem/EDCohortDynamicsMod.F90
JessicaNeedham: Sounds good, I've changed it.

biogeochem/EDMortalityFunctionsMod.F90
JessicaNeedham: I'm actually wondering now why I made that 0.995. Should it just be 1?

ckoven: yeah, I don't see why it shouldn't be.  maybe test and if it fails we can try to fix whatever underlying problem might be there?

JessicaNeedham: I don't think this cap actually makes sense. I was trying to avoid doubling up with background mortality and size/age mortality. But the way it is written I don't think daily mortality could ever exceed 1 if multiplied by hlm_freq_day. 
Maybe what I need is just bmort+smort or asmort cannot >1. i.e. if the annual rates of just those two terms are exceeding 1 we need to start reducing bmort.

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -75,12 +81,41 @@ subroutine mortality_rates( cohort_in,bc_in,cmort,hmort,bmort,frmort )
...
+    if (mort_ip_size_senescence < 10000.0_r8 ) then 
+       smort = 1.0_r8 / ( 1.0_r8 + exp( -1.0_r8 * mort_r_size_senescence * &
+            (cohort_in%dbh - mort_ip_size_senescence) ) ) 
+    else
+       smort = 0.0_r8
+    end if
+
+    ! if user has set cohort age fusion param to not be _ we calculate age
+    ! dependent mortality
+    if (mort_ip_age_senescence < 10000.0_r8) then
```
rgknox: Is it true that we only want to calculate an aging mortality rate if we are tracking aging and cohort_age_tracking is true?  Should we be using that logical here?

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -75,12 +81,41 @@ subroutine mortality_rates( cohort_in,bc_in,cmort,hmort,bmort,frmort )
     logical, parameter :: test_zero_mortality = .false. ! Developer test which
                                                         ! may help to debug carbon imbalances
                                                         ! and the like
+     
+   ! Size Dependent Senescence
+    ! rate (r) and inflection point (ip) define the increase in mortality rate with dbh
+    mort_r_size_senescence = EDPftvarcon_inst%mort_r_size_senescence(cohort_in%pft)
+    mort_ip_size_senescence = EDPftvarcon_inst%mort_ip_size_senescence(cohort_in%pft)
+
```
rgknox: I'm worried that a user will set mort_r_size_senescence to 0 and think that will turn off size dependent senescence.  I think that value generates an smort of 0.5.  Will check how this is read in.

rgknox: Lets add some metadata to the parameter value entry that explains _ turns this off, and what valid ranges are.
https://github.com/NGEET/fates/blob/master/parameter_files/fates_params_default.cdl#L322

There is also this subroutine here, a good place to check if fates parameters are in acceptable ranges (if for instance 0 is an unexpected/invalid, you can have fates gracefully complain and fail):
https://github.com/NGEET/fates/blob/d8e7935542222ca2333d7007c5b394a461538331/main/EDPftvarcon.F90#L2071

biogeochem/EDMortalityFunctionsMod.F90
JessicaNeedham: Yes, ageing mortality can only be calculated if cohort age tracking is on. I changed the logic so that cohort age tracking could be on without having ageing mortality. So now the 'cohort_age_tracking' parameter in the param file turns on cohort age tracking. And both size and age dependent mortality get turned on by setting the ip_age/size_senescence parameters. I can put in some flags so that the ip_age_senescence parameter can't be turned on without cohort_age_tracking being on.

biogeochem/EDMortalityFunctionsMod.F90
JessicaNeedham: Sounds good. I'll add some metadata to the parameter value entry and some checks in FatesCheckParams.

main/EDTypesMod.F90
JessicaNeedham: I changed maxCohortsPerPatch back to 100 and made a local variable maxCohortsPerPatch_age_tracking in EDCohortDynamicsMod.F90 that is 300 and that overwrites maxCohortsPerPatch if age tracking is on. Not sure if this is ok or if it should be in EDTypesMod.F90 also? I got rid of the cryptic ED_val_cohort_age_fusion_tol switch and made a cohort_age_tracking switch.

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -42,28 +42,35 @@ module EDMortalityFunctionsMod
 
 
 
-  subroutine mortality_rates( cohort_in,bc_in,cmort,hmort,bmort,frmort )
+  subroutine mortality_rates( cohort_in,bc_in,cmort,hmort,bmort,frmort,smort,asmort )
 
     ! ============================================================================
     !  Calculate mortality rates from carbon storage, hydraulic cavitation, 
-    !  background and freezing
+    !  background and freezing and size dependent senescence
```
rosiealice: and size _and age_ dependent senescence maybe? 

Also, is it me, or is 'senescence' really, really, hard to spell?

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -75,12 +82,41 @@ subroutine mortality_rates( cohort_in,bc_in,cmort,hmort,bmort,frmort )
...
+    
+    ! if param values have been set then calculate smort
+    if ( mort_ip_size_senescence < check_initialized ) then 
+       smort = 1.0_r8 / ( 1.0_r8 + exp( -1.0_r8 * mort_r_size_senescence * &
+            (cohort_in%dbh - mort_ip_size_senescence) ) ) 
+    else
+       smort = 0.0_r8
+    end if
+
+    ! if user has set cohort age fusion param to not be _ we calculate age
```
rosiealice: What is _ in this sentence? Also why doesn't this use 'cohort_age_tracking'?

main/EDPftvarcon.F90
```diff
@@ -167,8 +171,8 @@ real(r8), allocatable :: allom_smode(:)        ! sapwood allometry function type
      ! Prescribed Physiology Mode Parameters
      real(r8), allocatable :: prescribed_npp_canopy(:)           ! this is only for the special 
                                                                  ! prescribed_physiology_mode
-     real(r8), allocatable :: prescribed_npp_understory(:)       ! this is only for the special
```
rosiealice: Do we need 'special' in all of these notes? :)

main/EDPftvarcon.F90
```diff
@@ -2114,12 +2155,83 @@ subroutine FatesCheckParams(is_master, parteh_mode)
...
+             ( cohort_age_tracking .neqv. .TRUE. ) ) then
+
+           write(fates_log(),*) 'ip_age = ', EDPftvarcon_inst%mort_ip_age_senescence(ipft)
+           write(fates_log(),*) 'cohort_age_tracking = ', cohort_age_tracking
+
+           write(fates_log(),*) 'Age-dependent mortality cannot be on'
+           write(fates_log(),*) 'if cohort age tracking is off'
+           write(fates_log(),*) 'Set cohort_age_tracking to 1'
+           write(fates_log(),*) 'To turn on cohort age tracking or '
+           write(fates_log(),*) 'Set mort_ip_age_senescence to _ '
```
rosiealice: Do you really mean _ or is a placeholder for something else?

main/EDPftvarcon.F90
```diff
@@ -2114,12 +2155,83 @@ subroutine FatesCheckParams(is_master, parteh_mode)
...
+
+
+        ! Check that parameter ranges for age-dependent mortality make sense   
+        !-----------------------------------------------------------------------------------    
+        if ( ( EDPftvarcon_inst%mort_ip_age_senescence(ipft) < check_initialized ) .and. &
+             (  EDPftvarcon_inst%mort_r_age_senescence(ipft) > check_initialized ) ) then
+
+           write(fates_log(),*) 'Age-dependent mortality is on'
+           write(fates_log(),*) 'Please also set mort_r_age_senescence'
+           write(fates_log(),*) 'Sensible values are between 0.03-0.06'
```
rosiealice: I like the 'sensible values' guidance!!

main/EDPftvarcon.F90
```diff
@@ -2114,12 +2155,83 @@ subroutine FatesCheckParams(is_master, parteh_mode)
...
+           write(fates_log(),*) 'is negative which makes no biological sense.'
+           write(fates_log(),*) 'Sensible values for ip are between 10 and 2000?'
+           write(fates_log(),*) 'Sensible values for r are between 0.03-0.06'
+           write(fates_log(),*) 'Aborting'
+           call endrun(msg=errMsg(sourcefile, __LINE__))
+        end if
+
+        ! Check that parameter ranges for size-dependent mortality make sense   
+        !-----------------------------------------------------------------------------------    
+        if ( ( EDPftvarcon_inst%mort_ip_size_senescence(ipft) < check_initialized ) .and. &
```
rosiealice: are these > and < supposed to be different ways round?

parameter_files/fates_params_default.cdl
```diff
@@ -529,9 +529,12 @@ variables:
 	double fates_canopy_closure_thresh ;
 		fates_canopy_closure_thresh:units = "unitless" ;
 		fates_canopy_closure_thresh:long_name = "tree canopy coverage at which crown area allometry changes from savanna to forest value" ;
+	double fates_cohort_age_tracking ;
+		fates_cohort_age_tracking:units = "0 or 1" ;
```
rosiealice: I kind-of feel like this would work better, for the sake of constancy, as a 'use' flag with .true. and .false. (e.g. as for SPITFIRE and HYDRO), but I appreciate that that's probably a bigger question (as per the discussion elsewhere in the review).

biogeochem/EDCohortDynamicsMod.F90
```diff
@@ -128,10 +131,10 @@ module EDCohortDynamicsMod
 
   !-------------------------------------------------------------------------------------!
 
-  subroutine create_cohort(currentSite, patchptr, pft, nn, hite, dbh,   &
+    
+  subroutine create_cohort(currentSite, patchptr, pft, nn, hite, coage, dbh,   &
```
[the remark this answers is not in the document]
JessicaNeedham: Yes, good point. I was trying to avoid confusion with patch age which I think is just 'age' in the code. I could make it cohort_age?

biogeochem/EDCohortDynamicsMod.F90
```diff
@@ -472,8 +485,12 @@ subroutine nan_cohort(cc_p)
     currentCohort%size_class         = fates_unset_int  ! size class index
     currentCohort%size_class_lasttimestep = fates_unset_int  ! size class index
     currentCohort%size_by_pft_class  = fates_unset_int  ! size by pft classification index
+    currentCohort%coage_class        = fates_unset_int  ! cohort age class index
+    currentCohort%coage_by_pft_class = fates_unset_int  ! cohort age by pft class index 
+
     currentCohort%n                  = nan ! number of individuals in cohort per 'area' (10000m2 default)     
-    currentCohort%dbh                = nan ! 'diameter at breast height' in cm                            
+    currentCohort%dbh                = nan ! 'diameter at breast height' in cm
+    currentCohort%coage              = nan ! age of the cohort in years
```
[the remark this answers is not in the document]
JessicaNeedham: Oops. It is definitely not supposed to change units. That should be days.

biogeochem/EDCohortDynamicsMod.F90
```diff
@@ -943,24 +967,37 @@ subroutine fuse_cohorts(currentSite, currentPatch, bc_in)
...
 
      logical, parameter :: fuse_debug = .false.   ! This debug is over-verbose
                                                  ! and gets its own flag
 
      !----------------------------------------------------------------------
 
      !set initial fusion tolerance
-     dynamic_fusion_tolerance = ED_val_cohort_fusion_tol
+     dynamic_size_fusion_tolerance = ED_val_cohort_size_fusion_tol
+     ! set the cohort age fusion tolerance (this is in days - remains constant)
```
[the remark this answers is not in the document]
JessicaNeedham: Oops, it was originally constant but now it is dynamic so I've updated that comment. And I've added 'in cm' to the line above.

biogeochem/EDCohortDynamicsMod.F90
```diff
@@ -994,7 +1032,20 @@ subroutine fuse_cohorts(currentSite, currentPatch, bc_in)
...
+
+                    ! Only fuse if the cohorts are within x years of each other 
+                    ! if they are the same age we make diff 0- to avoid errors divding by zero
+                    !NB if cohort age tracking is off then the age of both should be 0
+                    ! and hence the age fusion criterion is met    
+                    if (abs(currentCohort%coage - nextc%coage)<nearzero ) then
+                       coage_diff = 0.0_r8
+                    else
+                       coage_diff = abs((currentCohort%coage - nextc%coage)/ &
+                            (0.5*(currentCohort%coage + nextc%coage)))
```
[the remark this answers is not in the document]
JessicaNeedham: Fixed.

biogeochem/EDMortalityFunctionsMod.F90
JessicaNeedham: Added 'age' in there. And yes! It is a terrible word to have to type repeatedly.

main/EDPftvarcon.F90
JessicaNeedham: I'll defer to @ckoven on this one. I don't think I meant to change any of the comments here.

main/EDPftvarcon.F90
JessicaNeedham: I did mean _. These values are _ by default, and if the user wants to turn size or age dependent mortality on then they need to set these values.

main/EDPftvarcon.F90
JessicaNeedham: That was @rgknox 's idea :) . They rate parameter is not very intuitive so hopefully this stops people wasting time experimenting with values that give unrealistic results.

main/EDPftvarcon.F90
ckoven: i guess all the modes are special in their own way, happy to let the word get deleted here.

biogeochem/EDCohortDynamicsMod.F90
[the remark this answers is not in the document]
ckoven: what units is patch age tracked in?  Seems like we ought to use the same units for both quantities to avoid confusion.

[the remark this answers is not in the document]
JessicaNeedham: Oh wait... they are both in years.. urgh, changing it back.

biogeochem/EDMortalityFunctionsMod.F90
JessicaNeedham: I've changed it to match the comment above for size mortality. The size and age dependent mortality parameters are _ by default in the parameter file. The user has to set them in order to turn on either mortality term. Cohort age tracking is turned on separately with the cohort_age_tracking parameter. This was in case anyone in the future wanted to track cohort age independently of having size or age mortality turned on.

main/EDPftvarcon.F90
JessicaNeedham: @rgknox did the value of check_initialized change recently? I thought this was a very large number but now it seems to be negative? These checks are no longer working. A _ in the parameter file is 9.969E+036

biogeochem/EDMortalityFunctionsMod.F90
rgknox: check_initialized is defined in PRTGenericMod.F90, but that check is designed to work in step with un_initialized.  See here:

https://github.com/NGEET/fates/blob/master/parteh/PRTGenericMod.F90#L43

It only works in situations where we are defining uninitialized values with "un_initialized".  In this case, the parameter values that have "_"  from the file, are not the same as un_initialized.  

I remember we thought it was 9.9E36, but there was some confusion if we were importing that value correctly...  I had thought that "spval" from CLM/ELM would give us this value.

We have an "ignore" value for history, but that value is not necessarily the "no data" value we get from netcdf:  https://github.com/NGEET/fates/blob/master/main/FatesInterfaceMod.F90#L94

JessicaNeedham: Ah ok, thank you.  I misunderstood the un_initialized value. Would it make most sense then for me to define a local arbitrary large number to compare these parameters against, in order to check whether or not they have been set by the user in the parameter file?

rgknox: I think the fill values we were printing out when they made there way to FATES were 1E36 right? If so, I think this is is where it is defined...

https://github.com/ESCOMP/CTSM/blob/master/src/main/ncdio_pio.F90.in#L146

JessicaNeedham: If I print a parameter that is "_" in the parameter file I get 9.9692099...E+036.

JessicaNeedham: I think it is defined in ./cime/src/externals/pio2/src/flib/pio_types.F90 L234-235

JessicaNeedham: I've made a new variable called 'fates_check_param_set' defined in FatesConstantsMod to check whether the mortality parameters have been set. This might not be the way to go so happy to change it if there is a better way of doing things. But it seems to be working.

parameter_files/fates_params_default.cdl
JessicaNeedham: I agree. The rule sets for when size or age mortality or cohort age tracking were on or off was confusing. I've removed the cohort_age_tracking parameter (hopefully before it goes in the next set of parameter file updates) and have made cohort age tracking a flag like spitfire or hydro.

rosiealice: cool...

biogeochem/EDMortalityFunctionsMod.F90
rosiealice: Ha ha. Glad it isn't just my brain senescing, or whatever.

biogeochem/EDCohortDynamicsMod.F90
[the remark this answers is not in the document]
rosiealice: I think it might be more bother than it's worth to change, given your point about avoiding confusion with patch age (and people do, indeed, get very confused about these things).

main/EDPftvarcon.F90
rgknox: weird, I could had sworn we had a back and forth about this yesterday, and now the discussion disappeared.  I did look at your code changes though, and they should work.  I see you created a new check_initialized in FatesConstants and compare to that

biogeochem/EDMortalityFunctionsMod.F90
JessicaNeedham: @rgknox I think the discussion from yesterday is here.

rgknox:
Tests are looking good, I'm expecting NLCOMP fails, and I don't expect the new test for size and age mortality to work until we update the parameter file:
/glade/scratch/rgknox/clmed-tests/fates-sci.1.34.0_api.9.0.0-clm5.0.30-C77e0b690-Fee90e0f8.fates.cheyenne.intel
/glade/scratch/rgknox/clmed-tests/fates-sci.1.34.0_api.9.0.0-clm5.0.30-C77e0b690-Fee90e0f8.fates.cheyenne.gnu

glemieux:
Two tests failed in the above suite.  @rgknox noted that it was likely due to the `maxCohortPerPatch` having been increased.  @JessicaNeedham commit changed back to the original value setting and all expected pass now:
```
/glade/u/home/glemieux/scratch/clmed-tests/pr599-SMS_Lm6-compare.fates.cheyenne.intel.C300c2f2d-F10809b93
/glade/u/home/glemieux/scratch/clmed-tests/pr599-SMS_Lm13.1x1-brazil-compare.fates.cheyenne.intel.C300c2f2d-F10809b93
```

PR #600, opened 2020-01-10
Enforcing line feed character for end of line

glemieux:
Adding a `.gitattributes` file to the root directory to ensure all users commit code with end of line (eol) as `lf`.

### Description:
This PR attempts to ensure that all developer's local git repos utilize the same configuration for "normalizing" end of line characters in the code.  The background on this is detailed in issue #601.  This adds `.gitattributes`to the root directory of the repo and changes the only remaining file that still contains `crlf` eol characters.  Reference information on `.gitattributes` can be found on [github](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings#per-repository-settings) and on the [official git webpage](https://git-scm.com/docs/gitattributes).   

The PR also includes a minor update to `.gitignore` adding code editor specific hidden file (in this case VScode) to the ignore list.

### Collaborators:
@rgknox, @xuchongang 

Does not affect any existing runtime code, so none expected.

### Checklist:

### Test Results:
TBD

FATES baseline hash-tag:

Test Output:


.gitattributes
```diff
@@ -0,0 +1,29 @@
...
+# and was checked out to local, then git will not change the eol character during check-in (i.e. commit).  For 
+# windows users they will see a warning like this:
+# warning: CRLF will be replaced by LF in functional_unit_testing/allometry/drive_allomtests.py.
+# The file will have its original line endings in your working directory
+
+## Explicitly declare to git which files should be normalized (i.e. treated as text files)
+*.cdl text
+*.F90 text
+*.F90_in text
+*.mod text
```
rgknox: @glemieux I thought this was binary, its also in the binary list below.  Can this be redundant?

glemieux: Ah your right, it is redundant.  Good catch!  I'll fix that up.

PR #602, opened 2020-01-14
Add some minor documentation

ckoven:
Update to add slightly better documentation to the parameter file.  Some of the parameters, particularly those that act as switches, need to be more self-explanatory in the parameter file so that we can avoid having to always dig through the code.  So I added a "possible_values" attribute to some of the variables, which describes what values are accepted and what they mean.  This  is only a starter for some of the more common confusion-sowing cases.  Let's try to use this more often, particularly as we add more switches to the model.

While going through the code, I also realized we hadn't updated our copyright statement in a while and that some of the in-code documentation could be a bit clearer as well, so made a couple other tiny changes.

### Collaborators:
discussed with @rgknox 

should be bit for bit

### Checklist:

### Test Results:

As-yet untested.  Should be trivial changes.

CTSM (or) E3SM (specify which) test hash-tag: 5ad8a8d2

CTSM (or) E3SM (specify which) baseline hash-tag: 097f4239

FATES baseline hash-tag: 17aaa879

Test Output: 
All expected PASS (excepting hydro).

```
/glade/u/home/glemieux/scratch/clmed-tests/pr602-ckoven-minor-docs.fates.cheyenne.intel.C5ad8a8d2-F65fd1f66
```


glemieux:
Passed B4B.  Will merge after #554 is merged.

PR #608, opened 2020-02-05
Adding Medlyn stomatal conductance model into CTSM-FATES

Qianyuxuan:
Adding the unified stomatal optimization codes into CTSM-FATES model
### Description:

I have adapted the unified stomatal optimization (abbreviated as USO, Medlyn et al. 2011) codes in CLM5 into FATES, including reading two parameters (medlynintercept and medlynslope) from the parameter file and adding a quadratic solution of stomatal conductance. I added the two parameters according to the values in CLM5 based on PFTs. A flag has been created to switch between the Ball-Berry stomatal conductance model and the USO model into FATES. I also removed some hard-coded numbers in existing codes about photosynthesis and stomatal conductance.

### Collaborators:

Discussions with @alistairrogers @serbinsh 

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


serbinsh:
@rgknox We still need to update the parameter file with this PR to make it consistent with the code changes. I just want to confirm that the appropriate way to do this is to modify the parameter file in the parameter_files directory with the edits and then add that to this pull request, correct?

rgknox:
Hi @serbinsh
Ideally we try to bundle changes to the parameter file into groups that are submitted together, having fewer total sets of changes.  This helps reduce the burden of folks adapting to changes frequently.  We just released a grouped set of parameter changes this January.  
I think there a few things we can do here.  Lets look around and see if anyone else has updates to the parameter file and see what we have. If nothing else is forthcoming, check in with the team and perhaps just move forward with a change.  Or for the short term perhaps we could consider using existing parameters as surrogates for the new ones (ie like bb_slope); until we have a critical mass of parameter updates.

rgknox:
Another question, I'm noticing that the Medlyn method, in both FATES and CLM, the conductance is decoupled from leaf water potential.  In other words, the Medlyn method does not use the brtan value from what I can tell (am I wrong here?).  I know that Medlyn is all about the VPD, but if conductance doesn't respond to LWP, then this seems like it will give the hydraulics module a tough time.  My gut is telling me that everyone knows this and its just how it is and I missed the boat, or that I'm missing a term somewhere....

rgknox:
nevermind, I forgot that btran scales down vcmax...

ckoven:
actually isn't it a valid point, but only as applied to the minimum conductance term?  in the b-b scheme, the bbopt gets multiplied by btran, but is that not the case for the medlynintercept term?

serbinsh:
@rgknox @ckoven I am not sure I am following this discussion. First, the underlying basic Medlyn or Ball Berry leaf stomatal models do not have any BTran or other g0/g1 scalars associated with them. These are bolt-ons that have been added to the model to account for soil water supply (or other stress factors), but I would argue should be considered separately from the basic underlying equations. That is, for clarity sake I think we should be using the core underlying stomatal models as they have been published and only scaling parameters outside of the quadratic before solving. That way we are using the core model and any other changes in FATES that could impact this wont accidentally use parameter scaling when it wasnt intended.  

So, as you state currently in FATES bbopt is multiplied by Btran because at some point this was added as a stress scalar.  But this is only one way or hypothesis on how this could happen, correct? Because we do this with BB does not mean would should do this with Medlyn, unless we have a reason for this scaling, correct?  Also we do not yet have evidence that we should be modifying Medlyn parameters or solved gs by LWP and as such I strongly advocate we do not bake this in yet.  Jin's paper, for example, showed that LWP was not a significant factor controlling gs.  This again is a hypothesis/research question that I agree we need to explore to identify how best we would incorporate things like Btran or LWP scalars, but which are not part of the underlying Medlyn model. As such if we add the core models we can then explore multipliers that modify the model parameters or results but I dont think we need to add that in now since how we would actually do that is an active area of research.  

I am not sure this is as clear as it could be via this thread and perhaps more easily discussed over the phone.  We could also look at how other models such as CABLE scale Medlyn in relation to soil/LWP (if at all, I actually dont remember now) and try out different approaches.  That said, our goal was to provide a first version of the basic model which we could test with, compare with data, etc and then explore if further edits were needed.

walkeranthonyp:
I would counter that if it's done for the BB intercept then for consistency
it should be done for the Medlyn intercept.

As you say Shawn, it is a bolt on, or you mighty say a separate and
additional hypothesis about how soil water affects gs. So it should be
applied equally to the models that calculate gs from solely leaf-scale
conditions. Whether you agree or not with the hypothesis of how soil water
affects leaf physiology is a different argument and part of a wider
discussion about where to apply btran type scalars (gs, g0, g1, vcmax, or
gs and vcmax, or nowhere).

imo this adds to the argument for why we should not have a separate
intercept parameter for medlyn. It is the same parameter, it does the same
thing and has v little affect on things apart from night-time and as Danica
rightly points out that is also a different story.

alistairrogers:
I'm with Anthony on the btran application. It's a separate hypothesis to link soil water content to leaf physiology and should be applied to both or neither in a consistent manner but as Shawn says should not be baked into the stomotal model formulation. It is my understanding that hydraulic approaches should not be using btran so this might be moot in FATES (hydro) anyway. And yes this is a bigger conversation.

We touched on it briefly in Rogers et al (2017) but also see DeKauwe et al (2013, 2015)

The question of whether you make the intercept the same for both models is tricky. When you fit slope and intercept from the same gas exchange dataset you obviously get different slopes but the intercepts are also different (admittedly small numbers anyway). That's why I would advocate for have slope and intercept as PFT specific parameters that are paired. Initially you could set the intercept to be the same for all PFTs and maybe the same with both models but keeping them as paired PFT specific parameters would future proof the model.

serbinsh:
Thanks @walkeranthonyp and @alistairrogers  Hopefully my comments made sense and didn't come across as anything other than trying to better understand the requests and rationale, but I think you both really expanded and clarified so thank you.  I agree we need to be consistent but as stated I was thinking that perhaps before we lock something in we should perhaps explore what approaches/hypotheses we should be using in FATES first, but of course that could also lead us to delay model updates.  As such, I am happy to go with the group on this and do what we think is best but also spend time exploring alternatives.  As long as we document what we have done well in the code and tech doc it should be easy for us to later address any of these assumptions and make changes without the risk of tuning other processes to something fixed in the model that we then later regret if we try to undo it.  That was my main concern.

walkeranthonyp:
@arogers I hear what you're saying about the intercepts being different. But conceptually it's the same parameter that has the same units, just with a different value. If we had a different parameter for every time a parameter took a different value we'd have an infinite number of parameters. 

... not to mention, as you pointed out earlier, you're not even estimating g0 with the medlyn model (nor did Lin 2015). A zero intercept is assumed. So conceptually and pragmatically what are we to do with g0? This goes back to an earlier discussion prompted by @Qianyuxuan's addition of the medlyn model. 

Should we abandon g0 from the gs calculation in response to environment and use it solely as a minimum value instead? It's unlikely to have a big effects in productive areas during the day time, but at night and in dry areas could be significant.  ... a bonus would be that it would simplify the photosynthesis code.

alistairrogers:
@walkeranthonyp point taken. My thought was that since estimates of slope and intercept are not independent PFT specific slopes should be paired to PFT specific intercepts. Obviously g0=0 simplifies things considerably and as you point out would have a very low impact (other than arid environments) on daytime gas exchange. If you do go with g0=0 how do we handle the issue of nighttime minimum conductance raised by Danica and described in her GMD paper?

rgknox:
I should clarify my earlier point.  When Brad and Chonggang's plant hydraulics module is turned on, we calculate a btran value that is based off of LWP.  It uses a LWP, P50 and a vulnerability parameter to generate a fraction of total conductance.  When that module is off, the btran value is a true btran, and based off of soil moisture conditions.  As it currently stands, this fraction of total conductivity value is only applied to vcmax, and bbopt.

arogers:
You guys have the wrong Aaron Rogers...

walkeranthonyp:
Sorry old chap, my bad - his email is arogers. Any thoughts on keeping g0 as one or multiple parameters?

walkeranthonyp:
> @walkeranthonyp point taken. My thought was that since estimates of slope and intercept are not independent PFT specific slopes should be paired to PFT specific intercepts. Obviously g0=0 simplifies things considerably and as you point out would have a very low impact (other than arid environments) on daytime gas exchange. If you do go with g0=0 how do we handle the issue of nighttime minimum conductance raised by Danica and described in her GMD paper?

@serbinsh - course, always up for the discussion. The btran question is definitely relevant, even to hydro imo given that there is evidence for signaling between roots and leaves. We should probably have a separate btran thread / discussion, especially as it sounds like transpiration in hydro is pretty low atm.

- hear you on the non-independence of slope and intercept @alistairrogers ,  it wants looking out for, but imo it's more efficiently dealt with in the parameter file through proper parameterisation than by adding more parameters to the code.
- re g0 = 0. I would recommend leaving as is for now and adding that discussion to the btran discussion. If we decide we want to handle g0 as a minimum rather than an intercept we would just need to remove g0 from the quadratic soln. for gs (i.e. derive new solns.). The code to apply g0 as a minimum gs already exists and is in operation.

rosiealice:
Just noting that as I recall, Polly was interested in looking at how variations in minimum.conductance affected PFT performance in her simulations, hence, I think I'd advocate for it being a PFT level parameter, at least (

rosiealice:
...and Ive no idea how I closed the pull request (writing on my phone). Apologies!

pollybuotte:
Yes @rosiealice , investigating how minimum conductance affects PFT performance is on my list of "want to dos".  It's currently pretty far from the top!

serbinsh:
@Qianyuxuan Here is a link to what I mentioned on the param file update thread: https://github.com/NGEET/fates/issues/616#issuecomment-614836638. This contains the basic steps needed to complete this PR.  Again we will not be using a namelist option for this, just a new param file switch which makes this PR easier

Qianyuxuan:
@serbinsh Thanks! I like the idea of using parameter file switch to choose between the two models. Now I am thinking about how do we deal with the different magnitudes of minimum conductance (g0) in the two models if we only use one parameter. Can we create a scalar in the codes to transfer that? Or do we need to create two parameter files with different values of g0?

parameter_files/fates_params_default.cdl
```diff
@@ -97,7 +82,7 @@ variables:
 	double fates_allom_hmode(fates_pft) ;
 		fates_allom_hmode:units = "index" ;
 		fates_allom_hmode:long_name = "height allometry function index." ;
-		fates_allom_hmode:possible_values = "1: O'Brien 1995; 2: Poorter 2006; 3: 2 parameter power law; 4: Chave 2014; 5: Martinez-Cano 2019." ;
+		fates_allom_hmode:possible_values = "1: O\'Brien 1995; 2: Poorter 2006; 3: 2 parameter power law; 4: Chave 2014; 5: Martinez-Cano 2019." ;
```
rgknox, line 85: nice catch

rgknox:
I looked over the new addition of the parameters.  Thanks for adding them @Qianyuxuan  I like the naming style (parameters are verbose and easy to understand what they mean), and units.  I see that you elected to keep medlyn slope and bb slope as separate parameters. That's fine with me, we can always unify them down the road if the team feels that is desireable.  

 
I propose the following:

1) we manually copy the contents of your last two commits, 565fd71 and 9ce3bc5 into a new branch that we are compiling in #616.  The changes in this PR shuffle around the order of many of the parameters, so I think it would be best to do this manually, and not use "git cherrypick".  Someone (probably me or @glemieux) will start getting a branch ready for #616, let us know if you (@Qianyuxuan ) would like to make the parameter changes, otherwise we can do it.  However, if you do, make sure not to modify any other parameters, even their ordering, that are outside the scope of this PR.

2) after the new parameters are in master, we go ahead and forcefully remove the last two commits in this branch

3) make last small adjustments to this PR so it works with master

Qianyuxuan:
@rgknox Thanks! For the update of parameter file, I first converted the latest ".cdl" file into the netcdf format and made changes. Then I converted the new netcdf file into ".cdl" file. I guess it caused the unexpected disorder in my current parameter file.  Can you suggest a better way to update the parameter file? Shall I edit the original ".cdl" file?

rgknox:
Thats a good question... Personally, I make changes directly to the CDL file in situations like these. Although, as a test, I would do an ncgen, and then ncdump the result, to see if get what I started with.

Qianyuxuan:
@rgknox Thanks for your advice! I updated the parameter file again. This time I reverted the parameter file with the original ".cdl" file (fortunately I saved essential files frequently)  first and made changes in the ".cdl" file. Then I tested the file by converting the file into the netcdf file and ran the model. The model worked well. I also merged the latest changes in host model and fates. Please feel free to let me know if I need to do more adjustments.

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -162,7 +169,7 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
     real(r8) :: mm_ko2             ! Michaelis-Menten constant for O2 (Pa)
     real(r8) :: co2_cpoint         ! CO2 compensation point (Pa)
     real(r8) :: btran_eff          ! effective transpiration wetness factor (0 to 1) 
-    real(r8) :: bbb                ! Ball-Berry minimum leaf conductance (umol H2O/m**2/s)
+    real(r8) :: stomatal_intercept_btran   ! minimum leaf conductance (umol H2O/m**2/s)
```
serbinsh, line 172: @Qianyuxuan is this the new g0*Bsw? which will go into both stomatal models?

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -243,17 +250,17 @@ real(r8) :: tcwood             ! Temperature response function for wood
...
-    real(r8), dimension(0:1) :: bbbopt 
+ 
 
     associate(  &
          c3psn     => EDPftvarcon_inst%c3psn  , &
          slatop    => EDPftvarcon_inst%slatop , & ! specific leaf area at top of canopy, 
                                                   ! projected area basis [m^2/gC]
-         woody     => EDPftvarcon_inst%woody)     ! Is vegetation woody or not? 
+         woody     => EDPftvarcon_inst%woody  , & ! Is vegetation woody or not?
+         stomatal_intercept   => EDPftvarcon_inst%stomatal_intercept ) !stomatal intercept for Ball-Berry model and Medlyn model
```
serbinsh, line 260: @Qianyuxuan shall we just comment as:

"Intercept or g0 of stomatal conductance model" ?

biogeophys/FatesPlantRespPhotosynthMod.F90
Qianyuxuan: Yes. "stomatal_intercept_btran" is the water-stressed g0, which will go into both stomatal models.

Qianyuxuan: @serbinsh Yes. "stomatal_intercept_btran" is the water-stressed g0, which will go into both stomatal models.

serbinsh:
@Qianyuxuan I didnt see an explicit logging of Medlyn stomatal error in the code, just a specific place in the code for logging Ball Berry errors.  Should we have something similar for Medlyn?  @rgknox ?  Or maybe I missed it?

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -1169,12 +1196,12 @@ subroutine LeafLayerPhotosynthesis(f_sun_lsl,         &  ! in
...
               ! Compare with Ball-Berry model: gs_mol = m * an * hs/leaf_co2_ppress p + b
               hs = (gb_mol*ceair + gs_mol* veg_esat ) / ((gb_mol+gs_mol)*veg_esat )
-              gs_mol_err = bb_slope(ft)*max(anet, 0._r8)*hs/leaf_co2_ppress*can_press + bbb
+              gs_mol_err = bb_slope(ft)*max(anet, 0._r8)*hs/leaf_co2_ppress*can_press + stomatal_intercept_btran
               
-              if (abs(gs_mol-gs_mol_err) > 1.e-01_r8) then
-                 write (fates_log(),*) 'CF: Ball-Berry error check - stomatal conductance error:'
-                 write (fates_log(),*) gs_mol, gs_mol_err
-              end if
+              if (abs(gs_mol-gs_mol_err) > 1.e-01_r8 .and.  (stomatal_model == 1)) then
```
serbinsh, line 1201: @Qianyuxuan @rgknox this part. It used to be specific to ball berry but maybe this is more general and should instead say something like "Stomatal model error check - stomatal conductance error:" or similar

I assume this is a relevant check for any stomatal model yes?

parameter_files/fates_params_default.cdl
```diff
@@ -951,6 +952,14 @@ data:
...
     0.03, 0.03, 0.03, 0.03 ;
 
+ fates_leaf_stomatal_intercept = 1000, 1000, 1000, 1000, 1000, 1000, 1000, 
+    1000, 1000, 1000, 1000, 1000 ;
+
+ fates_leaf_stomatal_slope_ballberry = 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8 ;
+
+ fates_leaf_stomatal_slope_medlyn = 4.1, 2.3, 2.3, 4.1, 4.4, 4.4, 4.7, 4.7, 
+    4.7, 2.2, 5.3, 1.6 ;
+
```
serbinsh, line 962: @Qianyuxuan looks good nice work on the updates here.   Question: I know you wrote up documentation to that is a companion to your code changes, we should make sure that makes it into the documentation for FATES/tech doc and that also includes these defaults and where they com from for Medlyn (USO).  have you shared that documentation you created on this PR yet?

biogeophys/FatesPlantRespPhotosynthMod.F90
Qianyuxuan: Sorry. I was thinking about it but then hesitated to include it. I found CLM5 didn't code the logging of Medlyn model error. I can definitely include those codes if it is a common procedure.

parameter_files/fates_params_default.cdl
Qianyuxuan: @serbinsh I haven't yet. I will drop a doc into the comment box here soon.

biogeophys/FatesPlantRespPhotosynthMod.F90
Qianyuxuan: @serbinsh okay.

rgknox:
> @Qianyuxuan I didnt see an explicit logging of Medlyn stomatal error in the code, just a specific place in the code for logging Ball Berry errors. Should we have something similar for Medlyn? @rgknox ? Or maybe I missed it?

Like here: https://github.com/NGEET/fates/blob/00b06c49b9e03e44603c3fb3e50c3b39790d20c9/biogeophys/FatesPlantRespPhotosynthMod.F90#L1201   ?

Yeah, if there is a way to quantify solution error and check-it, I think we should do that.

biogeophys/FatesPlantRespPhotosynthMod.F90
serbinsh: @Qianyuxuan OK.  @rgknox @glemieux this looks like a basic error/mismatch calculation, any reason to think it cant be generic to both models?  or maybe I am missing something?

rgknox: I think your right, I'd have to look more closely, but yup, I think that convergence error is used in both schemes

Qianyuxuan: Okay. I will add that to my next commit!

Qianyuxuan:
Here is the corresponding technical note:
[Note for Medlyn stomatal conductance model in FATES-LQY-0515.docx](https://github.com/NGEET/fates/files/4636984/Note.for.Medlyn.stomatal.conductance.model.in.FATES-LQY-0515.docx)

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -1193,15 +1191,20 @@ subroutine LeafLayerPhotosynthesis(f_sun_lsl,         &  ! in
...
-              gs_mol_err = bb_slope(ft)*max(anet, 0._r8)*hs/leaf_co2_ppress*can_press + stomatal_intercept_btran
-              
-              if (abs(gs_mol-gs_mol_err) > 1.e-01_r8 .and.  (stomatal_model == 1)) then
-               write (fates_log(),*) 'CF: Ball-Berry error check - stomatal conductance error:'
-               write (fates_log(),*) gs_mol, gs_mol_err
-            end if
+             
+             ! Compare with Medlyn model: gs_mol = 1.6*(1+m/sqrt(vpd)) * an/leaf_co2_ppress*p + b
+              if ( stomatal_model == 2 ) then
+                 gs_mol_err = 1.6*(1 + medlyn_slope(ft)/sqrt(vpd))*max(anet,0._r8)/leaf_co2_ppress*can_press + stomatal_intercept_btran
```
rgknox, line 1197: Thanks @Qianyuxuan , what is the 1.6, is that the diffusivity ratio of water to carbon?

h2o_co2_bl_diffuse_ratio

rgknox: wait, thats the boundary layer diffusion ratio, I think this is the stoma ratio:
h2o_co2_stoma_diffuse_ratio

Qianyuxuan: @rgknox Sorry. I forgot to change that. Yes. 1.6 is the factor that converts from conductance to CO2 to conductance to water vapour. I will adjust it soon.

serbinsh:
> Here is the corresponding technical note:
> [Note for Medlyn stomatal conductance model in FATES-LQY-0515.docx](https://github.com/NGEET/fates/files/4636984/Note.for.Medlyn.stomatal.conductance.model.in.FATES-LQY-0515.docx)

@Qianyuxuan We need to update the tech docs for FATES with this info.  Lets take a look at this together.

rgknox:
@Qianyuxuan , it looks like this PR is in good shape.  I don't think anymore action is required on your end. I will check back in after we have updated the master branch to include the parameters.

Qianyuxuan:
@walkeranthonyp @rgknox Sure! Thank you all for your thoughtful comments and continuous supports!

glemieux:
Testing in progress

Issue #377, opened 2018-05-04
we should cap canopy trim at 1.0


ckoven:
right now canopy trim can get slightly above 1.0, which is confusing and unexpected.  we should cap it at 1.0 to make it behave as expected.  relevant line of code is here:
https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L264

ckoven:
Closing because, while there may be edge cases where this happens, it doesn't seem to be a problem.

PR #623, opened 2020-03-23
Canopy trimming subroutine optimization update

### LINKED ISSUE #383: SLA profile implemented, weird interaction with canopy trimming
@aswann , @ckoven , and I have implemented a sla profile in FATES so that sla grows exponentially with canopy depth as supported by observations (Lloyd et al. 2010). In doing so, we've run into some strange behavior. We would like share these results with the wider community and ask for your feedback on our proposed next steps.

Briefly, we implemented a sla profile by replacing static sla values (same value at all canopy depths) with sla values that exponentially increase with depth in the canopy. These updates were made both to the calculation of LAI and, in the canopy trimming submodule, to the calculation of leaf cost. Both of these routines use 1/sla to represent the carbon cost of building a unit of leaf area. In our implementation, sla growth through the canopy uses the same coefficient as already used in FATES for nitrogen scaling (following Lloyd et al. 2010). Our expectation is that implementing this sla profile allows plants to grow more leaf area overall since leaf area at the bottom of the canopy costs less carbon to build.

We found that implementing this sla profile causes LAI to oscillate on a ~15yr timescale (light red line in plot below). This oscillation goes away when we turn trimming off (purple line at top) or turn trimming down by reducing the trim increment (how much canopy trimming increases each time trimming goes into effect) from 0.03 to 0.01 (light blue line) and 0.001 (dark red line). However, LAI is then sensitive to the trim increment parameter (trim_inc, abbreviated as tinc in plot legend), as shown by the difference between the light blue and dark red lines. For reference, the dark blue line in the plot shows LAI for a control simulation with trimming turned on; the yellow line shows a control simulation with trimming turned off. All simulations were run using a model version prior to recent conductance updates at BCI under 400 ppm CO2.

Given this weird behavior when the sla profile is implemented an trimming is on, we think our plan going forward will be to turn off canopy trimming. To bring LAI in line with observations (reduce from ~6.5 when trimming is off, purple line), our working plan is to alter the leaf biomass allometry so that less biomass is allocated to leaves. We were wondering if other FATES users/developers (e.g. @rosiealice) have thoughts on this? 

slaprofileexperiments

glemieux:
### Description:

This PR updates the `trim_canopy` subroutine to resolve #383 (LAI oscillation issue).  The update approaches the issue by circumventing the use of a trim increment.  Currently, the subroutine compares the `leaf_cost` against the `net_uptake` for all leaf layers within a cohort and decreases or increases that cohort's `canopy_trim` value by a fixed parameter defined increment.  This can result in trimming values that appear bang back and forth around some median value.

@ckoven hypothesized that a better method would be to determine the 'optimum' cumulative LAI for the whole cohort based on the cumulative LAI versus the "net-net" uptake (leaf cost subtracted from the net uptake per leaf layer).  The optimum in this case being the LAI associated with a zero net-net uptake.  This optimum LAI can then be used to compute the fractional `canopy_trim` value.  

The optimum cumulative LAI is computed by forming a least squares linear fit from the last `nll` layers of a given cohort and using the `dgels` routine from lapack to solve.  If the minimum number of leaf layers is not present `trim_canopy` reverts back to using the original trim increment method.

Note issue #638 was generated in the course of testing this PR.  The calculations of `optimum_laimem` can not be evaluated in the context of the larger `trim_canopy` implementation.

### Collaborators:

@ckoven @rgknox 

Yes, there will be changes to the trimming per cohort and as such LAI and related values will be different.

### Checklist:

### Test Results: 

CTSM (or) E3SM (specify which) test hash-tag: 242d6d14

CTSM (or) E3SM (specify which) baseline hash-tag: 242d6d14

FATES baseline hash-tag: 320a8fc0

Test Output:

Regression tests: All expected PASS

[ACRE comparison for 30 year BCI site run](https://github.com/glemieux/fates-jupyter/blob/develop/leaf-flutter/ACRE-output/issue383-comparison_plots.pdf)

### DIFF
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -525,13 +525,15 @@ contains
                 endif
 
                 ! Construct the arrays for a least square fit of the net_net_uptake versus the cumulative lai
-                nnu_clai_a(1,1) = nnu_clai_a(1,1) + 1 ! Increment for each layer used
-                nnu_clai_a(1,2) = nnu_clai_a(1,2) + currentCohort%year_net_uptake(z) - currentCohort%leaf_cost
-                nnu_clai_a(2,1) = nnu_clai_a(1,2)
-                nnu_clai_a(2,2) = nnu_clai_a(2,2) + (currentCohort%year_net_uptake(z) - currentCohort%leaf_cost)**2
-                nnu_clai_b(1,1) = nnu_clai_b(1,1) + cumulative_lai
-                nnu_clai_b(2,1) = nnu_clai_b(2,1) + (cumulative_lai * & 
-                                 (currentCohort%year_net_uptake(z) - currentCohort%leaf_cost))
+                if (currentCohort%nv - z < nll) then  ! Only for nll layers
+                  nnu_clai_a(1,1) = nnu_clai_a(1,1) + 1 ! Increment for each layer used
+                  nnu_clai_a(1,2) = nnu_clai_a(1,2) + currentCohort%year_net_uptake(z) - currentCohort%leaf_cost
+                  nnu_clai_a(2,1) = nnu_clai_a(1,2)
+                  nnu_clai_a(2,2) = nnu_clai_a(2,2) + (currentCohort%year_net_uptake(z) - currentCohort%leaf_cost)**2
+                  nnu_clai_b(1,1) = nnu_clai_b(1,1) + cumulative_lai
+                  nnu_clai_b(2,1) = nnu_clai_b(2,1) + (cumulative_lai * & 
+                                    (currentCohort%year_net_uptake(z) - currentCohort%leaf_cost))
+                end if
 
                 ! Check leaf cost against the yearly net uptake for that cohort leaf layer
                 if (currentCohort%year_net_uptake(z) < currentCohort%leaf_cost) then
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -413,6 +413,9 @@ contains
     real(r8) :: nnu_clai_b(m,nrhs)        ! RHS of linear least squares fit
     real(r8) :: work(workmax)             ! work array
 
+    real(r8) :: initial_trim              ! Initial trim
+    real(r8) :: optimum_trim              ! Optimum trim value 
+
     !----------------------------------------------------------------------
 
     ipatch = 1 ! Start counting patches
@@ -431,10 +434,13 @@ contains
        currentCohort => currentPatch%tallest
        do while (associated(currentCohort)) 
 
+         ! Save off the incoming trim
+         initial_trim = currentCohort%canopy_trim
+
           ! Add debug diagnstic output to determine which cohort
           if (debug) then
             write(fates_log(),*) 'Current cohort:', icohort
-            write(fates_log(),*) 'Starting canopy trim:', currentCohort%canopy_trim
+            write(fates_log(),*) 'Starting canopy trim:', initial_trim
             write(fates_log(),*) 'Starting laimemory:', currentCohort%laimemory
           endif   
 
@@ -618,6 +624,16 @@ contains
                ! write(fates_log(),*) 'LLSF optimium LAI info:', info
                ! write(fates_log(),*) 'LAI fraction (optimum_lai/cumulative_lai):', nnu_clai_b(1,1) / cumulative_lai
             endif
+
+            ! Calculate the optimum trim based on the initial canopy trim value
+            optimum_trim = (nnu_clai_b(1,1) / cumulative_lai) * initial_trim
+
+            ! Determine if the optimum trim value makes sense.  The smallest cohorts tend to have unrealistic fits.
+            if (optimum_trim > 0. .and. optimum_trim < 1.) then
+               currentCohort%canopy_trim = optimum_trim
+               trimmed = .true.
+            endif
+
          endif
 
           ! Reset activity for the cohort for the start of the next year
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -398,8 +398,15 @@ contains
     integer :: ipatch
     integer :: icohort
 
-    ! LAPACK least squares fit variables (A*X = B)
-    integer :: nll = 4                    ! Number of leaf layers to fit a regression to for calculating the optimum lai
+    ! LAPACK linear least squares fit variables
+    ! The standard equation for a linear fit, y = mx + b, is converted to a linear system, AX=B and has
+    ! the form: [n  sum(x); sum(x)  sum(x^2)] * [b; m]  = [sum(y); sum(x*y)] where 
+    ! n is the number of leaf layers 
+    ! x is yearly_net_uptake minus the leaf cost aka the net-net uptake
+    ! y is the cumulative lai for the current cohort
+    ! b is the y-intercept i.e. the cumulative lai that has zero net-net uptake
+    ! m is the slope of the linear fit
+    integer :: nll = 3                    ! Number of leaf layers to fit a regression to for calculating the optimum lai
     character(1) :: trans = 'N'           ! Input matrix is not transposed
     
     integer, parameter :: m = 2, n = 2    ! Number of rows and columns, respectively, in matrix A
@@ -410,8 +417,8 @@ contains
     integer :: lwork                      ! Dimension of work array
     integer :: info                       ! Procedure diagnostic ouput
     
-    real(r8) :: nnu_clai_a(m,n)           ! LHS of linear least squares fit
-    real(r8) :: nnu_clai_b(m,nrhs)        ! RHS of linear least squares fit
+    real(r8) :: nnu_clai_a(m,n)           ! LHS of linear least squares fit, A matrix
+    real(r8) :: nnu_clai_b(m,nrhs)        ! RHS of linear least squares fit, B matrix
     real(r8) :: work(workmax)             ! work array
 
     real(r8) :: initial_trim              ! Initial trim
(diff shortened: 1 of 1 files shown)


glemieux:
Note that the code partially addresses issue #377 by setting the optimum value to be less than one: https://github.com/NGEET/fates/pull/623/files#diff-b2580903c042bbb143bb267a7b882e3fR632-R635

glemieux:
Regression tests on cheyenne all expected PASS:
```/glade/u/home/glemieux/scratch/clmed-tests/PR623-canopytrim--C242d6d14-F76ec9103.fates.cheyenne.intel
/glade/u/home/glemieux/scratch/clmed-tests/PR623-canopytrim--C242d6d14-F76ec9103.fates.cheyenne.gnu
``` 
`ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesSizeAgeMort` FAILED on a `NPLANT_CACLS` and `NPLANT_CAPF` which is a known bug. 
`SMS_Lm13.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesColdDef` FAILED on the baseline comparison.  Values appear to be consistent and expected given that this is the longest run which should start showing differences in the emerging values based on a change to `canopy_trim`.

glemieux:
[100 year BCI acre comparison plots](https://github.com/glemieux/fates-jupyter/blob/develop/leaf-flutter/ACRE-output/pr623_plots.pdf).  These runs were done with the standard fates variable outputs; I can add more to the plots if necessary.

rgknox:
@glemieux @ckoven  is it possible that bfr_per_bleaf is uninitialized in some cases? I'm looking to see if that is possible, but maybe you can strike that for me:

https://github.com/NGEET/fates/pull/623/files#diff-b2580903c042bbb143bb267a7b882e3fR479

rgknox:
ok, I see that it is only used inside the same conditional where it is set.  nvm

rgknox:
Sorry if this is something that we have discussed ad nauseum and I've forgotten, but I see that we are only using slatop to calculate the construction costs per layer, shouldn't be using the actual costs which is somwhere between slatop and slamax?
https://github.com/NGEET/fates/pull/623/files#diff-b2580903c042bbb143bb267a7b882e3fR515

rgknox:
I think you had been bringing this point up before @glemieux .  I see here:
https://github.com/NGEET/fates/pull/623/files#diff-b2580903c042bbb143bb267a7b882e3fR511-R519

that we are kind-of re-constructing the wheel on calculating the actual SLA at the layer of interest.  I think we discussed if this PR should try to tidy this stuff up, and build some functions that can be used consistently in different places so there is no repeated code that could diverge. I still think it is fine to put this in another change-set, but just wanted to acknowledge it here

rgknox:
This is kind of a weird clause:

https://github.com/NGEET/fates/pull/623/files#diff-b2580903c042bbb143bb267a7b882e3fR579

It seems the intention here is really to filter out new plants. I think "isnew" might be more appropriate... Although, I think this routine comes after the daily growth happens, so at that point no plants are new.  So this clause only rejects plants that have not incremented their height growth yet.  Still don't know why we do that...?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -503,42 +553,111 @@ subroutine trim_canopy( currentSite )
...
+                   endif ! trim limit check
+                endif ! net uptake check
+             endif ! leaf activity check 
+          enddo ! z, leaf layer loop
+
+          ! Compute the optimal cumulative lai based on the cohort net-net uptake profile if at least 2 leaf layers
+          if (nnu_clai_a(1,1) > 1) then
+
+            ! Compute the optimum size of the work array
+            lwork = -1 ! Ask sgels to compute optimal number of entries for work
```
rgknox, line 600: checked through the lapack dgels description.  Verified that nnu_cla_b is not overwritten when work dne 0

biogeochem/EDPhysiologyMod.F90
```diff
@@ -503,42 +553,111 @@ subroutine trim_canopy( currentSite )
...
+                  nnu_clai_a(1,2) = nnu_clai_a(1,2) + currentCohort%year_net_uptake(z) - currentCohort%leaf_cost
+                  nnu_clai_a(2,1) = nnu_clai_a(1,2)
+                  nnu_clai_a(2,2) = nnu_clai_a(2,2) + (currentCohort%year_net_uptake(z) - currentCohort%leaf_cost)**2
+                  nnu_clai_b(1,1) = nnu_clai_b(1,1) + cumulative_lai
+                  nnu_clai_b(2,1) = nnu_clai_b(2,1) + (cumulative_lai * & 
+                                    (currentCohort%year_net_uptake(z) - currentCohort%leaf_cost))
+                end if
+
+                ! Check leaf cost against the yearly net uptake for that cohort leaf layer
+                if (currentCohort%year_net_uptake(z) < currentCohort%leaf_cost) then
```
rgknox, line 569: @glemieux , why do we allow wholesale trimming of the leaf layer in these conditions?  I thought the idea was to let the regression identify the intersection point and just go with that value. It looks like the regression would also over-write whatever was calculated here, so it seems unnecessary right?  Also, does the final solution adhere to the trim-limit?

biogeochem/EDPhysiologyMod.F90
ckoven: could we add some documentation here describing the arguments that dgels is expecting, etc, like what we have for the lapack dgesv call here: https://github.com/NGEET/fates/blob/master/biogeophys/FatesPlantHydraulicsMod.F90#L4812?

ckoven: actually i see now that's what is going on in the code starting at line 400 above... sorry.

glemieux: At @rgknox suggestion I am writing up a technote to help describe the setup: https://github.com/glemieux/fates-jupyter/blob/develop/leaf-flutter/pr623-linear-solver.ipynb

rgknox: I also like the idea of some comments in line.  Something that would explain what variables are equated to the matrix terms: ie  Y = X * b, and how the X and Y terms are manipulated into arguments of dgels()

biogeochem/EDPhysiologyMod.F90
glemieux: We do this in case we have less than `nll` leaf layers to work with for a given cohort.  In that case I just decided to let it stay with the old trimming methodology, baring some other rationale.  There are two thoughts I had on this:

1) I wasn't sure what we should do if we have a cohort with just one leaf layer come in to the routine `nv = 1`.  Presumably that would be for a new cohort so maybe we'd want to just leave it at the default starting trim fraction?
2) Right now `nll` is somewhat randomly hard coded to `4` for all cohorts regardless of number of leaf layer, `nv`.  I was thinking that there might be 'better' optimum intercepts if the `nll` layer is greater for cohorts with more leaf layers (up to a point).  That said, I could make the value of `nll` dependent on `nv` so that we'd never have to use the old method except for `nv = 1`.  Alternatively, I just set this to `nll = 2` and call it a day.

biogeochem/EDPhysiologyMod.F90
glemieux: Roger that.  I'll flesh out the comments in that part.

glemieux:
> I think you had been bringing this point up before @glemieux . I see here:
> https://github.com/NGEET/fates/pull/623/files#diff-b2580903c042bbb143bb267a7b882e3fR511-R519
> 
> that we are kind-of re-constructing the wheel on calculating the actual SLA at the layer of interest. I think we discussed if this PR should try to tidy this stuff up, and build some functions that can be used consistently in different places so there is no repeated code that could diverge. I still think it is fine to put this in another change-set, but just wanted to acknowledge it here
Roger that.  I've started working up a branch for another PR: https://github.com/glemieux/fates/blob/9e04bf7434b7c06fd5138dc2801755a6c7d85624/biogeochem/FatesAllometryMod.F90#L2213-L2278

glemieux:
> This is kind of a weird clause:
> 
> https://github.com/NGEET/fates/pull/623/files#diff-b2580903c042bbb143bb267a7b882e3fR579
> 

Yeah, my assumption was that we wanted to make sure that we weren't trimming new plants coming in.  `trim_canopy` gets called at the end of `ed_update_site`; is that the routine called at the end of daily growth?

rgknox:
Checking if the cohorts height is greater than the minimum should achieve the same effect as filtering on the "%isnew" flag.  The only situation it would not, and this could be meaningful, is this logic will filter out plants that have not grown in size yet because they have insufficient carbon balance (even though they , yet these are the plants that would potentially benefit from optimizing their allocation.  I think we should at least create an issue of this.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -453,8 +503,8 @@ subroutine trim_canopy( currentSite )
              lai_current       = min(leaf_inc, currentCohort%treelai - lai_layers_above)
              cumulative_lai    = lai_canopy_above + lai_layers_above + 0.5*lai_current
              
-             if (currentCohort%year_net_uptake(z) /= 999._r8)then !there was activity this year in this leaf layer.
-             
+             !there was activity this year in this leaf layer.  This should only occur for bottom most leaf layer
+             if (currentCohort%year_net_uptake(z) /= 999._r8)then
```
rgknox, line 507: These clauses where we compare real numbers for equivalence seem dangerous to me.  But as far as I know, the compilers seem to get it right as long as year_net_uptake is set with a constant and not the result of math. I'm always in favor of this style:   
> if( abs(this - that) > nearzero ) then

glemieux: Per our discussion, I created issue #644 to resolve this in a future, more broad canopy trimming refactor.

glemieux:
> Checking if the cohorts height is greater than the minimum should achieve the same effect as filtering on the "%isnew" flag. The only situation it would not, and this could be meaningful, is this logic will filter out plants that have not grown in size yet because they have insufficient carbon balance (even though they , yet these are the plants that would potentially benefit from optimizing their allocation. I think we should at least create an issue of this.
 Created issue #645

glemieux:
@rgknox I have an alternate branch with the adjustments to the `cumulative_lai` calculation you mentioned in our weekly meeting here: https://github.com/glemieux/fates/blob/issue-383-linearfit-trimfraction-cohortcumulativelai/biogeochem/EDPhysiologyMod.F90#L501-L628

The results comparing the two versions are pretty much spot on in the acre outputs: https://github.com/glemieux/fates-jupyter/blob/develop/leaf-flutter/acre-output/100year-cumulativelaitest_plots.pdf.  Interestingly the `TSAI` shows the most difference compared to the difference in the `TLAI`.  A comparison against the baseline and the updated `cumulative_lai` branch is here: https://github.com/glemieux/fates-jupyter/blob/develop/leaf-flutter/acre-output/100year-base-cumulativelaitest_plots.pdf

glemieux:
@rgknox I've added some more comments at your suggestion.  Would you give me your feedback when you get a chance?

rgknox:
All PASS:
/glade/scratch/rgknox/clmed-tests/fates-sci.1.37.0_api.11.2.0-clm5.0.30-Ce33b4658-Fc1d1372f.fates.cheyenne.gnu
/glade/scratch/rgknox/clmed-tests/fates-sci.1.37.0_api.11.2.0-clm5.0.30-Ce33b4658-Fc1d1372f.fates.cheyenne.intel

PR #628, opened 2020-04-02
update grass carbon allocation

xuchongang:
grass carbon allocation after reaching maximum height
### Description:
change the FatesAllometryMod.F90 so that when grass reach maximum height, put the carbon to storage instead of structural carbon. 
Simulation of grass PFTs will be changed to allow storage carbon to grow larger plants for next year

### Checklist:

Preliminary evaluations show model performs as expected

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/FatesAllometryMod.F90
+++ b/biogeochem/FatesAllometryMod.F90
@@ -394,6 +394,7 @@ contains
 	       dbh_maxh     => EDPftvarcon_inst%allom_dbh_maxheight(ipft), &
 	       woody        => EDPftvarcon_inst%woody(ipft), &
                allom_amode  => EDPftvarcon_inst%allom_amode(ipft))
+      d1 = d
       !for grasses, no growth of structural wood after maximum height
       if(d >= dbh_maxh .and. woody .ne. itrue)d1 = dbh_maxh 
       select case(int(allom_amode))
--- a/biogeochem/FatesAllometryMod.F90
+++ b/biogeochem/FatesAllometryMod.F90
@@ -997,7 +997,7 @@ contains
            write(fates_log(),*) 'Aborting'
            call endrun(msg=errMsg(sourcefile, __LINE__))
          end select   
-	 dbstoredd = dbstoredd + dbagwdd   
+	 dbstoredd = dbstoredd + dbagwdd/agb_frac   
        endif
        
      end associate
--- a/biogeochem/FatesAllometryMod.F90
+++ b/biogeochem/FatesAllometryMod.F90
@@ -381,6 +381,7 @@ contains
     real(r8),intent(out),optional :: dbagwdd  ! change in agbw per diameter [kgC/cm]
 
     real(r8)               :: h       ! height
+    real(r8)               :: d1      ! constrained plant diameter [cm]
     real(r8)               :: dhdd    ! change in height wrt d
 
     associate( p1           => EDPftvarcon_inst%allom_agb1(ipft), &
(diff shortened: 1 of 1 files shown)


rgknox:
@xuchongang 
 - I think this implementation is fine as long as the below ground structural biomass hypothesis stays the same, but if the below ground structural biomass changes so that below is not tied to above, then you will get below-ground structure growth after max_dbh.
- In the calculation of storage; I see that the derivative of storage gets increased by the structural biomass, but the target value (bstore) does not.  I think this would lead to inconsistencies and break the solver, no?

xuchongang:
@rgknox , thanks for the comments.  Please see my response below:

Comment: I think this implementation is fine as long as the below ground structural biomass hypothesis stays the same, but if the below ground structural biomass changes so that below is not tied to above, then you will get below-ground structure growth after max_dbh.

Answer:  I have set  dbstoredd = dbstoredd + dbagwdd/agb_frac   and also I have set dbagwdd to zero when the dbh reach max_dbh. Isn't the belowground component is divivided by agb_frac? If I set  dbagwdd to zero, then the beloground part should also be zero? Right?

Comment: In the calculation of storage; I see that the derivative of storage gets increased by the structural biomass, but the target value (bstore) does not. I think this would lead to inconsistencies and break the solver, no?

Answer: Good point. In my testing, the solvers works fine. I did increase the cushion parameter to be large (3.75).  But this does seem to be a potential issue. Is there a better way to fix this?

xuchongang:
@rgknox , I further checked into the solver, which uses dbstoreddt to grow the plants. Because the solver did not put a constrain on storage/leaf biomass for target biomass, it runs without problem. So I do not see an issue here. The only impact will be on the carbon allocations when storage pool is large. It will mostly grow leaf/roots for turnover but again make the carbon storage large, which make sense in the real world.

rgknox:
If the allometry target for storage does not match the derivative, it will generate problems at some point, I made some quick modifications to show how I would update the storage target to match:

https://github.com/xuchongang/fates-1/pull/12/files

I feel like a better way to do this would be to add a new storage and structural allocation hypothesis though. (ie something that would be a its own distinct  allom_amode and allom_smode for grasses).

rgknox:
the key in that PR is line 1020:  https://github.com/xuchongang/fates-1/pull/12/files#diff-6403a5d68498c9a46a99461bedeb157aR1020

xuchongang:
@rgknox, thank you for the pull request that updates the bstore state variable.  I think it will resolve the inconsistency of the derivative and the state variables. 
I agree with you it might be better to have a distinct allom_amode and allom_smode for grasses; however, right now, the allocation modes are not PFT specific yet. Do we have plans to make those PFT specific? I can update this allocation when the allometry mode became PFT-specific.

rgknox:
The allocation modes are indeed pft specific:
checkout fates_allom_stmode: 
https://github.com/NGEET/fates/blob/master/parameter_files/fates_params_default.cdl#L121-L124

xuchongang:
@rgknox , that is great to hear and I will update the codes correspondingly.

rgknox:
Thats great @xuchongang , let me me know if I can help.

xuchongang:
@rgknox, did some further testing on the codes. In our cases, we still want to use the ratio of leaf to storage to calculate the target carbon. The additional carbon from structural will be the carbon more than it should be. For my previous testing, it seems that the solver works fine. Did you have any concerns if  I revert it back?

rgknox:
I'm not sure I follow 100%.  Maybe could you draft up an alternative PR, or just a new branch with the proposed set of changes and link it here, and we can take a look?

xuchongang:
@rgknox , before I proceed ahead, I would like to have a quick chat with you some time next week.

xuchongang:
@rgknox , in summary, dabgw1 will never larger than dabgw0, as the dbh will not grow more than dbh_max.

Issue #636, opened 2020-04-24
Site-level-vegetation temperature bug


xuchongang:
When running ELM(FATES), t_veg24_si will not be updated and cause errors in phenology (never cold) and cause vegetation to die off from carbon starvation. See the detailed description and solution in the attached pdf file. Please let me and @rgknox  know if you have any suggestions for the solution. 
[Bug of site-level-vegetation temperature.pdf](https://github.com/NGEET/fates/files/4530536/Bug.of.site-level-vegetation.temperature.pdf)

rgknox:
@xuchongang , I just talked this over with @ckoven and @glemieux and we feel the fix to this is simple, and can be fixed on the FATES side of the code.  
The solution is to just calculate the area weighted mean of the patch-level t_veg_24 (a variable we get along with the site level anyway) ourselves in fates.  I will self assign.

PR #637, opened 2020-04-27
Removing bc_in%t_veg24_si

rgknox:
### Description:

This set of changes removes usage of t_veg24_si.  @xuchongang identified in #636 that ELM is potentially deactivating the bare-ground patch,, and at the least is generating non dynamic changes on that patch, and we should not rely on that patch's temperature to make site-level decisions in FATES.  The fix applied here, removes its usage in FATES, and instead either takes the mean of the patch temperatures (for phenology timing), or uses the temperature from the patch of interest (for mortality).

### Collaborators:

@xuchongang 

Yes, this should change answers.  This will generate subtle changes to phenology timing on cold-decidous, as well as freezing mortality in all plants.  In most cases, I'm guessing the differences will be very subtle, but potentially not.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDMortalityFunctionsMod.F90
+++ b/biogeochem/EDMortalityFunctionsMod.F90
@@ -62,7 +62,7 @@ contains
     real(r8),intent(out) :: smort  ! size dependent senescence term
     real(r8),intent(out) :: asmort ! age dependent senescence term 
 
-
+    integer  :: ifp
     real(r8) :: frac  ! relativised stored carbohydrate
     real(r8) :: leaf_c_target      ! target leaf biomass kgC
     real(r8) :: store_c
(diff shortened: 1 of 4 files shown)


rgknox:
forgot to area weight the site level temperature during phenology, applying correction

rgknox:
all expected PASS;

/glade/scratch/rgknox/clmed-tests/fates-sci.1.35.1_api.10.0.0-clm5.0.30-C242d6d14-F987f521a.fates.cheyenne.intel/
/glade/scratch/rgknox/clmed-tests/fates-sci.1.35.1_api.10.0.0-clm5.0.30-C242d6d14-F987f521a.fates.cheyenne.gnu/

Issue #630, opened 2020-04-09
History Refactor Wish-list


rgknox:
There are a few things that bother me about history output.

1) It is both not clear to the user, and not implemented with consistency, how we treat non-fates columns in the upscaling to grid-cell level output.  All of our output in the history file, ultimately gets upscaled to the grid-scale.   This issue is roughly the same as #535, because how we flush values on non-fates columns essentially controls how/if those values are part of the final averaging.  In some cases we tell non fates columns to be zero (adds a zero to the average), and in some cases we tell it to use an ignore flag.  

My position is that if the output variable is on the fates side of the code, the diagnostic is inherently relevant only to naturally vegetated columns within FATES, and all other columns should be ignored.  If we want a diagnostic that uses zeros from other columns, or real information from other columns, this should be defined on the host-model side, and FATES should pass the boundary conditions necessary to make the calculation.

2) Get rid of these patch-scale diagnostic types, (ie _PA). These are confusing because it looks like we can have patch-scale output, but what is really happening is that we are just asking CLM/ELM to upscale from patches to grid for us.  We already do this, and having two ways of going about the same thing is confusing.

3) The filling of all history diagnostics happens in 1 central location, FatesHistoryInterfaceMod. This makes things complicated, particularly because the daily calls happen after all of our dynamics completes.  If we want a diagnostic on growth or mortality rates (for example) , we have to then upscale the growth rates and save them to a site level intermediary variable, so that the information does not get messed-up during disturbance.  Another example, if a cohort becomes terminated, we can't diagnose its rates because it ceases to exist.  Long story short, if we were allowed to immediately update the history pointer arrays that are defined in FatesHistoryInterfaceMod when the salient variable is meaningfully updated, we would be able to remove all of these site-level intermediary arrays.

rgknox:
Also:

All history diagnostic written by FATES, should have a "FATES_" namespace prefix to the variable.

No output variables should have a "COL" or "PA" to indicate if it is patch or column, because the refactor will remove patch level averaging, everything is implicitly a calculation of fates columns.

ckoven:
@rgknox per your comment here about the _pa history variables.  about half of these are fire-related.  I am going through some of the fire history variables right now and will add the task of moving all of those current indexed directly by patch to the standard way we handle site-level variables to that effort.

glemieux:
[...]

PR #640, opened 2020-04-28
Updates to fire history variables

ckoven:
A bunch of updates to fire-related history variables, including moving all of the ones that had been on the patch structure off of it, fixing the fire carbon loss variable to not be zero'd out before history writing, and adding some further diagnostics.

### Description:
Sort of a grab bag of changes, but all related to outputting fire behavior information to the history fields:
1. Eleven variables that had been on the patch structure needed to get moved off of it (so that we can deprecate that way of doing history weighting, see #630).
2. the variable `Fire_Closs` was just passing zeros, because the structure it is on was getting zeroed before the history call. I also added another variable, `FIRE_FLUX `, which has the PARTEH ELEMENT dimension, so that when we start adding nutrients we can keep track of fire-related N and P losses too.  Its possible that we could thus delete `Fire_Closs` as redundant to the carbon index of `FIRE_FLUX` but I haven't done that here.
3. some of the variables, such as fire intensity and rate of spread, were being averaged in an evenly-weighted manner, whether or not a fire had occurred on a given patch on a given timestep.  Since we typically look at monthly or longer period for time averaging, these were thus getting diluted by the many times when no fire was occurring.  So I added a few new variables, like the product (intensity * area).  That way one can calculate the longer-term burned-area-weighted mean intensity as the ratio mean(intensity * area) / mean(area).  The variables that I've added this for are intensity, rate of spread, and total fuel consumption.  The names on this are a bit awkward, but so far I am thinking that calling them XXX_AREA_PRODUCT is simplest so that one can see that they only make sense if you divide by burned area; so they are currently: `FIRE_INTENSITY_AREA_PRODUCT`, `FIRE_ROS_AREA_PRODUCT`, and `FIRE_TFC_ROS_AREA_PRODUCT`. I don't love that, but it seemed safer than calling them averages, since they are not actually average quantities, but needed to calculate the averages.
4. I added three new patch-age-structured variables and one new fuel-pool-structured variable so as to be able to better see what's going on in the fire model.  The three patch-age-structured variables are: `AREA_BURNT_BY_PATCH_AGE`, `FIRE_INTENSITY_BY_PATCH_AGE`, and `SUM_FUEL_BY_PATCH_AGE`.  The fuel-pool-structured variable is the fraction of fuel combusted, which I output as in (3) above, as multiplied by burned area so that one can calculate burned-area-weighted means of this.

### Collaborators:
changes based on discussions with @jkshuman, @rgknox 

Expected changes only to the history variables that I've messed with in some way.  I'm not totally sure that I understand why changing the _pa variables to use the site-level weighting leads to greater than roundoff changes, but I think maybe it has to do with patches where the canopy isn't closed not having been weighted properly in the old system (which is part of the reason why we want to get rid of that).

### Checklist:

### Test Results:
none as yet.

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -141,7 +141,6 @@ module FatesHistoryInterfaceMod
   integer :: ih_trimming_pa
   integer :: ih_area_plant_pa
   integer :: ih_area_treespread_pa
-  integer :: ih_nesterov_fire_danger_pa
   integer :: ih_spitfire_ROS_pa
   integer :: ih_effect_wspeed_pa
   integer :: ih_TFC_ROS_pa
@@ -268,6 +267,7 @@ module FatesHistoryInterfaceMod
   integer :: ih_dleafon_si
   integer :: ih_meanliqvol_si
 
+  integer :: ih_nesterov_fire_danger_si
 
   integer :: ih_nplant_si_scpf
   integer :: ih_gpp_si_scpf
@@ -326,7 +326,6 @@ module FatesHistoryInterfaceMod
   integer :: ih_ar_frootm_si_scpf
   
   integer :: ih_c13disc_si_scpf
-
 
   ! indices to (site x scls [size class bins]) variables
   integer :: ih_ba_si_scls
@@ -433,6 +432,10 @@ module FatesHistoryInterfaceMod
   integer :: ih_c_lblayer_si_age
   integer :: ih_agesince_anthrodist_si_age
   integer :: ih_secondaryforest_area_si_age
+  integer :: if_area_burnt_si_age
+  ! integer :: if_fire_rate_of_spread_front_si_age
+  ! integer :: if_fire_intensity_si_age
+  integer :: if_fire_sum_fuel_si_age
 
   ! indices to (site x height) variables
   integer :: ih_canopy_height_dist_si_height
@@ -1733,7 +1736,7 @@ end subroutine flush_hvars
                hio_recruitment_si_pft  => this%hvars(ih_recruitment_si_pft)%r82d, &
                hio_mortality_si_pft    => this%hvars(ih_mortality_si_pft)%r82d, &
                hio_crownarea_si_pft    => this%hvars(ih_crownarea_si_pft)%r82d, &
-               hio_nesterov_fire_danger_pa => this%hvars(ih_nesterov_fire_danger_pa)%r81d, &
+               hio_nesterov_fire_danger_si => this%hvars(ih_nesterov_fire_danger_si)%r81d, &
                hio_spitfire_ros_pa     => this%hvars(ih_spitfire_ROS_pa)%r81d, &
                hio_tfc_ros_pa          => this%hvars(ih_TFC_ROS_pa)%r81d, &
                hio_effect_wspeed_pa    => this%hvars(ih_effect_wspeed_pa)%r81d, &
@@ -1909,6 +1912,10 @@ end subroutine flush_hvars
                hio_woodproduct_si                 => this%hvars(ih_woodproduct_si)%r81d, &
                hio_agesince_anthrodist_si_age     => this%hvars(ih_agesince_anthrodist_si_age)%r82d, &
                hio_secondaryforest_area_si_age    => this%hvars(ih_secondaryforest_area_si_age)%r82d, &
+               hio_area_burnt_si_age              => this%hvars(ih_area_burnt_si_age)%r82d, &
+               ! hio_fire_rate_of_spread_front_si_age  => this%hvars(ih_fire_rate_of_spread_front_si_age)%r82d, &
+               ! hio_fire_intensity_si_age          => this%hvars(ih_fire_intensity_si_age)%r82d, &
+               hio_fire_sum_fuel_si_age           => this%hvars(ih_fire_sum_fuel_si_age)%r82d, &
                hio_canopy_height_dist_si_height   => this%hvars(ih_canopy_height_dist_si_height)%r82d, &
                hio_leaf_height_dist_si_height     => this%hvars(ih_leaf_height_dist_si_height)%r82d, &
                hio_litter_moisture_si_fuel        => this%hvars(ih_litter_moisture_si_fuel)%r82d, &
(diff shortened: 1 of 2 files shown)


glemieux:
This looks good to me.  Regression testing underway.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -4001,65 +4041,106 @@ subroutine define_history_vars(this, initialize_variables)
...
 
     call this%set_history_var(vname='FIRE_INTENSITY', units='kJ/m/s',          &
          long='spitfire fire intensity: kJ/m/s', use_default='active',          &
-         avgflag='A', vtype=patch_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
-         ivar=ivar, initialize=initialize_variables, index = ih_fire_intensity_pa )
+         avgflag='A', vtype=site_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
+         ivar=ivar, initialize=initialize_variables, index = ih_fire_intensity_si )
+
+    call this%set_history_var(vname='FIRE_INTENSITY_AREA_PRODUCT', units='kJ/m/s',          &
+         long='spitfire product of fire intensity and burned area (divide by FIRE_AREA to get area-weifghted mean intensity)', &
```
jkshuman, line 4078:
```suggestion
         long='spitfire product of fire intensity and burned area (divide by FIRE_AREA to get area-weighted mean intensity)', &
```

main/FatesHistoryInterfaceMod.F90
```diff
@@ -4001,65 +4041,106 @@ subroutine define_history_vars(this, initialize_variables)
...
+         ivar=ivar, initialize=initialize_variables, index = ih_fire_intensity_area_product_si )
 
     call this%set_history_var(vname='FIRE_AREA', units='fraction',             &
          long='spitfire fire area burn fraction', use_default='active',                    &
-         avgflag='A', vtype=patch_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
-         ivar=ivar, initialize=initialize_variables, index = ih_fire_area_pa )
+         avgflag='A', vtype=site_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
+         ivar=ivar, initialize=initialize_variables, index = ih_fire_area_si )
 
     call this%set_history_var(vname='fire_fuel_mef', units='m',                &
```
jkshuman, line 4088:
```suggestion
    call this%set_history_var(vname='FIRE_FUEL_MEF', units='m',                &
```

main/FatesHistoryInterfaceMod.F90
```diff
@@ -4001,65 +4041,106 @@ subroutine define_history_vars(this, initialize_variables)
...
          long='spitfire fuel moisture',  use_default='active',                  &
-         avgflag='A', vtype=patch_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
-         ivar=ivar, initialize=initialize_variables, index = ih_fire_fuel_mef_pa )
+         avgflag='A', vtype=site_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
+         ivar=ivar, initialize=initialize_variables, index = ih_fire_fuel_mef_si )
 
     call this%set_history_var(vname='fire_fuel_bulkd', units='kg biomass/m3',              &
```
jkshuman, line 4093:
```suggestion
    call this%set_history_var(vname='FIRE_FUEL_BULKD', units='kg biomass/m3',              &
```

main/FatesHistoryInterfaceMod.F90
```diff
@@ -4001,65 +4041,106 @@ subroutine define_history_vars(this, initialize_variables)
...
+         ivar=ivar, initialize=initialize_variables, index = ih_fire_fuel_bulkd_si )
 
     call this%set_history_var(vname='FIRE_FUEL_EFF_MOIST', units='m',          &
          long='spitfire fuel moisture', use_default='active',                   &
-         avgflag='A', vtype=patch_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
-         ivar=ivar, initialize=initialize_variables, index = ih_fire_fuel_eff_moist_pa )
+         avgflag='A', vtype=site_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
+         ivar=ivar, initialize=initialize_variables, index = ih_fire_fuel_eff_moist_si )
 
     call this%set_history_var(vname='fire_fuel_sav', units='per m',                &
```
jkshuman, line 4103:
```suggestion
    call this%set_history_var(vname='FIRE_FUEL_SAV', units='per m',                &
```

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5218,6 +5299,11 @@ subroutine define_history_vars(this, initialize_variables)
           avgflag='A', vtype=site_r8, hlms='CLM:ALM', flushval=hlm_hio_ignore_val,    &
           upfreq=1, ivar=ivar, initialize=initialize_variables, index = ih_fire_c_to_atm_si )
    
+    call this%set_history_var(vname='FIRE_FLUX', units='g/m^2/s', &
+          long='ED/SPitfire loss to atmosphere of elements', use_default='active', &
```
jkshuman, line 5303:
```suggestion
          long='ED-spitfire loss to atmosphere of elements', use_default='active', &
```

rgknox:
I like this PR, good stuff.

One thing to note.  When you use flushval = 0.0, like here:

https://github.com/NGEET/fates/pull/640/files#diff-682b3e29ce096fa74f8fa95ff48735c4R4122

You are zero'ing literally every column, and will potentially include some zeros from non-fates columns.  In the next refactor of the history variables, we will automatically flush non-fates columns to the ignore value.  If you want to be assured that you are not including zero's from non-fates columns in your average, flush to the ignore value, and make sure to pre-zero any column level diagnostics that are tallied up over patches and cohorts.

glemieux:
All expected PASS:
```
/glade/u/home/glemieux/scratch/clmed-tests/pr640-fire-hist-vars-C242d6d14-F82cb0e45.fates.cheyenne.intel
/glade/u/home/glemieux/scratch/clmed-tests/pr640-fire-hist-vars-C242d6d14-F82cb0e45.fates.cheyenne.gnu
```
Confirmed all DIFFs due to changed fire variables.

PR #649, opened 2020-05-11
Added some metadata for Zenodo

ckoven:
Added a file, `.zenodo.json`, which provides metadata for zenodo to autopopulate fields in archive, in order to create a citable doi for the FATES code itself (which we will do shortly).  After some discussion, we've decided to—for now anyway—use a group authorship, "FATES Development Team", for the code rather than a long authorship list of individual contributors.  If there is sufficient demand to change this to a list of real names instead, we can absolutely do that, we'd just need to first agree on and write out some policy for generating such a list.  So leaving that as a future decision for now.

Based on discussions with @rgknox @glemieux @rosiealice, as well as some cross-discussion with/ looking at what some related models have done (https://e3sm.org/resources/policies/acknowledge-e3sm/) or are doing (https://github.com/ESCOMP/CTSM/issues/1005) on the software citation front.

Not tested since no functional code was touched.


glemieux:
Thanks Charlie.

I will create a release for this pretty soon.  @rgknox and I are discussing how to tag this since github releases are based off of git tags (and our tagging format doesn't account for updates like this).

Issue #616, opened 2020-02-25
New parameters for next parameter file update


JessicaNeedham:
I have a new parameter that needs to go into the next update to the parameter file. 
This parameter is a switch that will turn cohort age tracking on (1) or off (0). 

https://github.com/JessicaNeedham/fates/blob/dd28a9b10476e56126976bc6a8074ec88bd7aa06/parameter_files/fates_params_default.cdl#L532-L534

https://github.com/JessicaNeedham/fates/blob/dd28a9b10476e56126976bc6a8074ec88bd7aa06/parameter_files/fates_params_default.cdl#L1232

The only other change is that the default number of cohort age bin edges needs to be 2 and not 1. When age tracking is off, cohorts will be assigned age 0 and it will not be incremented. But they still need to go into the first(only) age bin to stop crashes.

https://github.com/JessicaNeedham/fates/blob/dd28a9b10476e56126976bc6a8074ec88bd7aa06/parameter_files/fates_params_default.cdl#L683

Other parameters associated with cohort age tracking and size- and age-dependent mortality are already in the default parameter file.

rgknox:
Thanks @JessicaNeedham , folks if we can come up with other parameters we want to add along with this one, feel free to describe them here

jenniferholm:
Cool! Here is a list of additional parameters that are needed in FatesPlantRespPhotosynthMod so that the availability and competition for nutrients has an affect on photosynthesis. 

These parameters should go into the parameter file ==
'fates_leaf_flnr' = fraction of leaf N in Rubisco by PFT.
These are used in the Walker et al. 2014 equations for N and P limitation on photosynthesis ("vcmax~np relationship coefficient") ==
vcmax_np1,  vcmax_np2, vcmax_np3, vcmax_np4 which vary by PFT.


https://github.com/jenniferholm/fates-1/commit/aba747c3915bbb1455b00574f05014b6a423f013

In my first go around I labeled these are starting with "fates_prt_" but then realized they should be labeled "fates_eca_" to match Ryan's naming convention. 

Global parameters that I hard coded in FatesPlantRespPhotosynthMod ==
jmax_np1, jmax_np2, jmax_np3 
! vcmax25 parameters from Bonan et al (2011) JGR, 116, doi:10.1029/2010JG001593
      fnr = 7.16_r8
      act25 = 3.6_r8 !umol/mgRubisco/min

alistairrogers:
Hi Jennifer, 

[...]

Issue #290, opened 2017-10-13
Potential inconsistency in root profile parameterization


rosiealice:
Looking through the parameter list, I noted there are two parameter types for defining the root profile. 

One of them (roota_par and rootb_par) is used to define the root profile for water stress (patch%rootfr_ft) at the very bottom of https://github.com/NGEET/fates/blob/master/biogeochem/EDPatchDynamicsMod.F90

The other (rootprof_beta) is used to calculate the input of root litter into the different soil layers in https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90 on line 1552.

My vote would be to only use one of these, and ideally the simpler Jackson term (rootprof_beta), but which one is less important than being consistent. Is there a reason that we are preserving both that I don't know about?  CLM moved from the 2-parameter to 1-parameter model in the last couple of years, which is probably how this arose.

rgknox:
From what I am following in this thread, it seems there is no argument for keeping two different parameter sets correct?

I am in the process of cleaning up some parameters right now and would like some resolution here.

I propose the following as a the definitive way to define the fine root profile.

```
  real(r8), allocatable :: rootprof_mode              ! Fine root profile functional form
  real(r8), allocatable :: rootprof_a(:)              ! Fine root profile scaling parameter A
  real(r8), allocatable :: rootprof_b(:)              ! Fine root profile scaling parameter B
```

Note that the rooprof_mode will allow us to choose between a generic 1 parameter exponential, 2 parameter exponential, Jackson's method and Zeng's method.  This would also allow us add alternative functions that normalize these profiles to a maximum rooting depth that is based on plant size stature.

Currently, we do use the rootprof beta for things like biomass turnover and respiration, but we use the roota and rootb parameters for all things hydraulics.  I suppose we would need to choose a default (Zeng? Jackson?)

dlawrenncar:
I don't really understand all the uses of rooting distribution across CLM
and FATES, but during CLM5 development, we made an effort to remove
inconsistencies in rooting distribution across different components of the
model (water was using Zeng, carbon using Jackson).  It seems like this
might be a situation where you might want to consider having the control on
rooting distribution to be maintained within the host land model and we
could agree on a default rooting distribution when running CLM(FATES) or
ELM(FATES).

Perhaps at the same time, we should consider what would happen if we moved
to a dynamic rooting distribution, which is partially functional in CLM and
maybe more fully functional in ELM.

rosiealice:
[...]

PR #659, opened 2020-05-27
parameter file updates

rgknox:
### Description:

This is a group of changes to the parameter file, not backwards compatible.

### Collaborators:

Many, see : #616

Implementing non-answer changing modificaitons first, will test incrementally

### Checklist:

- [O] FATES PASS/FAIL regression tests were run
- [O] If answers were expected to change, evaluation was performed and provided

### Test Results:

Tested addition of Medlyn stomatal parameters, matches baseline, moving on to next set

/glade/scratch/rgknox/clmed-tests/fates-sci.1.36.0_api.11.1.0-clm5.0.30-v1-Cd525994c-Fed442bb3.fates.cheyenne.gnu

### DIFF
--- a/biogeophys/FatesPlantRespPhotosynthMod.F90
+++ b/biogeophys/FatesPlantRespPhotosynthMod.F90
@@ -239,10 +239,6 @@ contains
     ! Bonan et al (2011) JGR, 116, doi:10.1029/2010JG001593
     ! -----------------------------------------------------------------------------------
 
-    ! Ball-Berry minimum leaf conductance, unstressed (umol H2O/m**2/s)
-    ! For C3 and C4 plants
-    ! -----------------------------------------------------------------------------------
-
     associate(  &
          stomatal_intercept   => EDPftvarcon_inst%stomatal_intercept, & 
          c3psn     => EDPftvarcon_inst%c3psn  , &
--- a/biogeophys/FatesPlantRespPhotosynthMod.F90
+++ b/biogeophys/FatesPlantRespPhotosynthMod.F90
@@ -102,7 +102,6 @@ contains
     use FatesAllometryMod, only : bleaf
     use FatesAllometryMod, only : storage_fraction_of_target
     use FatesAllometryMod, only : set_root_fraction
-    use FatesAllometryMod, only : i_hydro_rootprof_context
     use FatesAllometryMod, only : decay_coeff_kn
 
     ! ARGUMENTS:
@@ -267,7 +266,7 @@ contains
 
          do ft = 1,numpft
              call set_root_fraction(rootfr_ft(ft,:), ft, &
-                   bc_in(s)%zi_sisl,icontext = i_hydro_rootprof_context)
+                   bc_in(s)%zi_sisl)
          end do
           
 
--- a/biogeophys/FatesPlantRespPhotosynthMod.F90
+++ b/biogeophys/FatesPlantRespPhotosynthMod.F90
@@ -46,7 +46,7 @@ module FATESPlantRespPhotosynthMod
    use PRTGenericMod,     only : store_organ
    use PRTGenericMod,     only : repro_organ
    use PRTGenericMod,     only : struct_organ
-   use EDParamsMod, only : ED_val_bbopt_c3, ED_val_bbopt_c4, ED_val_base_mr_20
+   use EDParamsMod,       only : ED_val_base_mr_20
 
    ! CIME Globals
    use shr_log_mod , only      : errMsg => shr_log_errMsg
@@ -161,8 +161,8 @@ contains
     real(r8) :: mm_kco2            ! Michaelis-Menten constant for CO2 (Pa)
     real(r8) :: mm_ko2             ! Michaelis-Menten constant for O2 (Pa)
     real(r8) :: co2_cpoint         ! CO2 compensation point (Pa)
-    real(r8) :: btran_eff          ! effective transpiration wetness factor (0 to 1) 
-    real(r8) :: bbb                ! Ball-Berry minimum leaf conductance (umol H2O/m**2/s)
+    real(r8) :: btran_eff          ! effective transpiration wetness factor (0 to 1)
+    real(r8) :: stomatal_intercept_btran   ! water-stressed minimum stomatal conductance (umol H2O/m**2/s)
     real(r8) :: kn                 ! leaf nitrogen decay coefficient
     real(r8) :: cf                 ! s m**2/umol -> s/m (ideal gas conversion) [umol/m3]
     real(r8) :: gb_mol             ! leaf boundary layer conductance (molar form: [umol /m**2/s])
@@ -243,18 +243,14 @@ contains
     ! Ball-Berry minimum leaf conductance, unstressed (umol H2O/m**2/s)
     ! For C3 and C4 plants
     ! -----------------------------------------------------------------------------------
-    real(r8), dimension(0:1) :: bbbopt 
 
     associate(  &
+         stomatal_intercept   => EDPftvarcon_inst%stomatal_intercept, & 
          c3psn     => EDPftvarcon_inst%c3psn  , &
          slatop    => EDPftvarcon_inst%slatop , & ! specific leaf area at top of canopy, 
                                                   ! projected area basis [m^2/gC]
          woody     => EDPftvarcon_inst%woody)     ! Is vegetation woody or not? 
 
-
-      bbbopt(0) = ED_val_bbopt_c4
-      bbbopt(1) = ED_val_bbopt_c3
-      
       do s = 1,nsites
 
          ! Multi-layer parameters scaled by leaf nitrogen profile.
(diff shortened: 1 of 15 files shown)


rgknox:
Added rooting profile updates, per #290 
These changes make it so that the same rooting profile is used for both btran and decomposition, but also is a non-b4b change.  Results should be nearly identical though for all non-decomposition related output.  Test results at BCI:  
[bci-test-ufnrt_plots.pdf](https://github.com/NGEET/fates/files/4699785/bci-test-ufnrt_plots.pdf)

rgknox:
Compared commit after adding in the fire-threshold and dbhmin parameters with commit before it, results are b4b:
/glade/scratch/rgknox/clmed-tests/fates-sci.1.36.0_api.11.1.0-clm5.0.30-v3-Cd525994c-F8a1317a9.fates.cheyenne.gnu

rgknox:
paired with:  https://github.com/ESCOMP/CTSM/pull/1027

ckoven:
do we want to pull in @rosiealice's PFT matrix term into this?https://github.com/rosiealice/fates/commit/f0170fd2834a1db26d83df0b9d0e3b0c9e3adcb5
and https://github.com/rosiealice/fates/commit/a69ac89051a57b55f453d079a01550f3ea661d76

fire/SFMainMod.F90
```diff
@@ -774,7 +774,7 @@ subroutine  area_burnt_intensity ( currentSite )
          endif
 
          !'decide_fire' subroutine 
-         if (currentPatch%FI > fire_threshold) then !track fires greater than kW/m2 energy threshold
+         if (currentPatch%FI > SF_val_fire_threshold) then !track fires greater than kW/m2 energy threshold
```
jkshuman, line 777:
```suggestion
         if (currentPatch%FI > SF_val_fire_threshold) then !track fires greater than kW/m energy threshold
```

fire/SFMainMod.F90
```diff
@@ -668,6 +667,7 @@ subroutine  area_burnt_intensity ( currentSite )
     use FatesConstantsMod, only : years_per_day
     use SFParamsMod,       only : SF_val_fdi_alpha,SF_val_fuel_energy, &
          SF_val_max_durat, SF_val_durat_slope
```
jkshuman, line 669:
```suggestion
         SF_val_max_durat, SF_val_durat_slope, SF_val_fire_threshold
```

fire/SFMainMod.F90
```diff
@@ -668,6 +667,7 @@ subroutine  area_burnt_intensity ( currentSite )
...
+    use SFParamsMod, only  : SF_val_fire_threshold
```
jkshuman, line 670:
```suggestion
```

fire/SFParamsMod.F90
```diff
@@ -28,6 +28,7 @@ module SFParamsMod
    real(r8),protected, public :: SF_val_max_durat
    real(r8),protected, public :: SF_val_durat_slope
    real(r8),protected, public :: SF_val_drying_ratio
+   real(r8),protected, public :: SF_val_fire_threshold    ! threshold for fires that spread or go out. KW/m (Pyne 1986)
```
jkshuman, line 31:
```suggestion
   real(r8),protected, public :: SF_val_fire_threshold    ! threshold for fires that spread or go out. kW/m (Pyne 1996)
```

jkshuman: The default of 50 kW/m comes from SPITFIRE implementation in Thonicke et al 2010 per Pyne 1996, but maybe that does not need to go into the comments? More than likely we will not use that 50 kW/m value for all cases, so this reference may not make sense to keep.

parameter_files/fates_params_default.cdl
```diff
@@ -97,7 +97,7 @@ variables:
 	double fates_allom_hmode(fates_pft) ;
 		fates_allom_hmode:units = "index" ;
 		fates_allom_hmode:long_name = "height allometry function index." ;
-		fates_allom_hmode:possible_values = "1: O'Brien 1995; 2: Poorter 2006; 3: 2 parameter power law; 4: Chave 2014; 5: Martinez-Cano 2019." ;
+		fates_allom_hmode:possible_values = "1: O\'Brien 1995; 2: Poorter 2006; 3: 2 parameter power law; 4: Chave 2014; 5: Martinez-Cano 2019." ;
```
jkshuman, line 100:
```suggestion
		fates_allom_hmode:possible_values = "1: O'Brien 1995; 2: Poorter 2006; 3: 2 parameter power law; 4: Chave 2014; 5: Martinez-Cano 2019." ;
```

parameter_files/fates_params_default.cdl
```diff
@@ -591,6 +591,9 @@ variables:
 	double fates_fire_part_dens ;
 		fates_fire_part_dens:units = "kg/m2" ;
 		fates_fire_part_dens:long_name = "spitfire parameter, oven dry particle density, Table A1 Thonicke et al 2010" ;
+	double fates_fire_threshold ;
+		fates_fire_threshold:units = "kW/m" ;
+		fates_fire_threshold:long_name = "spitfire parameter, fire intensity threshold for tracking fires that spread" ;
```
jkshuman, line 596: this fire_threshold parameter gives weird behavior if it is set to zero. as a minimum I suggest 0.001, or something not quite zero. that comment will go into documentation, but should we add a suggested range here?

parameter_files/fates_params_default.cdl
```diff
@@ -951,6 +946,14 @@ data:
  fates_leaf_slatop = 0.012, 0.01, 0.024, 0.012, 0.03, 0.03, 0.012, 0.03, 
     0.03, 0.03, 0.03, 0.03 ;
 
+ fates_leaf_stomatal_intercept = 10000, 10000, 10000, 10000, 10000, 10000, 10000,
```
Qianyuxuan: We planned to assign the value of "1000" for all PFTs. Do you have any thoughts on that?

rgknox: @Qianyuxuan,  I was planning on modifying the default values when we integrate your pull request. In this pull request, we are keeping the numbers the same so the tests produce the same results.

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -243,18 +243,14 @@ real(r8) :: tcwood             ! Temperature response function for wood
     ! Ball-Berry minimum leaf conductance, unstressed (umol H2O/m**2/s)
```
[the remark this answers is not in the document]
rgknox: will do

parameter_files/fates_params_default.cdl
jkshuman: @rgknox including a zero value for fire_threshold here is not valid. Perhaps I should add a check in the code? thoughts on that.

parameter_files/fates_params_default.cdl
Qianyuxuan: @rgknox Okay! No more questions. Thanks.

rgknox: thanks for the review!

parameter_files/fates_params_default.cdl
rgknox: I think that sounds sensible.  I can put in a check that the number is greater than 0.001.
Also, this parameter is used in all contexts/scenarios when spitfire is turned on right? It is not used in a subset of spitifre that is not-necessarily active right?

jkshuman: @rgknox you are correct. the fire_threshold is used in all scenarios where spitfire is on.

jkshuman: @rgknox and thank you for adding a check. that will be a good fail safe.

parameter_files/fates_params_default.cdl
serbinsh: Great, thanks @Qianyuxuan

rgknox:
@serbinsh yup, exactly.   

I'm wondering what extra tests, if any, we should perform before switching to Medlyn as the default model in the parameter file.  I feel it might be a good idea to run a global simulation with it for a few decades just to see if anything weird happens (like an f10_f10 grid).  Since the default parameter set (right now) is more about stress testing the model in different modes, and less about providing scientifically reasonable base set, the bar is low in terms of producing reasonable global benchmarks.  Although, arriving at a default parameter set that does the latter is the long-term goal, and adding Medlyn as the default their may help get us to that goal.

serbinsh:
@rgknox Both @Qianyuxuan and I would be happy to talk through a baseline run.  We can even do that run here if helpful.  But before we do that we should perhaps think about exactly what we need.

parameter_files/fates_params_default.cdl
rgknox: @jkshuman I'm just going to remove the apostrophe in Obrien's name.  I feel like there is a back and forth on the preference here, and I'm not sure how netcdf interprets that character

jkshuman: @rgknox no worries. I had thought that \ character was a mistake. happy to go with your choice on this.

fire/SFMainMod.F90
rgknox: github's resolve feature did not get this fix correct, will apply manually

parameter_files/fates_params_default.cdl
rgknox: ok, sounds good

rgknox:
Test updates look nominal:
/glade/scratch/rgknox/clmed-tests/fates-sci.1.36.0_api.11.1.0-clm5.0.30-Cdab82c29-Feaa02cf4.fates.cheyenne.gnu/

I did the cheyenne gnu test suite, which is nice for small syntax updates like these (two tests), and baseline comparisons were b4b.

I propose we freeze changes on this PR.  In your hands now @glemieux

Issue #660, opened 2020-06-02
current code not conserving primary/second land areas


ckoven:
the current code isn't conserving primary and secondary land areas, for reasons I believe having to do with the forced fusion of very small patches.  I encountered this in a transient historical run with land use, but since that is a new capability which is uncovering a bug that exists in the prior code, I wanted to make an issue in case anyone else is running with secondary forest dynamics in the old harvest logic.  

i made a change to the new patch insertion logic that seems to fix the bug, at least in the case where I encountered it.  basically, rather than adding newly-disturbed patches to the youngest side of the patch linked list irrespective of their land-use label, which results in interleaved primary and secondary patches through the list, I changed it to add the newly-disturbed primary patch after all of the secondary patches, if there are any.  This results in contiguous primary and secondary patches in the linked list, which then avoids the force-merge bug.  I also have added a logical check to fail if two patches with different labels are being fused, in case there are other edge cases that I haven't run into yet.

@rosiealice I'm not sure if this logic might apply to your pft labeling case in the prescribed biogeography mode with disturbance too?

proposed fix here: https://github.com/ckoven/fates/commit/0b6ab98e9fff54dff5695385f3279038b2371425

rosiealice:
Interesting. I haven't been using the secondary forest capabilities at all yet in the NOCOMP mode, but I did have a bunch of problems with the forced fusion of v v small patches in terminate cohorts (I was triggering the 'you only have one patch like this and it is tiny' check). 

To that end, as i mentioned earlier, I remove the very small patches  (<1%) in the initialization and distribute their area to the other present PFTs ( in fixed biogeog mode).

Having said all of that, I do still have a memory leak or something when I try and turn on my implementation of patch dynamics with no competition.  I am essentially looping round  the PFTs and creating one primary and one secondary patch, starting here:
https://github.com/rosiealice/fates/blob/ea3c635d00e7b3863b69d75ffb555c7617eb8f93/biogeochem/EDPatchDynamicsMod.F90#L489

and ending here
https://github.com/rosiealice/fates/blob/ea3c635d00e7b3863b69d75ffb555c7617eb8f93/biogeochem/EDPatchDynamicsMod.F90#L1093

and so we end up with patches that aren't able to fuse with each other interleaved all over the place!  Obviously they need to not fuse together, as per:
[...]

Issue #491, opened 2019-02-25
define land use drivers for FATES LULCC-v1


ckoven:
In order to keep things moving on the FATES LULCC-v1 (#450), I think a next step should be to define the exact variables that FATES would use to drive the primary to secondary forest transitions in the model.  So I'd like to start discussion on that, and once we agree on exactly what info FATES needs, we can then work outwards from there on putting the LULCC information flow into an interface and having ELM and CLM pass it the info that it needs.

Looking at the CMIP6 drivers, it seems like the relevant fields that we'd want to use initially are the harvest area fields.  So, e.g. looking at either the historical transitions file `multiple-transitions_input4MIPs_landState_CMIP_UofMD-landState-2-1-h_gn_0850-2015.nc`, or one of the future scenario transitions files, e.g. `multiple-transitions_input4MIPs_landState_ScenarioMIP_UofMD-MAGPIE-ssp534-2-1-f_gn_2015-2100.nc`, as downloaded from the input4mips website (https://esgf-node.llnl.gov/search/input4mips/) I see the following fields:

```
	float primn_harv(time, lat, lon) ;
		primn_harv:_FillValue = 1.e+20f ;
		primn_harv:missing_value = 1.e+20f ;
		primn_harv:long_name = "wood harvest area from primary non forest land" ;
		primn_harv:standard_name = "area_fraction" ;
		primn_harv:cell_methods = "time:mean" ;
		primn_harv:units = "1" ;
	float secmf_harv(time, lat, lon) ;
		secmf_harv:_FillValue = 1.e+20f ;
		secmf_harv:missing_value = 1.e+20f ;
		secmf_harv:long_name = "wood harvest area from secondary mature forest land" ;
		secmf_harv:standard_name = "area_fraction" ;
		secmf_harv:cell_methods = "time:mean" ;
		secmf_harv:units = "1" ;
	float secyf_harv(time, lat, lon) ;
		secyf_harv:_FillValue = 1.e+20f ;
		secyf_harv:missing_value = 1.e+20f ;
		secyf_harv:long_name = "wood harvest area from secondary young forest land " ;
		secyf_harv:standard_name = "area_fraction" ;
		secyf_harv:cell_methods = "time:mean" ;
		secyf_harv:units = "1" ;
	float primf_harv(time, lat, lon) ;
		primf_harv:_FillValue = 1.e+20f ;
		primf_harv:missing_value = 1.e+20f ;
[...]

Issue #276, opened 2017-09-22
Long-term/mid-term feature wishlist for the logging module


rgknox:
The version 1 of the logging module is nigh upon us.  

I wanted to summarize some of the wishlist items people have brought up in discussion, and solicit input on other logging features that may be useful in further implementation.

glemieux:
This was fixed in #663

PR #663, opened 2020-06-13
Use HLM-provided land-use time series data to drive FATES harvest

ckoven:
These changes provide a first version of land use change, as driven by the CLM & ELM landuse.timeseries files that currently drive the big-leaf versions of the models, and should allow global transient simulations with land use.

The idea is that the HLMs read a series of land-use driving datasets (as they already do), which they'll now then pass through the HLM-FATES interface via the interface variable `bc_in%hlm_harvest(num_lu_harvest_cats)`.  Currently the default is that the HLM reads and passes five area-based logging streams, for harvest from various categories of land (primary forest, primary nonforest, secondary young forest, secondary young nonforest, secondary old forest).  FATES uses this data to harvest from three separate types: primary lands, secondary young lands, and secondary old lands.  FATES does not distinguish between forest and no-forest lands, though possibly a future step could be to infer that either from the composition of a given patch or as specified if prescribed biogeography is turned on.  Mass-based harvest and associated disturbance is not yet implemented here, that is also a next step.

The harvest parameters are kept from before, though their usage and meaning are slightly different.
The event code triggers when within the year to apply the harvest rates, the direct_frac determines what fraction of trees within the harvested area are removed (so `fates_logging_direct_frac=1` means clearcut), etc.  Plants within the logged area that are not harvested, whether because they are not trees, are not within the dbhmin threshold, or because `(fates_logging_direct_frac + fates_logging_mechanical_frac  + fates_logging_collateral_frac ) < 1` are all moved to a newly-disturbed secondary patch, along with an equivalent fraction of bare ground of each harvested patch if the canopy is not closed.  So the definition of secondary lands includes everything from full clearcut disturbance to potentially relatively low-intensity forest degradation.  But for consistency with the intention of the land-use datasets, we should probably set the default values to be closer to clear-cut type of conditions.  Future steps may include making these arguments harvest-type-dependent. Also we still need to wire the `max_dbh` parameter introduced in #659 to this logic, once that code is merged into this.

We added a bunch of diagnostics to serve as either sanity checks or other trackers of things that might be useful, including tracking various disturbance rates, as well as 'potential disturbace' rates (i.e. the disturbance rates before the part of the code where it says only resolve a given patch's most dominant disturbance type on a given day).  Also changed the newly-disturbed patch insertion to fix a bug as described in #660.

This isn't really ready to be merged just yet, in particular the merging with the prescribed biogeography (#632 and follow-on) will most likely be a bit tricky.  But putting up for interested parties to start reviewing.

fixes: #660, #491, #276, and the first batch of changes described in #450.

### Collaborators:
co-written by @aldivi and me.  @aldivi also wrote the E3SM-side changes of these (PR to come later) and I wrote the CTSM-side changes (PR https://github.com/ESCOMP/CTSM/pull/1040).

This should only change answers when logging is operational.  But this hasn't been formally tested yet and probably requires some new tests that use the land-use timeseries data.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -58,7 +58,7 @@ module EDLoggingMortalityMod
    use PRTGenericMod     , only : fnrt_organ, store_organ, repro_organ
    use FatesAllometryMod , only : set_root_fraction
    use FatesAllometryMod , only : i_biomass_rootprof_context
-  use FatesConstantsMod    , only : primaryforest, secondaryforest, secondary_age_threshold
+   use FatesConstantsMod , only : primaryforest, secondaryforest, secondary_age_threshold
 
    implicit none
    private
@@ -84,6 +84,8 @@ module EDLoggingMortalityMod
    public :: logging_litter_fluxes
    public :: logging_time
    public :: IsItLoggingTime
+   integer, parameter, public :: hlm_harvest_area_fraction = 1
+   integer, parameter, public :: hlm_harvest_carbon = 2
 
 contains
 
@@ -199,13 +201,14 @@ contains
                                                 ! are moved to newly-anthro-disturbed secondary
                                                 ! forest patch)
 
-      ! Parameters
+      ! Local variables
       integer :: icode   ! Integer equivalent of the event code (parameter file only allows reals)
-      real(r89) :: harvest_rate ! the final harvest rate to apply to this cohort today
+      real(r8) :: harvest_rate ! the final harvest rate to apply to this cohort today
+      integer :: h_index   ! for looping over harvest categories
+
+      ! Parameters
       real(r8), parameter   :: adjustment = 1.0 ! adjustment for mortality rates
       real(r8), parameter :: months_per_year = 12.0
-
-      icode = int(logging_event_code)
 
       if (logging_time) then 
          if(EDPftvarcon_inst%woody(pft_i) == 1)then ! only set logging rates for trees
@@ -213,17 +216,18 @@ contains
 ! todo: add a logging_dbhmax parameter, and probably lower the dbhmin one to 30 cm
 ! todo: change the default logging_event_code to 1 september (-244)
 ! todo: change the default logging_direct_frac to 0.7, which is closer to a clearcut event
+! todo: check secondary forest creation
 ! todo: check outputs against the LUH2 carbon data
 ! todo: eventually set up distinct harvest practices, each with a set of input paramaeters
 
             ! Pass logging rates to cohort level 
 
             if (hlm_use_lu_harvest == 0) then
-               ! 0=use fates logging when logging_time == .true.
-               ! this means harvest the whole cohort
+               ! 0=use fates logging parameters directly when logging_time == .true.
+               ! this means harvest the whole cohort area
                harvest_rate = 1._r8
 
-            else if (hlm_use_lu_harvest == 1) then
+            else if (hlm_use_lu_harvest == hlm_harvest_area_fraction) then
                ! 1=use area fraction from hlm
                ! combine forest and non-forest fracs and then apply:
                ! primary and secondary area fractions to the logging rates, which are fates parameters
@@ -242,29 +246,28 @@ contains
                do h_index = 1,hlm_num_lu_harvest_cats
                   if (use_history .eq. primaryforest) then
                      if(hlm_harvest_catnames(h_index) .eq. "HARVEST_VH1" .or. &
-                        hlm_harvest_catnames(h_index) .eq. "HARVEST_VH2")
+                        hlm_harvest_catnames(h_index) .eq. "HARVEST_VH2") then
                         harvest_rate = harvest_rate + hlm_harvest(h_index)
                      endif
-                  else if (use_history .eq. secondaryforest .and.
+                  else if (use_history .eq. secondaryforest .and. &
                      secondary_age >= secondary_age_threshold) then
-                     if(hlm_harvest_catnames(h_index) .eq. "HARVEST_SH1")
+                     if(hlm_harvest_catnames(h_index) .eq. "HARVEST_SH1") then
                         harvest_rate = harvest_rate + hlm_harvest(h_index)
                      endif
-                  else if (use_history .eq. secondaryforest .and.
+                  else if (use_history .eq. secondaryforest .and. &
                      secondary_age < secondary_age_threshold) then
                      if(hlm_harvest_catnames(h_index) .eq. "HARVEST_SH2" .or. &
-                        hlm_harvest_catnames(h_index) .eq. "HARVEST_SH3")
+                        hlm_harvest_catnames(h_index) .eq. "HARVEST_SH3") then
                         harvest_rate = harvest_rate + hlm_harvest(h_index)
                      endif
                   endif
                end do
 
                ! calculate today's harvest rate
-               ! the timing has already been determined by IsItLoggingTime
+               ! whether to harvest today has already been determined by IsItLoggingTime
                ! for icode == 2, icode < 0, and icode > 10000 apply the annual rate one time (no calc)
                ! Bad logging event flag is caught in IsItLoggingTime, so don't check it here
                icode = int(logging_event_code)
-               !harvest_rate = harvest_rate / hlm_days_per_year
                if(icode .eq. 1) then
                   ! Logging is turned off - not sure why we need another switch
                   harvest_rate = 0._r8
@@ -275,9 +278,10 @@ contains
                   ! logging event once a month
                   if(hlm_current_day.eq.1  ) then
                      harvest_rate = harvest_rate / months_per_year
+                  end if
                end if
 
-            else if (hlm_use_lu_harvest == 2) then
+            else if (hlm_use_lu_harvest == hlm_harvest_carbon) then
                ! 2=use carbon from hlm
                ! not implemented yet
                write(fates_log(),*) 'HLM harvest carbon data not implemented yet. Exiting.'
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -187,8 +187,8 @@ contains
       integer,  intent(in)  :: pft_i            ! pft index 
       real(r8), intent(in)  :: dbh              ! diameter at breast height (cm)
       integer,  intent(in)  :: canopy_layer     ! canopy layer of this cohort
-      real(r8), intent(in) :: hlm_harvest       ! annual harvest rate per hlm category
-      character(len=64), intent(in) :: hlm_harvest_catnames ! names of hlm harvest categories
+      real(r8), intent(in) :: hlm_harvest(:)       ! annual harvest rate per hlm category
+      character(len=64), intent(in) :: hlm_harvest_catnames(:) ! names of hlm harvest categories
       integer, intent(in) :: use_history        ! patch level anthro_disturbance_label
       real(r8), intent(in) :: secondary_age     ! patch level age_since_anthro_disturbance
       real(r8), intent(out) :: lmort_direct     ! direct (harvestable) mortality fraction
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -45,7 +45,7 @@ module EDLoggingMortalityMod
    use FatesInterfaceMod , only : hlm_day_of_year 
    use FatesInterfaceMod , only : hlm_days_per_year
    use FatesInterfaceMod , only : hlm_use_lu_harvest
-   use FatesInterfaceMod , only : hlm_use_num_lu_harvest_cats
+   use FatesInterfaceMod , only : hlm_num_lu_harvest_cats
    use FatesInterfaceMod , only : hlm_use_logging 
    use FatesInterfaceMod , only : hlm_use_planthydro
    use FatesConstantsMod , only : itrue,ifalse
@@ -239,7 +239,7 @@ contains
 
                ! determine the annual hlm harvest rate for the current cohort based on patch info
                harvest_rate = 0._r8
-               do h_index = 1,hlm_use_num_lu_harvest_cats
+               do h_index = 1,hlm_num_lu_harvest_cats
                   if (use_history .eq. primaryforest) then
                      if(hlm_harvest_catnames(h_index) .eq. "HARVEST_VH1" .or. &
                         hlm_harvest_catnames(h_index) .eq. "HARVEST_VH2")
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -289,8 +289,7 @@ contains
 
          ! the area occupied by all plants in the canopy that aren't killed is still disturbed at the harvest rate
          if (canopy_layer .eq. 1) then
-            l_degrad = harvest_rate * (1._r8 - &
-                 (logging_direct_frac + logging_collateral_frac + logging_mechanical_frac)) ! fraction passed to 'degraded' forest.
+            l_degrad = harvest_rate - (lmort_direct + lmort_infra + lmort_collateral) ! fraction passed to 'degraded' forest.
          else
             l_degrad = 0._r8
          endif
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -61,6 +61,8 @@ module EDLoggingMortalityMod
    use FatesConstantsMod , only : primaryforest, secondaryforest, secondary_age_threshold
    use FatesConstantsMod , only : fates_tiny
    use FatesConstantsMod , only : months_per_year
+   use FatesConstantsMod , only : hlm_harvest_area_fraction
+   use FatesConstantsMod , only : hlm_harvest_carbon
    use FatesConstantsMod, only : fates_check_param_set
 
    implicit none
@@ -88,8 +90,6 @@ module EDLoggingMortalityMod
    public :: logging_time
    public :: IsItLoggingTime
    public :: get_harvest_rate_area
-   integer, parameter, public :: hlm_harvest_area_fraction = 1 ! Code for harvesting by area
-   integer, parameter, public :: hlm_harvest_carbon = 2 ! Code for harvesting based on carbon extracted. 
 
 contains
 
@@ -187,6 +187,7 @@ contains
    subroutine LoggingMortality_frac( pft_i, dbh, canopy_layer, lmort_direct, &
                                      lmort_collateral,lmort_infra, l_degrad, &
                                      hlm_harvest_rates, hlm_harvest_catnames, &
+                                     hlm_harvest_units, &
                                      patch_anthro_disturbance_label, secondary_age, &
                                      frac_site_primary)
 
@@ -196,6 +197,7 @@ contains
       integer,  intent(in)  :: canopy_layer     ! canopy layer of this cohort
       real(r8), intent(in) :: hlm_harvest_rates(:) ! annual harvest rate per hlm category
       character(len=64), intent(in) :: hlm_harvest_catnames(:) ! names of hlm harvest categories
+      integer, intent(in) :: hlm_harvest_units     ! unit type of hlm harvest rates: [area vs. mass]
       integer, intent(in) :: patch_anthro_disturbance_label    ! patch level anthro_disturbance_label
       real(r8), intent(in) :: secondary_age     ! patch level age_since_anthro_disturbance
       real(r8), intent(out) :: lmort_direct     ! direct (harvestable) mortality fraction
@@ -220,12 +222,12 @@ contains
       if (logging_time) then 
          ! Pass logging rates to cohort level 
          
-         if (hlm_use_lu_harvest == 0) then
+         if (hlm_use_lu_harvest == ifalse) then
             ! 0=use fates logging parameters directly when logging_time == .true.
             ! this means harvest the whole cohort area
             harvest_rate = 1._r8
             
-         else if (hlm_use_lu_harvest == hlm_harvest_area_fraction) then
+         else if (hlm_use_lu_harvest == itrue .and. hlm_harvest_units == hlm_harvest_area_fraction) then
             ! We are harvesting based on areal fraction, not carbon/biomass terms. 
             ! 1=use area fraction from hlm
             ! combine forest and non-forest fracs and then apply:
@@ -239,11 +241,12 @@ contains
             ! HARVEST_SH1 = harvest from secondary mature forest
             ! HARVEST_SH2 = harvest from secondary young forest
             ! HARVEST_SH3 = harvest from secondary non-forest (assume this is young for biomass)
-      ! Get the area-based harvest rates based on info passed to FATES from the bioundary condition
+
+            ! Get the area-based harvest rates based on info passed to FATES from the boundary condition
             call get_harvest_rate_area (patch_anthro_disturbance_label, hlm_harvest_catnames, &
                  hlm_harvest_rates, frac_site_primary, secondary_age, harvest_rate)
 
-         else if (hlm_use_lu_harvest == hlm_harvest_carbon) then
+         else if (hlm_use_lu_harvest == itrue .and. hlm_harvest_units == hlm_harvest_carbon) then
             ! 2=use carbon from hlm
             ! not implemented yet
             write(fates_log(),*) 'HLM harvest carbon data not implemented yet. Exiting.'
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -60,6 +60,7 @@ module EDLoggingMortalityMod
    use FatesAllometryMod , only : set_root_fraction
    use FatesConstantsMod , only : primaryforest, secondaryforest, secondary_age_threshold
    use FatesConstantsMod , only : fates_tiny
+   use FatesConstantsMod, only : fates_check_param_set
 
    implicit none
    private
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -60,6 +60,7 @@ module EDLoggingMortalityMod
    use FatesAllometryMod , only : set_root_fraction
    use FatesConstantsMod , only : primaryforest, secondaryforest, secondary_age_threshold
    use FatesConstantsMod , only : fates_tiny
+   use FatesConstantsMod , only : months_per_year
    use FatesConstantsMod, only : fates_check_param_set
 
    implicit none
@@ -185,17 +186,17 @@ contains
 
    subroutine LoggingMortality_frac( pft_i, dbh, canopy_layer, lmort_direct, &
                                      lmort_collateral,lmort_infra, l_degrad, &
-                                     hlm_harvest, hlm_harvest_catnames, &
-                                     use_history, secondary_age, &
+                                     hlm_harvest_rates, hlm_harvest_catnames, &
+                                     patch_anthro_disturbance_label, secondary_age, &
                                      frac_site_primary)
 
       ! Arguments
       integer,  intent(in)  :: pft_i            ! pft index 
       real(r8), intent(in)  :: dbh              ! diameter at breast height (cm)
       integer,  intent(in)  :: canopy_layer     ! canopy layer of this cohort
-      real(r8), intent(in) :: hlm_harvest(:)       ! annual harvest rate per hlm category
+      real(r8), intent(in) :: hlm_harvest_rates(:) ! annual harvest rate per hlm category
       character(len=64), intent(in) :: hlm_harvest_catnames(:) ! names of hlm harvest categories
-      integer, intent(in) :: use_history        ! patch level anthro_disturbance_label
+      integer, intent(in) :: patch_anthro_disturbance_label    ! patch level anthro_disturbance_label
       real(r8), intent(in) :: secondary_age     ! patch level age_since_anthro_disturbance
       real(r8), intent(out) :: lmort_direct     ! direct (harvestable) mortality fraction
       real(r8), intent(out) :: lmort_collateral ! collateral damage mortality fraction
@@ -239,8 +240,8 @@ contains
             ! HARVEST_SH2 = harvest from secondary young forest
             ! HARVEST_SH3 = harvest from secondary non-forest (assume this is young for biomass)
       ! Get the area-based harvest rates based on info passed to FATES from the bioundary condition
-            call get_harvest_rate_area (use_history, hlm_harvest_catnames, hlm_harvest, &
-                 frac_site_primary, secondary_age, harvest_rate)
+            call get_harvest_rate_area (patch_anthro_disturbance_label, hlm_harvest_catnames, &
+                 hlm_harvest_rates, frac_site_primary, secondary_age, harvest_rate)
 
          else if (hlm_use_lu_harvest == hlm_harvest_carbon) then
             ! 2=use carbon from hlm
@@ -299,7 +300,7 @@ contains
 
    ! ============================================================================
 
-   subroutine get_harvest_rate_area (use_history, hlm_harvest_catnames, hlm_harvest, &
+   subroutine get_harvest_rate_area (patch_anthro_disturbance_label, hlm_harvest_catnames, hlm_harvest_rates, &
                  frac_site_primary, secondary_age, harvest_rate)
 
 
@@ -310,9 +311,9 @@ contains
      !  assumes logging_time == true
 
       ! Arguments
-      real(r8), intent(in) :: hlm_harvest(:)       ! annual harvest rate per hlm category
+      real(r8), intent(in) :: hlm_harvest_rates(:) ! annual harvest rate per hlm category
       character(len=64), intent(in) :: hlm_harvest_catnames(:) ! names of hlm harvest categories
-      integer, intent(in) :: use_history        ! patch level anthro_disturbance_label
+      integer, intent(in) :: patch_anthro_disturbance_label    ! patch level anthro_disturbance_label
       real(r8), intent(in) :: secondary_age     ! patch level age_since_anthro_disturbance
       real(r8), intent(in) :: frac_site_primary
       real(r8), intent(out) :: harvest_rate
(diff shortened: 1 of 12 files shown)


biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -104,10 +113,17 @@ subroutine IsItLoggingTime(is_master,currentSite)
       logging_time = .false.
       icode = int(logging_event_code)
 
+      ! this is true for either hlm harvest or fates logging
```
rosiealice, line 116: How is it true for both those cases? Is it set elsewhere?

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -104,10 +113,17 @@ subroutine IsItLoggingTime(is_master,currentSite)
...
       if(hlm_use_logging.eq.ifalse) return
 
+      ! all of these are valid for hlm harvest
+      ! so adjust annual harvest inputs accordingly in LoggingMortality_frac
+      ! note that the specific event will allow only one hlm harvest, regardless of input
+      ! code 3, every day, may not work properly because of the exclusive mortality selection
+      ! even less fequent low harvest rates may be excluded - may need to give harvest priority
+
```
rosiealice, line 124: Are they valid in the sense that logging happens in addition to the harvesting? (with the caveat that the model has to choose one of them?) Would it make sense to have a harvesting operation AND a logging operation happening in the same simulation?

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -168,41 +184,88 @@ end subroutine IsItLoggingTime
...
+            harvest_rate = 1._r8
+            
+         else if (hlm_use_lu_harvest == hlm_harvest_area_fraction) then
+            ! 1=use area fraction from hlm
+            ! combine forest and non-forest fracs and then apply:
+            ! primary and secondary area fractions to the logging rates, which are fates parameters
+            
+            ! Definitions of the underlying harvest land category variables
+            ! these are hardcoded to match the LUH input data via landuse.timseries file (see dynHarvestMod)
+            ! these are fraction of vegetated area harvested, split into five land category variables
```
rosiealice, line 234:
```suggestion
            ! these are fractions of vegetated area harvested, split into five land category variables
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -168,41 +184,88 @@ end subroutine IsItLoggingTime
...
       if (logging_time) then 
+         ! Pass logging rates to cohort level 
+         
+         if (hlm_use_lu_harvest == 0) then
+            ! 0=use fates logging parameters directly when logging_time == .true.
+            ! this means harvest the whole cohort area
+            harvest_rate = 1._r8
+            
+         else if (hlm_use_lu_harvest == hlm_harvest_area_fraction) then
+            ! 1=use area fraction from hlm
```
rosiealice, line 228:
```suggestion
            ! We are harvesting based on areal fraction, not carbon/biomass terms. 
            ! 1=use area fraction from hlm
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -80,6 +86,9 @@ module EDLoggingMortalityMod
    public :: logging_litter_fluxes
    public :: logging_time
    public :: IsItLoggingTime
+   public :: get_harvest_rate_area
+   integer, parameter, public :: hlm_harvest_area_fraction = 1
```
rosiealice, line 90:
```suggestion
   integer, parameter, public :: hlm_harvest_area_fraction = 1 ! Code for harvesting by area
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -80,6 +86,9 @@ module EDLoggingMortalityMod
...
+   integer, parameter, public :: hlm_harvest_carbon = 2
```
rosiealice, line 91:
```suggestion
   integer, parameter, public :: hlm_harvest_carbon = 2 ! Code for harvesting based on carbon extracted. 
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -168,41 +184,88 @@ end subroutine IsItLoggingTime
...
-            ! Pass logging rates to cohort level 
-
-            if (dbh >= logging_dbhmin ) then
-               lmort_direct = logging_direct_frac * adjustment
-               l_degrad = 0._r8
+            
+            if (dbh >= logging_dbhmin .and. .not. &
+                 ((logging_dbhmax < fates_check_param_set) .and. (dbh < logging_dbhmax )) ) then
+               lmort_direct = harvest_rate * logging_direct_frac
+               l_degrad = harvest_rate * (1._r8 - logging_direct_frac)
```
rosiealice, line 258:
```suggestion
               l_degrad = harvest_rate * (1._r8 - logging_direct_frac) ! fraction passed to 'degraded' forest. 
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -168,41 +184,88 @@ end subroutine IsItLoggingTime
...
+            
+            ! Definitions of the underlying harvest land category variables
+            ! these are hardcoded to match the LUH input data via landuse.timseries file (see dynHarvestMod)
+            ! these are fraction of vegetated area harvested, split into five land category variables
+            ! HARVEST_VH1 = harvest from primary forest
+            ! HARVEST_VH2 = harvest from primary non-forest
+            ! HARVEST_SH1 = harvest from secondary mature forest
+            ! HARVEST_SH2 = harvest from secondary young forest
+            ! HARVEST_SH3 = harvest from secondary non-forest (assume this is young for biomass)
+
```
rosiealice, line 240:
```suggestion
      ! Get the area-based harvest rates based on info passed to FATES from the bioundary condition
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -235,6 +298,93 @@ end subroutine LoggingMortality_frac
...
+      real(r8), intent(out) :: harvest_rate
+
+      ! Local Variables
+      integer :: h_index   ! for looping over harvest categories
+      integer :: icode   ! Integer equivalent of the event code (parameter file only allows reals)
+
+      ! Parameters
+      real(r8), parameter :: months_per_year = 12.0
+
+     ! determine the annual hlm harvest rate for the current cohort based on patch info
```
rosiealice, line 326:
```suggestion
     !  Loop around harvest categories to determine the annual hlm harvest rate for the current cohort based on patch history info
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -235,6 +298,93 @@ end subroutine LoggingMortality_frac
...
+        else if (use_history .eq. secondaryforest .and. &
+             secondary_age < secondary_age_threshold) then
+           if(hlm_harvest_catnames(h_index) .eq. "HARVEST_SH2" .or. &
+                hlm_harvest_catnames(h_index) .eq. "HARVEST_SH3") then
+              harvest_rate = harvest_rate + hlm_harvest(h_index)
+           endif
+        endif
+     end do
+
+     ! normalize by site-level fractio primary or secondary forest fraction
```
rosiealice, line 348:
```suggestion
     !  Normalize by site-level primary or secondary forest fraction
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -235,6 +298,93 @@ end subroutine LoggingMortality_frac
...
+     ! since harvest_rate is specified in fraction of the gridcell
```
rosiealice, line 349:
```suggestion
     !  since harvest_rate is specified as a fraction of the gridcell
```

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -235,6 +298,93 @@ end subroutine LoggingMortality_frac
...
+        endif
+     endif
+
+     ! calculate today's harvest rate
+     ! whether to harvest today has already been determined by IsItLoggingTime
+     ! for icode == 2, icode < 0, and icode > 10000 apply the annual rate one time (no calc)
+     ! Bad logging event flag is caught in IsItLoggingTime, so don't check it here
+     icode = int(logging_event_code)
+     if(icode .eq. 1) then
+        ! Logging is turned off - not sure why we need another switch
```
rosiealice, line 372: indeed. This does cause extra confusion.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -213,6 +244,23 @@ subroutine disturbance_rates( site_in, bc_in)
     ! Calculate Disturbance Rates based on the mortality rates just calculated
     ! ---------------------------------------------------------------------------------------------
 
+    ! zero the diagnostic disturbance rate fields
+    site_in%potential_disturbance_rates(1:N_DIST_TYPES) = 0._r8
+
+    ! summarize how open the canopy is prior to resolving the disturbance
```
rosiealice, line 250:
```suggestion
    ! Recalculate total canopy area prior to resolving the disturbance
```

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -255,9 +303,26 @@ subroutine disturbance_rates( site_in, bc_in)
...
 
        ! Fire Disturbance Rate
-       ! Fires can't burn the whole patch, as this causes /0 errors. 
        currentPatch%disturbance_rates(dtype_ifire) = currentPatch%frac_burnt
 
+       ! for non-closed-canopy areas subject to logging, add an additional increment of area disturbed
+       ! equivalent to the fradction loged to account for transfer of interstitial ground area to new secondary lands
+       if ( logging_time .and. &
+            (currentPatch%area - currentPatch%total_canopy_area) .gt. fates_tiny ) then
+
```
rosiealice, line 312: The canopy is NOT closed.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -255,9 +303,26 @@ subroutine disturbance_rates( site_in, bc_in)
...
+          call get_harvest_rate_area (currentPatch%anthro_disturbance_label, bc_in%hlm_harvest_catnames, bc_in%hlm_harvest, &
+                 frac_site_primary, currentPatch%age_since_anthro_disturbance, harvest_rate)
+
+          currentPatch%disturbance_rates(dtype_ilog) = currentPatch%disturbance_rates(dtype_ilog) + &
+               (currentPatch%area - currentPatch%total_canopy_area) * harvest_rate / currentPatch%area
+       endif
+
```
rosiealice, line 319:
```suggestion
! Sum of disturbance rates for different classes of disturbance across all patches in this site. 
```

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1005,12 +1094,41 @@ subroutine spawn_patches( currentSite, bc_in)
...
+             end do
+             if (associated(currentPatch)) then
+                ! the case where we've found a youngest primary patch
+                new_patch_primary%older    => currentPatch
+                new_patch_primary%younger  => currentPatch%younger
+                currentPatch%younger%older => new_patch_primary
+                currentPatch%younger       => new_patch_primary
+             else
+                ! the case where we haven't, and are putting a primary patch at the oldest end of the 
+                ! linked list (not sure how this could happen, but who knows...)
```
rosiealice, line 1116: haha..

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1005,12 +1094,41 @@ subroutine spawn_patches( currentSite, bc_in)
...
+             foundfirstprimary = .false.
+             do while(associated(currentPatch) .and. .not. foundfirstprimary) 
+                currentPatch => currentPatch%older
+                if (associated(currentPatch)) then
+                   if (currentPatch%anthro_disturbance_label .eq. primaryforest) then
+                      foundfirstprimary = .true.
+                   endif
+                endif
+             end do
+             if (associated(currentPatch)) then
```
rosiealice, line 1108: At this point, 'currentpatch' is the 'firstpatchprimary', right?  
The sense of 'first' isn't necessarily obvious. Could it be 'foundyoungestprimary'?

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1005,12 +1094,41 @@ subroutine spawn_patches( currentSite, bc_in)
...
+             else
+                ! the case where we haven't, and are putting a primary patch at the oldest end of the 
+                ! linked list (not sure how this could happen, but who knows...)
+                new_patch_primary%older    => null()
+                new_patch_primary%younger  => currentSite%oldest_patch
+                currentSite%oldest_patch%older   => new_patch_primary
+                currentSite%oldest_patch   => new_patch_primary
+             endif
+          else
+             ! the case where there are no secondary patches at the start of the LL (prior logic)
```
rosiealice, line 1123: LL?

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2255,6 +2383,19 @@ subroutine fuse_patches( csite, bc_in )
...
+
+       if (currentPatch%anthro_disturbance_label .eq. primaryforest) then
+          primary_land_fraction_afterfusion = primary_land_fraction_afterfusion + &
+               currentPatch%area * AREA_INV
+       endif
+
+       currentPatch => currentPatch%older
+    enddo
+
+    currentSite%primary_land_patchfusion_error = primary_land_fraction_afterfusion - primary_land_fraction_beforefusion
```
rosiealice, line 2398: what happens to this? Does it get checked somewhere?

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2295,13 +2436,19 @@ subroutine fuse_2_patches(csite, dp, rp)
     inv_sum_area = 1.0_r8/(dp%area + rp%area)
     
     rp%age = (dp%age * dp%area + rp%age * rp%area) * inv_sum_area
+    rp%age_since_anthro_disturbance = (dp%age_since_anthro_disturbance * dp%area &
+         + rp%age_since_anthro_disturbance * rp%area) * inv_sum_area
```
rosiealice, line 2440: what is this in the case of primary patches?

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2468,25 +2618,52 @@ subroutine terminate_patches(currentSite)
...
                 
-             elseif( associated(currentPatch%younger) ) then
+                   ! This logic checks to make sure that the younger patch is not the youngest
+                   ! patch. As mentioned earlier, we try not to fuse it.
+                   
+                   gotfused = .true.
+                else
+                   if (count_cycles .gt. 0) then
+                      ! if we're having an incredibly hard time fusing patches because of their differing anthropogenic disturbance labels, 
+                      ! since the size is so small, let's sweep the problem under the rug and change the tiny patch's label to that of its older sibling
```
rosiealice, line 2636:
```suggestion
                      ! since the size is so small, let's sweep the problem under the rug and change the tiny patch's label to that of its older sibling
                      ! and then allow them to fuse together. 
```

rgknox:
@ckoven , I'm noticing an issue that may crop up now that we are using logging with gridded runs.  We currently determine if it is "logging time" by evaluating the date, but only on the master node.  In a gridded run, the worker threads may not even get evaluated, and/or get to the calculation of the areas before the master updates the global "logging time".

https://github.com/NGEET/fates/blob/master/biogeochem/EDLoggingMortalityMod.F90#L86

I have to look more at how this routine is used, but it might be good to change "logging time" to a site level variable (not a global) and have this evaluated and updated on all threads.

main/FatesConstantsMod.F90
```diff
@@ -34,6 +34,9 @@ module FatesConstantsMod
   integer, parameter, public :: n_anthro_disturbance_categories = 2
   integer, parameter, public :: primaryforest = 1
   integer, parameter, public :: secondaryforest = 2
+  integer, parameter, public :: secondary_age_threshold = 94 ! less than this value is young
```
rosiealice, line 37: What does 'young' imply? That it won't be harvested as 'secondary' forest?

main/FatesHistoryInterfaceMod.F90
```diff
@@ -2199,6 +2247,12 @@ subroutine update_history_dyn(this,nc,nsites,sites)
                hio_crownarea_si_pft(io_si, ft) = hio_crownarea_si_pft(io_si, ft) + &
                     ccohort%c_area 
 
+               if (ccohort%canopy_layer .eq. 1) then
```
rosiealice, line 2250: this is useful!

main/FatesHistoryInterfaceMod.F90
```diff
@@ -4200,6 +4259,52 @@ subroutine define_history_vars(this, initialize_variables)
...
+         long='Disturbance rate from fire',  use_default='active',     &
+         avgflag='A', vtype=site_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
+         ivar=ivar, initialize=initialize_variables, index = ih_fire_disturbance_rate_si )
+
+    call this%set_history_var(vname='DISTURBANCE_RATE_LOGGING', units='m2 m-2 d-1',                   &
+         long='Disturbance rate from logging',  use_default='active',     &
+         avgflag='A', vtype=site_r8, hlms='CLM:ALM', flushval=0.0_r8, upfreq=1,   &
+         ivar=ivar, initialize=initialize_variables, index = ih_logging_disturbance_rate_si )
+
+    call this%set_history_var(vname='DISTURBANCE_RATE_FALL', units='m2 m-2 d-1',                   &
```
rosiealice, line 4293:
```suggestion
    call this%set_history_var(vname='DISTURBANCE_RATE_TREEFALL', units='m2 m-2 d-1',                   &
```

main/FatesInterfaceTypesMod.F90
```diff
@@ -426,7 +446,14 @@ module FatesInterfaceTypesMod
       real(r8),allocatable :: hksat_sisl(:)        ! hydraulic conductivity at saturation (mm H2O /s)
       real(r8),allocatable :: h2o_liq_sisl(:)      ! Liquid water mass in each layer (kg/m2)
       real(r8) :: smpmin_si                        ! restriction for min of soil potential (mm)
-   
+      
+      ! Land use
+      ! ---------------------------------------------------------------------------------
+      logical :: hlm_do_harvest_today           ! denotes whether hlm harvest data
+                                                ! are available for today
+      real(r8),allocatable :: hlm_harvest(:)    ! annual harvest rate per cat from hlm for a site
```
rosiealice, line 454: I think I'd rename this hlm_harvest_rate. Might have said that earlier as well.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -213,6 +244,23 @@ subroutine disturbance_rates( site_in, bc_in)
     ! Calculate Disturbance Rates based on the mortality rates just calculated
     ! ---------------------------------------------------------------------------------------------
 
+    ! zero the diagnostic disturbance rate fields
+    site_in%potential_disturbance_rates(1:N_DIST_TYPES) = 0._r8
+
+    ! summarize how open the canopy is prior to resolving the disturbance
+    currentPatch => site_in%oldest_patch
```
rosiealice, line 251: Has this changed on account of growth since we last calculated it?

ckoven:
@rgknox to this point: its a good question and I don't know the exact answer, except that if the logging time wasn't getting set properly on some gridcells, then I'd expect that the global integral of logging wouldn't match the expected value, which it does.  so I think the code is workin gnow, at least int he configuration that I tested. But if there is a more robust way to do it then let's!

> @ckoven , I'm noticing an issue that may crop up now that we are using logging with gridded runs. We currently determine if it is "logging time" by evaluating the date, but only on the master node. In a gridded run, the worker threads may not even get evaluated, and/or get to the calculation of the areas before the master updates the global "logging time".
> 
> https://github.com/NGEET/fates/blob/master/biogeochem/EDLoggingMortalityMod.F90#L86
> 
> I have to look more at how this routine is used, but it might be good to change "logging time" to a site level variable (not a global) and have this evaluated and updated on all threads.

rgknox:
When harvesting area datasets are active and being passed in, does this change the time-code, or rather how does this somehow set logging_time to true?    I am noticing that new logic is found inside a logging_time check.  For instance here:

https://github.com/NGEET/fates/pull/663/files#diff-135c876b8ed5286434c050b3b223605dL188

biogeochem/EDLoggingMortalityMod.F90
ckoven: I think we left this in to allow the old event-based harvesting code to work.  but maybe we need to get rid of that and make the timeseries code the only way to do the harvest?

biogeochem/EDPatchDynamicsMod.F90
ckoven: Since all patches age at the same rate, the order can only change due to fusion, right?

biogeochem/EDPatchDynamicsMod.F90
ckoven:
```suggestion
             ! the case where there are no secondary patches at the start of the linked list (prior logic)
```

biogeochem/EDPatchDynamicsMod.F90
ckoven: yes, it becomes a history variable.  its possible that we don't need it, I had only added it to try to test a hypothesis about what was going wrong, and found that the values of it were very small.

biogeochem/EDPatchDynamicsMod.F90
ckoven: good question. I guess that it is only used or incremented for secondary patches, but maybe it needs a special fill value for primary patches?

main/FatesConstantsMod.F90
ckoven: this is the distinction between secondary old and secondary young, i.e. in the harvest rates `HARVEST_SH1` vs `HARVEST_SH2` and `HARVEST_SH3`.  It thus should effectively act as a rotation age for the secondary forest harvest.

main/FatesHistoryInterfaceMod.F90
ckoven: agreed.  but should we change the units on both this and the preceding variable (which becomes `PFTcrownarea`) to be m2/m2 rather than m2/ha?

biogeochem/EDLoggingMortalityMod.F90
aldivi: To clarify, hlm_use_logging is set to true if hlm harvest is active or if fates logging is active  (based on namelist inputs). This is so that just one variable is used to determine whether today is a logging/harvest day or not.
The relationship between hlm harvest and logging is this:
Hlm harvest uses hlm input timeseries of harvest prescription to drive the fates logging module, so if hlm harvest is on, then fates logging is automatically turned on.
The fates logging module can still be used on its own by setting its namelist option to true (use_fates_logging), and not setting the hlm do_harvest namelist option (leaving as false).

biogeochem/EDLoggingMortalityMod.F90
aldivi: This means that they can each be set as an input value, and they should run.
The issue is that if the harvest is distributed in time it may be superseded by another disturbance on any given day because the largest disturbance wins the day (which would reduce the total annual harvest simulated).

biogeochem/EDLoggingMortalityMod.F90
aldivi: I think both options can be useful, especially since the fates logging parameters are useful for determining harvest date and the other specifics that are not in the hlm harvest inputs.
This particular value can be eliminated, but it is rather inconsequential, unless set accidentally, so it was left in.

main/FatesConstantsMod.F90
aldivi: It also determines how to apply HARVEST_SH2 (young secondary) and HARVEST_SH3 (secondary non-forest), which presumably have less biomass than mature secondary (to which HARVEST_SH1 is applied). SH2 and SH3 are applied to the "young" patches, and SH1 is applied to the "mature" patches.

rgknox: this is compared against a real (secondary_age), recommend setting the parameter to a real as well

main/FatesInterfaceTypesMod.F90
```diff
@@ -88,9 +88,29 @@ module FatesInterfaceTypesMod
...
+   integer, public :: hlm_use_lu_harvest      ! This flag signals whether or not to use
+                                                         ! harvest area data from the hlm
+                                                         ! 0 = do not use lu harvest from hlm
+                                                         ! 1 = use area fraction of vegetated land
+                                                         ! from from hlm
+                                                         ! 2 = use carbon from hlm
+                                                         ! If 1 or 2, it automatically sets
+                                                         ! hlm_use_logging to 1
+
+   integer, public :: hlm_num_lu_harvest_cats    ! number of hlm harvest categories
```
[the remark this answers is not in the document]
ckoven:
```suggestion
   integer, public :: hlm_num_lu_harvest_cats    ! number of hlm harvest categories (e.g. primary forest harvest, secondary young forest harvest, etc.)
```

main/FatesInterfaceTypesMod.F90
```diff
@@ -426,7 +446,14 @@ module FatesInterfaceTypesMod
       real(r8),allocatable :: hksat_sisl(:)        ! hydraulic conductivity at saturation (mm H2O /s)
       real(r8),allocatable :: h2o_liq_sisl(:)      ! Liquid water mass in each layer (kg/m2)
       real(r8) :: smpmin_si                        ! restriction for min of soil potential (mm)
-   
+      
+      ! Land use
+      ! ---------------------------------------------------------------------------------
+      logical :: hlm_do_harvest_today           ! denotes whether hlm harvest data
```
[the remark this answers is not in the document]
ckoven: Yeah I think this gets resolved wholly within FATES, so we don't need it here?  @aldivi should we remove it?

main/FatesConstantsMod.F90
ckoven:
```suggestion
  real(fates_r8), parameter, public :: secondary_age_threshold = 94._fates_r8 ! less than this value is young secondary forest
```

main/FatesInterfaceTypesMod.F90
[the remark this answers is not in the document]
aldivi: I think it can be removed. It appears that it was added initially as something that may be used, but then wasn't needed, and then not deleted.

ckoven:
some note from a discussion with @rgknox on this PR:

(1) right now, the CTSM code is assuming that the harvest is area-based (see here: https://github.com/ckoven/ctsm/blob/fates_landuse/src/utils/clmfates_interfaceMod.F90#L340).  the problem is that right now, the way we are specifying the units happens early in the call sequence. Instead, the suggestion is to change this to make sure that the units that are read in (here: https://github.com/ckoven/ctsm/blob/fates_landuse/src/biogeochem/dynHarvestMod.F90#L124) are passed through the interface alongside the rates (also rename these as per https://github.com/NGEET/fates/pull/663#discussion_r448431617).  At some later point we can update that to use a parameter interface that @rgknox has developed for the nutrient code to handle a similar problem, which should be more efficient.


(2) at some point we should add some checks to ensure that parameter values make sense.  in particular, the harvest timeseries code only really makes sense for certain values of the 
need to add a new subroutine to check these constants (like that it is less than 1000).  call this subroutine `fates_checkscalarparams`, called from fatesinterfacemod after here: https://github.com/NGEET/fates/blob/master/main/FatesInterfaceMod.F90#L1410 and like the checks that happen before it.  


(3) a few little things, e.g. change name of `use_history` to `patch_ anthro_disturbance_label` here: https://github.com/ckoven/fates/blob/fates_harvest_offmaster/biogeochem/EDLoggingMortalityMod.F90#L315

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -168,41 +184,88 @@ end subroutine IsItLoggingTime
...
+      real(r8) :: harvest_rate ! the final harvest rate to apply to this cohort today
+
+      ! todo: probably lower the dbhmin default value to 30 cm
+      ! todo: change the default logging_event_code to 1 september (-244)
+      ! todo: change the default logging_direct_frac to 1.0 for cmip inputs
+      ! todo: check outputs against the LUH2 carbon data
+      ! todo: eventually set up distinct harvest practices, each with a set of input paramaeters
+      ! todo: implement harvested carbon inputs
+      
       if (logging_time) then
```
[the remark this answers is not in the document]
ckoven: we discussed this and agreed that the code is correct as-is, since the bc_in%harvest_rates are only really updated yearly and the point of the IsItLoggingTime is to downscale those yearly average rates to some fates-specified frequency (on one specific day per year, or monthly, etc).

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -235,6 +298,93 @@ end subroutine LoggingMortality_frac
...
+      real(r8), intent(in) :: secondary_age     ! patch level age_since_anthro_disturbance
+      real(r8), intent(in) :: frac_site_primary
+      real(r8), intent(out) :: harvest_rate
+
+      ! Local Variables
+      integer :: h_index   ! for looping over harvest categories
+      integer :: icode   ! Integer equivalent of the event code (parameter file only allows reals)
+
+      ! Parameters
+      real(r8), parameter :: months_per_year = 12.0
```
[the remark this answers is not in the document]
ckoven: done

biogeochem/EDLoggingMortalityMod.F90
rosiealice: fair enough!

biogeochem/EDLoggingMortalityMod.F90
rosiealice: Ah yes. I think I understood this more as I went through the code...

biogeochem/EDPatchDynamicsMod.F90
rgknox: current patch is the youngest patch, could be primary or secondary as far as I can tell.
https://github.com/NGEET/fates/pull/663/files/cbc4ca001f77edf68b6651082d8150fd9e1e40b2#diff-531e45e75218bb11634691970470c90bR1096

rosiealice: ...but you would have only exited the 'do while' loop if you had -just- found firstpatchprimary, right? 

Unless there are no primary patches?

rgknox: I agree with @rosiealice , also, could you change the comment above to indicate it is looking for the youngest primary patch?  It was confusing trying to figure out the intent as "! insert first primary patch after all secondary patches, if any" was misleading

rgknox: Also, I'm working through what would happen if you only had one patch, and it was secondary...  actually, you provide for that condition below. I see, it would become the oldest patch..  But then again, if there was only 1 patch and it was secondary, we would have no site_areadis_primary, which would mean we are not in this loop anyway.  

I can't find any holes in this logic.

rgknox:
> @ckoven , I'm noticing an issue that may crop up now that we are using logging with gridded runs. We currently determine if it is "logging time" by evaluating the date, but only on the master node. In a gridded run, the worker threads may not even get evaluated, and/or get to the calculation of the areas before the master updates the global "logging time".
> 
> https://github.com/NGEET/fates/blob/master/biogeochem/EDLoggingMortalityMod.F90#L86
> 
> I have to look more at how this routine is used, but it might be good to change "logging time" to a site level variable (not a global) and have this evaluated and updated on all threads.

After looking at the code more, I realize that we do calculate logging time on all threads, we only use the pointer to the main thread to filter out print statements to prevent redundant messaging.  This call may be inefficient (although it is incredibly cheap compared to other operations its not worth improving) but I don't think there is anything wrong anymore.

biogeochem/EDPatchDynamicsMod.F90
ckoven: ok, changing foundfirstprimary to found_youngest_primary, and also changing/adding to the comments.

biogeochem/EDPatchDynamicsMod.F90
ckoven: actually, I think this can't happen in the current code, but if we use this as a template for the prescribed biography with competition then it might?  the way i think we could generalize this is to make an integer patch label that includes both the pft type (if used) and the anthro land use type and then iterate through this to put a given new patch such that they are in patch-label order and then within that in age order?

ckoven: actually, thinking about this even more, i think it could possibly happen if there were only one primary patch that was wholly consumed by natural disturbance in a single day.  not sure how that might happen but I'm sure in an ESM it somehow will.  so best to allow for that edge case.

biogeochem/EDPatchDynamicsMod.F90
ckoven: looking at this (here: https://github.com/NGEET/fates/blob/master/biogeochem/EDPatchDynamicsMod.F90#L1884), it seems that the current code sets this as -1 for primary patches.  there's a note to replace it with fates_unset_r8 when possible.  I guess I'll do that for consistency?

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -168,41 +185,92 @@ end subroutine IsItLoggingTime
...
+            
+            if (dbh >= logging_dbhmin .and. .not. &
+                 ((logging_dbhmax < fates_check_param_set) .and. (dbh < logging_dbhmax )) ) then
+               lmort_direct = harvest_rate * logging_direct_frac
+               l_degrad = harvest_rate * (1._r8 - logging_direct_frac) ! fraction passed to 'degraded' forest. 
             else
-               lmort_direct = 0.0_r8 
-               l_degrad = logging_direct_frac * adjustment
+               lmort_direct = 0.0_r8
+               l_degrad = harvest_rate
```
KatieMurenbeeld, line 266: Does this mean that if trees are not large enough to log, the cohort is still degraded/moved to secondary patch?

KatieMurenbeeld: Oh, never mind I see, background degradation rate regardless of logging.

ckoven:
@aldivi and I have had a bunch of discussions in the past days with various stakeholders (@JoshuaRady @KatieMurenbeeld @lawrencepj1 @dlawrenncar @wwieder) and I think that all are okay with this PR as it currently stands, so I think it is ready to test and then merge if tests pass.

lawrencepj1:
Hi Charlie

Yes, I agree that PR will enable the functionality required by FATES to do
prescribed wood harvest with no impact on the non FATES wood harvest. I
also agree with Dave @David Lawrence <dlawren@ucar.edu> idea on using the
streams functionality long term as there will be information from the LUH2
data that is needed for FATES which has been normalized to PFT level in the
non FATES wood harvest.

Peter
-- 
Dr Peter Lawrence
Terrestrial Science Section
National Center for Atmospheric Research
1850 Table Mesa Drive
Boulder Colorado 80305

Work: 1-303-497-1727
Cell: 1-303-956-6932

glemieux:
Regression tests underway

ckoven:
noting here some diagnostics based on a global 4x5 historical tranisent run that I made in a jupyter notebook here.  The notebook shows some issues that needed to be worked out on the version of the code that I used for the run, which I believe to have been corrected since, but I am also noting here that while I think this code is now ready to be merged I want to return to these diagnostics at a later date once we are using the raw LUH2 as the driver (and also have better global productivity and biomass behavior).

https://github.com/ckoven/landuse_runscripts/blob/master/landuse_global_analysis.ipynb

glemieux:
Ran regression test on update post bugfixes.  All expected PASS.  Only differences due to new history variables added and answer changing results for the `FatesLogging` test which is expected.  There is also an expected difference in the `SITE_DAYSINCE_COLDLEAFOFF` (and related) that is due to reinstating [CTSM PR 820](https://github.com/ESCOMP/CTSM/pull/820) which was accidentally dropped during updating fates_next_api to relase-clm5.0.30.  Test results located here:
```
/glade/u/home/glemieux/scratch/clmed-tests/pr663-landuse-C39a5a119-F8872a912.fates.cheyenne.intel
/glade/u/home/glemieux/scratch/clmed-tests/pr663-landuse-C39a5a119-F8872a912.fates.cheyenne.gnu
```

PR #685, opened 2020-08-14
Added the phen_model_day timer to work with all restart conditions

### LINKED ISSUE #627: Phenology and restarting
So, in trying to make a run using a restart files from a previous run (and turning spitfire on, a'la @jkshuman 's discovery that we can't just run fire from the beginning), I discovered that all of the cold deciduous trees had died in the new run (is that a 'branch' run? I've never been very au fait with that terminology). 

Anyway, I think the issue is that we are recording as our 'state' variable for the phenology counters the absolute date of the leafon and leafoff events. As in, after 30 years, the 
fates_drought_leafoffdate and fates_leafoffdate variables have values in the 10'000s... 

When we restart the run from the beginning, the model recalculates 'timesinceleafoff' and 'timesinceleafon' from model_day_int and the restart leafoffdate and leafondate variables read in from the restart. That means that we get huge negative values for timesinceleafoff and timesinceleafon
https://github.com/NGEET/fates/blob/3b6ac22180b0cc873005a141d7967c4fa9fd6b45/biogeochem/EDPhysiologyMod.F90#L701

Which of course then breaks all of the conditional statements that use them and the phenology routines go haywire. 

I'm not sure if there is a solution to this that doesn't require a re-write of this part of the phenology code. Putting this here for sanity checking.

### LINKED ISSUE #894: Phenology flags get locked in when starting. anew case using. prior case's restart
I ran into a weird issue that I want to note here.  TLDR is that the problem is fixed by #685, so once that is integrated it shouldn't happen any more, and if using older code and seeing similar behavior, merge in the  #685 code.

Slightly longer description is that I was spinning up runs and then trying to take the restarts from the spinup run to start new simulations. But when I did that, I was seeing fundamentally different model behavior for the new cases than the spun-up dynamics of the prior case. It happened both when I set the restart file via a hybrid case or when I manually set the `finidat` argument in the `user_nl_clm` file.  On discussing with @jenniferholm, @rgknox and others, we isolated the problem as one where the phenology flags were no longer updating in the new cases, and instead just remaining constant because of how the calendar logic interacted with the phenology logic.  So plants that happened to have their leaves on at the time of the restart got infinite leaf lifespan and those that didn't got zero leaf lifespan.  When I merged in the logic from #685, the behavior went back to what is expected.

rgknox:
### Description:

The phen_model_day timer was added to the site structure.  This timer is initialized as the day of year on a cold-start, and is read in from the restart file.  This replaces the hlm_model_day, which for some types of restarts, will not be continuous from the last simulation.  (ie it may be reset) By using phen_model_day, we have consistency across restarts, and this absolute integer day will always be preserved.

Note: Users who restart the model via an finidat, or non CONTINUE_RUN simulation, AND, have a gap in the DAY OF YEAR from the end of one simulation to the start of the next, will likely still encounter phenology weirdness.  However, it seems reasonable that they should expect weirdness if they decide to magically remove a portion of the year from the plants life-cycle.

Fixes #627 
Fixes #894

### Collaborators:

@rosiealice @adrifoster 

Yes, this will change answers, but only on non CONTINUE_RUN restart scenarios, and only on deciduous vegetation.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:


rgknox:
See test results from 2 simulations at bci, 10 years each, 4 pfts (our drought deciduous pfts), the base is master, the test is this branch.
[PMD-test_plots.pdf](https://github.com/NGEET/fates/files/5077218/PMD-test_plots.pdf)

Running a hybrid test now, where I start a second simulation in the 5th out of 10th year.  In a hybrid, the reference date should change, so this test should indicate if the our timers are preserved through the restart.

rgknox:
hybrid test looks good.  I ran two simulations, 1) a base simulation that was a 10 year cold-start, and a 5 year hybrid run that used the base simulations 5th year as an finidat (initial condition).  Model output from the hybrid run and the base run are indistinguishable, included "days since leaf on and leaf off" diagnostics.

[hybr-test2_plots.pdf](https://github.com/NGEET/fates/files/5085092/hybr-test2_plots.pdf)

rosiealice (approved):
Sorry for taking so long to have the headspace to get to this! It all looks fine to me, and I think will prove useful for ongoing phenology updates...

rgknox:
tested on cheyenne, "fates" test suite:
/glade/scratch/rgknox/tests_0825-130322ch
I believe tests are nominal (still checking), all tests have expected diffs with BASELINE, but the diffs are only in the variables that had their calculations changed:
```
 FATES_DAYSINCE_COLDLEAFOFF
 FATES_DAYSINCE_COLDLEAFON
 FATES_DAYSINCE_DROUGHTLEAFOFF
 FATES_DAYSINCE_DROUGHTLEAFON
```
Expected FAILS: 
ERS_Lm12.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesFireLightningPopDens COMPARE_base_rest
ERS_D_Ld15.f45_f45_mg37.I2000Clm50FatesRs.cheyenne_intel.clm-FatesColdDefTreeDamage

PR #686, opened 2020-08-14
Moved taper exponent to parameter file value

### LINKED ISSUE #566: parameter: fates_hydr_p_taper is not used
The parameter that controls the effects of taper on xylem conductivity are not being used.  The value in the default parameter file does match the constant used in the code.

Here is the value in the file:
https://github.com/NGEET/fates/blob/master/parameter_files/fates_params_default.cdl#L772

Here is the value in the memory:
https://github.com/NGEET/fates/blob/master/main/EDPftvarcon.F90#L229

But this is the value used in the code:
https://github.com/NGEET/fates/blob/master/biogeophys/FatesPlantHydraulicsMod.F90#L660

rgknox:
### Description:

As title implies, this simple fix removes the hard-coded taper exponent in favor of the value in the parameter file.

Fixes #566 

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeophys/FatesPlantHydraulicsMod.F90
+++ b/biogeophys/FatesPlantHydraulicsMod.F90
@@ -2691,7 +2691,6 @@ contains
     real(r8) :: surfarea_aroot_layer ! Surface area of absorbing roots in each
                                      ! soil layer [m2]
     real(r8) :: sum_l_aroot          ! sum of plant's total root length
-    real(r8),parameter :: taper_exponent = 1._r8/3._r8 ! Savage et al. (2010) xylem taper exponent [-]
     real(r8),parameter :: min_pet_stem_dz = 0.00001_r8  ! Force at least a small difference
                                                        ! in the top of stem and petiole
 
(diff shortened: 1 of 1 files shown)


rgknox:
testing on cheyenn

rgknox:
The taper exponent in the file is 0.333, but the parameter in code was defined as the fraction, 1./3., so this does not generate b4b results in hydro runs.  All other tests are b4b with base. 

/glade/scratch/rgknox/clmed-tests/fates-sci.1.40.4_api.13.0.1-clm5.0.30-Cabcd5937-F4832567e.fates.cheyenne.intel
/glade/scratch/rgknox/clmed-tests/fates-sci.1.40.4_api.13.0.1-clm5.0.30-Cabcd5937-F4832567e.fates.cheyenne.gnu

The differences in answers seem to be about the same order of magnitude as the rounding error between the two parameters, which is e-4.   Some differences have absolute values at e-3.

ckoven:
doesn't seem like there's a lot of scope in the PR for any other changes to have snuck in, so that non-bit-for-bit-ness seems a-ok with me.

glemieux:
I agree with Charlie.  By manually tracing a few of the history variables that have the RMS differences, this looks to be consistent with the changes to the `kmax` calculations within the `UpdatePlantKmax`subroutine.

PR #689, opened 2020-08-31
fixes to logging: dbhmax logic and hlm_use_lu_harvest checking

ckoven:
Fixes two things: the logic on logging_dbhmax was reversed, and the ok-values checking on hlm_use_lu_harvest wasn't right.

### Description:
1. Basically because of the .and. .not. in the dbh logic, the comparison to logging_dbhmax was reversed.
2. Since the area-based vs mass-based logic is now handled by a separate variable (hlm_harvest_units), hlm_use_lu_harvest is only a true/false switch so updated the error checking and documentation to reflect that.

I worked through all the combinations of possible logic on the dbh checking to make sure it is right this time...

          [dbh > dbhmin] .and. not. ([dbhmax invoked] .and. [dbh >= dbhmax]) =>  [do we log this tree?]
          T                                  F                      T            T
          T                                  F                      F            T
          T                                  T                      T            F
          T                                  T                      F            T
          F                                  F                      T            F
          F                                  F                      F            F
          F                                  T                      T            F
          F                                  T                      F            F

### Collaborators:
Both of these were caught by @aldivi.  

not unless logging_dbhmax is set as other than missing value in the parameter file.

### Checklist:

### Test Results:
not tested yet.

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDLoggingMortalityMod.F90
+++ b/biogeochem/EDLoggingMortalityMod.F90
@@ -259,7 +259,10 @@ contains
             
             ! direct logging rates, based on dbh min and max criteria
             if (dbh >= logging_dbhmin .and. .not. &
-                 ((logging_dbhmax < fates_check_param_set) .and. (dbh < logging_dbhmax )) ) then
+                 ((logging_dbhmax < fates_check_param_set) .and. (dbh >= logging_dbhmax )) ) then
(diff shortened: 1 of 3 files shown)


glemieux:
All tests PASS, with two DIFFs on hydro tests:
```
/glade/u/home/glemieux/scratch/clmed-tests/pr689.fates.cheyenne.intel.Cfde33f56-F2eb4da6e
```

PR #692, opened 2020-09-10
No Competition (NOCOMP) mode

rosiealice:
### Description:
This PR implements the initial phase of the 'no competition' mode.
Each patch is occupied by a single PFT, and has an area that is fixed in time and determined from the surface dataset.

This can run either:
-with fixed_biogeog mode (in which case the areas of PFTs are inputs)
or 
-without fixed_biogeog mode (in which case each PFT gets the same area). 

This mode is turned on using the 'use_fates_nocomp' switch in user_nl_clm
This does NOT implement satellite phenology mode. 

To be fully functional, additional measures are needed to allow multiple patches to occur within the same PFT space. I have code that does this, but it currently has a memeory leak. Adding this PR now as other developments downstream need this version to be operational. 

This should be paired with the PR in CTSM (to be added shortly) which changes the interface to add the use_fates_nocomp switch. 

Sync with: https://github.com/ESCOMP/CTSM/pull/1147

### Collaborators:
@ckoven 
@glemieux 
@rgknox 

This will change answers if it is on. 
and ideally not if it is off! 

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2674,43 +2697,149 @@ subroutine terminate_patches(currentSite)
                       call fuse_2_patches(currentSite, youngerPatch, currentPatch)
                       gotfused = .true.
                    endif
+                endif               
+                ! The fusion process has updated the "younger" pointer on currentPatch
+                
+             endif ! older or younder patch
+          endif ! very small area
+
+        else !nocomp. We cannot fuse to patches with a different PFT identity in no competition mode.
```
ckoven, line 2706: I guess this is also still placeholder code (since disturbance isn't turned on if nocomp is active for this PR) but I still don't understand why this code below would be required -- why can't you just change the logic on line 2632 (old numbering) from 
`                if (currentPatch%anthro_disturbance_label .eq. olderPatch%anthro_disturbance_label) then`

 to

`                if (currentPatch%anthro_disturbance_label .eq. olderPatch%anthro_disturbance_label .and. 
currentPatch%nocomp_pft_label .eq. olderPatch%nocomp_pft_label) then`

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2674,43 +2697,149 @@ subroutine terminate_patches(currentSite)
...
-             endif
-          endif
-       endif
+               endif
+              fusingPatch => fusingPatch%younger 
+            enddo !fusing patch
+
+           is_oldest = itrue !try and find a younger same-PFT patch                                                 
+           !--------------------------------------------------
+           ! n.b. The following code is to figure out how to
```
ckoven, line 2730: this also seems like it should be subsumed under the anthro_dist_label logic?  the sum of all patch areas with a given nocomp_pft_label is fixed in time, so the only case where we should be able to run into logic with tiny patches that need to be force-merged is if there is a tiny patch with a potential fuser whose nocomp_pft_label matches but a anthro_dist_label that doesn't match? or am I thinking about this wrong?

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2674,43 +2697,149 @@ subroutine terminate_patches(currentSite)
...
        ! small patch, resulting in an incredibly small patch.  It is also possible that this
        ! resulting incredibly small patch is the oldest patch.  If this was true than
        ! we would had been at the end of the loop, and left with an incredibly small patch.
        ! Think this is impossible? No, this really happens, especially when we have fires.
        ! So, we don't move forward until we have merged enough area into this thing.
 
+
        if(currentPatch%area > min_patch_area_forced)then
-          currentPatch => currentPatch%older
+          currentPatch => oldercPatch
```
ckoven, line 2805: why is this needed? (not suggesting it isn't, but curious)

main/EDInitMod.F90
```diff
@@ -293,11 +301,40 @@ subroutine set_site_properties( nsites, sites,bc_in )
...
           sites(s)%frac_burnt = 0.0_r8
          
-         ! PLACEHOLDER FOR PFT AREA DATA MOVED ACROSS INTERFACE                                                                                   
-          if(hlm_use_fixed_biogeog.eq.itrue)then
+
+          if(hlm_use_fixed_biogeog.eq.itrue)then  
+           ! MAPPING OF FATES PFTs on to HLM_PFTs
+            ! add up the area associated with each FATES PFT
+            ! where pft_areafrac is the area of land in each HLM PFT and (from surface dataset)
+            ! hlm_pft_map is the area of that land in each FATES PFT (from param file)
```
ckoven, line 309: just to be clear, are the logical constraints of hlm_pft_map that all rows and all columns need to sum to one? if so, can we add a check after parameter reading to ensure that that condition is satisfied?

main/EDInitMod.F90
```diff
@@ -293,11 +301,40 @@ subroutine set_site_properties( nsites, sites,bc_in )
...
+            sites(s)%area_pft(1:numpft) = 0._r8
+            do hlm_pft = 1,size( EDPftvarcon_inst%hlm_pft_map,2)
+               do fates_pft = 1,numpft ! loop round all fates pfts for all hlm pfts        
+                 sites(s)%area_pft(fates_pft) = sites(s)%area_pft(fates_pft) + &
+                     EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft) * bc_in(s)%pft_areafrac(hlm_pft) 
+               end do
+            end do !hlm_pft
+
+           ! re-normalize PFT area to ensure it sums to one.
```
ckoven, line 318: is this to handle precision errors or to handle cases where the above (rows and columns don't sum to one) isn't satisfied?  if the latter, how do we know that the numbers make physical sense?

main/EDInitMod.F90
```diff
@@ -384,51 +429,107 @@ subroutine init_patches( nsites, sites, bc_in)
 
      else
 
-        !FIX(SPM,032414) clean this up...inits out of this loop
```
ckoven, line 387: :)

main/EDInitMod.F90
```diff
@@ -384,51 +429,107 @@ subroutine init_patches( nsites, sites, bc_in)
...
+           newparea = area
+          end if
+          is_first_patch = 1
+          do n = 1, no_new_patches
+
+           ! set the PFT index for patches if in nocomp mode. 
+           if(hlm_use_nocomp.eq.itrue)then
+             nocomp_pft = n
+           else
+             nocomp_pft = 999
```
ckoven, line 452:
```suggestion
             nocomp_pft = fates_unset_int
```

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2603,15 +2622,19 @@ subroutine terminate_patches(currentSite)
 
     currentPatch => currentSite%youngest_patch
     do while(associated(currentPatch)) 
-       
+       oldercpatch => currentPatch%older       
        if(currentPatch%area <= min_patch_area)then
-          
+
+         if(hlm_use_nocomp.eq.ifalse)then !just fuse to older or younger patch
```
ckoven, line 2628: see below -- I am not sure we want to use alternate logic here, but instead treat both the anthro_disturbance_label and the nocomp_pft_label as similarly as possible

main/EDInitMod.F90
```diff
@@ -487,15 +590,50 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
 
     patch_in%tallest  => null()
     patch_in%shortest => null()
-    
+
+    ! Manage interactions of ixed biogeg (site level filter) and
```
ckoven, line 594:
```suggestion
    ! Manage interactions of fixed biogeog (site level filter) and 
```

main/EDInitMod.F90
```diff
@@ -487,15 +590,50 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
+           use_pft_local(pft) = ifalse ! Case 4 
+        endif
+      endif
+
+    end do
+      
     do pft =  1,numpft
-     if(site_in%use_this_pft(pft).eq.itrue)then
+     if(use_pft_local(pft).eq.itrue)then
        if(EDPftvarcon_inst%initd(pft)>1.0E-7) then
```
ckoven, line 623:
```suggestion
       if(EDPftvarcon_inst%initd(pft)>nearzero) then
```

main/EDInitMod.F90
```diff
@@ -487,15 +590,50 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
 
        allocate(temp_cohort) ! temporary cohort
 
        temp_cohort%pft         = pft
        temp_cohort%n           = EDPftvarcon_inst%initd(pft) * patch_in%area
+       if(hlm_use_nocomp.eq.itrue)then !in nocomp mode we only have one PFT per patch
```
ckoven, line 629: would it be simpler to define this on the line above as: 
`temp_cohort%n           = EDPftvarcon_inst%initd(pft) * AREA`
since, in the default case, there should only be one patch and thus patch_in%area would equal AREA, and then skip the logic below?

main/EDMainMod.F90
```diff
@@ -131,7 +132,7 @@ subroutine ed_ecosystem_dynamics(currentSite, bc_in)
     ! !LOCAL VARIABLES:
     type(ed_patch_type), pointer :: currentPatch
     integer :: el              ! Loop counter for elements
-
+    integer :: do_patch_dynamics ! for some modes, we turn off patch dynamics
```
ckoven, line 135: is this the best place to store this variable, or should it be in some place where flags in general get stored?

ckoven (commented):
Thanks @rosiealice, looks great.  A few comments/suggestions, mainly on patch fusion stuff which I guess shouldn't be active yet if this is meant to only handle the case where nocomp is on but disturbance is off?

main/EDInitMod.F90
rosiealice: Just the one way round, so that for each HLM_PFT the values need to add up to one. I think in principle a FATES_PFT could accept >1 or <1 quantities of HLM_PFT area. I've added a check in 
https://github.com/rosiealice/fates/commit/9eefda82636e318b9594b175bee79d10eaff757f

main/EDInitMod.F90
rosiealice: It's really to handle bare ground.  i.e. in cases where area_pft(1:numpft) is less than one. As I have it now, this bare ground gets subsumed into the PFT areas. This made sense when I was thinking about -only- nocomp mode, (as I had thought that the unclosed canopy in FATES, where things don't grow, should in principle generate bare ground  but actually in SP mode, it's going to cause all of the areas where there is lots of bare ground to have the LAI of the other plant functional types and therefore a big positive discrepancy in grid cell LAI between the surface dataset and FATES. So i need to think of a better way of handling bare ground in SP mode... 

Also it might well be true in dynamic nocomp mode that we want to specific an area of bare ground. I should extend that logic as well.

main/EDInitMod.F90
rosiealice: I guess that should result in the same answer as AREA/cp%area = numpft

I'm a little afraid of just adding in the initd *  AREA statement though, as it is less clear what is going on.

main/EDMainMod.F90
rosiealice: I guess it's not a thing I'd want to pass from the HLM?  Perhaps in FatesGlobals, so if we needed to look at it somewhere else for some reason we could? What do think @rgknox ?

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -560,7 +562,7 @@ subroutine spawn_patches( currentSite, bc_in)
           allocate(new_patch_primary)
 
           call create_patch(currentSite, new_patch_primary, age, &
-                site_areadis_primary, primaryforest)
+                site_areadis_primary, primaryforest,fates_unset_int)
```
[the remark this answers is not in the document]
rosiealice: Yes, this routine doesn't really have a concept yet (in this version) of which PFT we are making the new patches for....

biogeochem/EDPatchDynamicsMod.F90
rosiealice: So, where the patches are not ordered in any particular way with reference to their PFT, the 'younger' pointer does not necessarily point to a patch of the same PFT and therefore the logic from the anthro_disturbance_label section here won't find the right patch to merge with.

Am I right in thinking that the anthro label patch dynamics code aligned all the primary and secondary patches together in the younger-older hierarchy?  If so, we should/could also enforce that when I re do the dynamic nocomp code? 

I think that it is probably easier for now to delete this code, and bring it back once I've figured out how the multi-patch-per-PFT logic works in spawn_patches.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -45,6 +45,8 @@ module EDPatchDynamicsMod
   use FatesInterfaceTypesMod    , only : bc_in_type
   use FatesInterfaceTypesMod    , only : hlm_days_per_year
   use FatesInterfaceTypesMod    , only : numpft
+  use FatesInterfaceTypesMod    , only : hlm_use_nocomp
```
rgknox, line 48: @rosiealice , would we consider a system where the structural controls on demographics are passed in as an integer code, or a string name; as an alternative to more than 1 binary switches (eg hlm_use_nocomp and hlm_use_fixed_biogeog)

main/EDPftvarcon.F90
```diff
@@ -2750,8 +2775,22 @@ subroutine FatesCheckParams(is_master, parteh_mode)
         end if
 
 
-    end do
-
+    end do !ipft
+    
+    ! check that the host-fates PFT map adds to one in both dimension
+
```
rgknox, line 2781: This is a great check to add in.

main/EDInitMod.F90
```diff
@@ -348,9 +385,17 @@ subroutine init_patches( nsites, sites, bc_in)
      real(r8) :: biomass_stock
      real(r8) :: litter_stock
      real(r8) :: seed_stock
-     
+     integer  :: n
```
rgknox, line 388: variable definitions text please

main/EDInitMod.F90
```diff
@@ -384,51 +429,107 @@ subroutine init_patches( nsites, sites, bc_in)
...
-
-           call create_patch(sites(s), newp, age, area, primaryforest)
-           
-           ! Initialize the litter pools to zero, these
-           ! pools will be populated by looping over the existing patches
-           ! and transfering in mass
-           do el=1,num_elements
-              call newp%litter(el)%InitConditions(init_leaf_fines=0._r8, &
+          if(hlm_use_nocomp.eq.itrue)then
+           no_new_patches = numpft
```
rgknox, line 439: no_new_patches is short for number of new patches.  is the no_ part latin?

rgknox: or is no_ for no_comp?

main/EDInitMod.F90
```diff
@@ -384,51 +429,107 @@ subroutine init_patches( nsites, sites, bc_in)
...
+           else  ! The default case is initialized w/ one patch with the area of the whole site. 
+             newparea = area       
+           end if 
+
+           if(newparea.gt.0._r8)then ! Stop patches being initilialized when PFT not present in nocomop mode 
+              allocate(newp)
+
+              call create_patch(sites(s), newp, age, newparea, primaryforest, nocomp_pft)
+             
+              if(is_first_patch.eq.1)then !is this the first patch?
```
rgknox, line 473: recommend using itrue instead of 1

main/EDInitMod.F90
```diff
@@ -384,51 +429,107 @@ subroutine init_patches( nsites, sites, bc_in)
...
+                ! set poointers for first patch (or only patch, if nocomp is false)
+                newp%patchno = 1
+                newp%younger => null()
+                newp%older   => null()
+                sites(s)%youngest_patch => newp
+                sites(s)%oldest_patch   => newp
+                is_first_patch = 0
```
rgknox, line 480: recommend using ifalse instead of 0

main/EDInitMod.F90
```diff
@@ -384,51 +429,107 @@ subroutine init_patches( nsites, sites, bc_in)
...
-           do el=1,num_elements
-              call newp%litter(el)%InitConditions(init_leaf_fines=0._r8, &
+          if(hlm_use_nocomp.eq.itrue)then
+           no_new_patches = numpft
+!           allocate(newppft(numpft))
+          else
+           no_new_patches = 1
+           newparea = area
+          end if
+          is_first_patch = 1
```
rgknox, line 445: recommend setting to itrue instead of 1

main/EDInitMod.F90
```diff
@@ -384,51 +429,107 @@ subroutine init_patches( nsites, sites, bc_in)
...
+        !check if the total area adds to the same as site area
+         tota = 0.0_r8
+         newp => sites(s)%oldest_patch
+         do while (associated(newp))
+           tota=tota+newp%area
+             if ( debug ) write(fates_log(),*)  'test links',s,newp%nocomp_pft_label,tota
+           newp=>newp%younger
+         end do
+         if(abs(tota-area).gt.nearzero)then
+             write(*,*) 'error in assigning areas in init patch',s,tota-area
```
rgknox, line 518: should this have an endrun() call along with it?

main/EDInitMod.F90
```diff
@@ -384,51 +429,107 @@ subroutine init_patches( nsites, sites, bc_in)
...
+         endif 
+
+          ! For carbon balance checks, we need to initialize the 
+          ! total carbon stock
+         write(*,*) 'calling sitemassstock',s
```
rgknox, line 523: recommend using:  if ( debug ) write(fates_log(),*) 'calling sitemassstock',s

main/EDInitMod.F90
```diff
@@ -487,15 +590,50 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
-    
+
+    ! Manage interactions of fixed biogeog (site level filter) and 
+    ! nocomp (patch level filter) 
+    ! Need to cover all potential biogeog x nocomp combinations 
+    ! 1. biogeog = false. nocomp = false: all PFTs on (DEFAULT)
+    ! 2. biogeog = true.  nocomp = false: site level filter
+    ! 3. biogeog = false. nocomp = true : patch level filter
+    ! 4. biogeog = true.  nocomp = true : patch and site level filter
+    ! in principle this could be a patch level variable.
```
rgknox, line 601: I think I follow the logic in this section and have no issues

main/EDMainMod.F90
rgknox: I think its ok where it is, because it is local to just this routine right?

main/EDInitMod.F90
rosiealice: Yes, the the latin-y one...

main/EDInitMod.F90
rosiealice: sure

main/EDInitMod.F90
rosiealice: yep

main/EDInitMod.F90
rosiealice: yeah i should've got rid of that one...

biogeochem/EDPatchDynamicsMod.F90
rosiealice: punting this until the reduced complexity mode structure is more established....

main/EDInitMod.F90
rosiealice: I've added a comment to the effect that this  is the same as initd* AREA

PR #693, opened 2020-09-18
minor fixes to ncvarsort and parameter file

ckoven:
This PR has two changes: (1) is to make ncvarsort.py python-3 compliant, and (2) is to get rid of the no-longer-used fates_variants dimension to both the param file and to ncvarsort.

### Description:
that's pretty much it.  the fates_variants was being used to handle the alternate specification of rooting profiles but isn't used anymore.  so we may as well get rid of it.

### Collaborators:
none

none

### Checklist:

### Test Results:
non tested, should be trivial though since no fortran or other code used during the testing was modified.

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/parameter_files/fates_params_default.cdl
+++ b/parameter_files/fates_params_default.cdl
@@ -11,7 +11,6 @@ dimensions:
 	fates_pft = 12 ;
 	fates_prt_organs = 6 ;
 	fates_string_length = 60 ;
-	fates_variants = 2 ;
 variables:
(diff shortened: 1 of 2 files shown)


rgknox (approved):

rgknox:
I performed a simple run test with a newly built parameter file and there are no issues.  As for updating the default parameter file in the SVN repository, I see no need currently.  Merging.

PR #695, opened 2020-09-21
[WIP] Changes to be compatible with renames in ELM

bishtgautam:
Minor changes in FATES to be compatible with name changes in ELM
### Description:
These changes make FATES to be consistent with https://github.com/E3SM-Project/E3SM/pull/3842

### Collaborators:
None

No expected answer changes.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeophys/FatesPlantHydraulicsMod.F90
+++ b/biogeophys/FatesPlantHydraulicsMod.F90
@@ -94,7 +94,7 @@ module FatesPlantHydraulicsMod
   use PRTGenericMod,          only : leaf_organ, fnrt_organ, sapw_organ
   use PRTGenericMod,          only : store_organ, repro_organ, struct_organ
 
-  use clm_time_manager  , only : get_step_size, get_nstep
+  use elm_time_manager  , only : get_step_size, get_nstep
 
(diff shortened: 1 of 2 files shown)


rgknox:
@bishtgautam , these changes identify some vulnerabilities in the FATES code.  I didn't realize that we were pointing to these ELM/CLM global data structures. If we impose these changes as is, we will generate conflicts with CTSM.   I will create and issue and then craft some calls to pass these values as arguments and boundary condition variables.

rgknox:
@bishtgautam ,  I added the following changes, which will update PR #692 

https://github.com/rosiealice/ctsm/pull/4
https://github.com/rosiealice/fates/pull/3

We also have logic around the FATES code that will allow us to do things differently, depending on which HLM we are connected to.  The variable we pass through the interface is here (in E3SM): https://github.com/E3SM-Project/E3SM/blob/master/components/elm/src/main/clmfates_interfaceMod.F90#L290

Before we change that string to "ELM", I need to update the code on the FATES side to use "ELM" and not "ALM" which is what we now have.

bishtgautam:
I'm closing this PR as another PR will be addressing the issues raised here.

Issue #631, opened 2020-04-14
Mapping HLM PFTs onto FATES PFTS in reduced complexity modes


rosiealice:
Our current plan for the SP and fixed biogeography (FBG) modes, is to have FATES read the HLM surface datasets and pass the PFT areal fractions into FATES, whereupon FATES will use that information to restrict the distribution of PFTs geographically. However, until now we have had no need for the FATES and HLM PFT definitions to be in any way compatible, and so this new structure raise the question of how to manage that interface.  

@ckoven @rgknox @glemieux and I had a discussion about thin on 13/04/20 which was inconclusive and I am going to try and paraphrase here for wider discussion. 

There are broadly **three scenarios**, defined here with HLM:FATES as the nomenclature. 

**1:1** -  The FATES PFT are the same as the HLM PFTs, and no mapping is required.

**many:1** - Each FATES PFT corresponds to >=1 HLM PFTs. e.g. FATES Evergreen Needleaf Tree corresponds to the boreal and temperature ENT's in the HLM. 

**1:many** - The HLM defines one PFT per biome whereas FATES uses 'n' PFTs to describe that biome. (e.g. the HLM has one BET whereas FATES has five BETs defined by hydraulic strategy).

The options we discussed for managing this were 
### 1. Invent a tool to convert the surface datasets to the same PFT format as FATES . 
I had thought this might be easiest but @ckoven pointed out that once we get into managing land use datasets the overheads would be much higher, so probably not going to go with that. 

### 2. Embed mapping into FATES that converts from HLM to FATES PFT structures
The dimensionality of the mapping is perhaps the most critical decision here. It could either have:
i) The dimensions of the HLM PFT file. And implement the many:1 case by having integers in each element corresponding to the FATES PFTs e.g.  for a case where the HLM has 12 PFTs and FATES 6, that would look something like this:
[1 1 2 2 2 3 4 4 5 5 6 6]

ii)  The dimensions of the FATES PFT file x a maximum umber of HLM PFTs that could be shared by a FATES PFT. For the many:1 case that would look something like
[1,2; 3,4,5;  6;  7,8;  9,10;  11,12] (where semi-colons delineate new columns and commas rows in columns) 

iii) The dimensions of a matrix, wherein the elements could be fractions. For the many:1 case this would simply make the above information more sparse (and, I think, would basically need to be collapsed down into something like option 1 or 2 to be used in the code anyway) 
FATES PFT:  1 2 3 4 5 6
[...]

Issue #610, opened 2020-02-12
Proscribing LAI/Satellite Phenology


rosiealice:
At the EMERALD meeting here in Oslo, @huitang-earth was describing some of his difficulties trying to test the FATES performance wrt albedo (one of the work packages in EMERALD is to do with albedo).  Using static stand structure to generate LAI consistant with observations appears tricky; the mapping between what we give the model in terms of tree density & size is non-trival to map onto LAI, in the presence of changing canopy spread, etc. Thus, even 'backing out' an initialization file which would be consistant with a given LAI is not a very promising approach as far as I can tell. 

Hui was thus wondering if we could conceive of a way to proscribe LAI. I thought that a literal 'satellite phenology' mode might work, overriding the code before we call 'leaf_area_profile', and then, of course disable all code related to allocation...  

Can anyone think of alternatives/objections/potential for such an idea? We are going to try and think it though some more tomorrow but I wondered if anyone had ideas in the interim.

rgknox:
How about some iterative approach, where we search through parameters defining a Weibul (or other) size distribution of stem densities, using fully flushed crowns, seeking to have best agreement with maximum observed LAI? (I don't know how this would work for more than 1 pft).  We would then run in ST3 mode.

glemieux:
@rosiealice we can close this per #703 yeah?

PR #684, opened 2020-08-14
Added carbon storage buffer to drought deciduous leaf-out

rgknox:
### Description:

As title says, this protects the transfer of storage to leaf flushing on drought deciduous trees by capping the maximum transferable. This is already done with cold deciduous trees.

Fixes #615 

### Collaborators:

Issue identified by @rosiealice 

### Expectation of Answer Changes:

This will change answers (with drought deciduous pfts present, which is true in the testing file).

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/master/CONTRIBUTING.md) document.
- [x] FATES PASS/FAIL regression tests were run
- [ ] If answers were expected to change, evaluation was performed and provided

### Test Results:

TBD

CTSM (or) E3SM (specify which) test hash-tag:

CTSM (or) E3SM (specify which) baseline hash-tag:

FATES baseline hash-tag:

Test Output:

PR #703, opened 2020-09-28
Fates SP mode

rosiealice:
### Description:
This implements FATES SP (satellie phenology) mode (issues #167, #610 and #631.)
and also is part of the Reduced Complexity project
https://github.com/NGEET/fates/projects/5

It allows FATES TLAI, TSAI and HTOP to be driven from the HLM datasets. 
Each PFT has one patch, and that one patch has one cohort. 
The height of the cohort is driven using HTOP, from which we find DBH, crown area, and therefore 'n' (assuming a closed canopy in all cases). 
Tree leaf carbon is the determined by inverting the treelai calculation (forced again from the HLM surface dataset). 

Cohort dynamics and patch dynamics are OFF. Thus allowing only surface exchange/canopy physiology/photosynthesis processes to be assessed, in the absence of 'errors' in LAI (other than those present in the surface dataset itself). 

This PR is paired with changes to the HLM interface as per 
https://github.com/ESCOMP/CTSM/pull/1182
.

### Collaborators:
@ckoven @rgknox @glemieux @huitang-earth 

Should be answer changing with fates_sp mode ON. Should not change answers with it OFF. 
n.b.  that (for now) to turn FATES_SP on, we also need to turn on fates_nocomp and fates_fixed_biogeog 

### Checklist:

### Test Results:
This runs and produces the expected results (TLAI SAI and HTOP at fixed values except for bare ground. 

CTSM (or) E3SM (specify which) test hash-tag: ctsm5.1.dev055

CTSM (or) E3SM (specify which) baseline hash-tag: ctsm5.1.dev046

FATES baseline hash-tag: sci.1.46.2_api.16.1.0

Test Output: 
`/glade/u/home/glemieux/scratch/ctsm-tests/tests_spmode-fatessuite-dev046compare`

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -1387,7 +1387,7 @@ contains
           enddo ! ends 'do while(associated(currentCohort))
           
           if ( currentPatch%total_canopy_area>currentPatch%area ) then
-             if ( currentPatch%total_canopy_area-currentPatch%area > 1.0e-10_r8 ) then
+             if ( currentPatch%total_canopy_area-currentPatch%area > 0.001_r8 ) then
                 write(fates_log(),*) 'FATES: canopy area bigger than area', &
                      currentPatch%total_canopy_area ,currentPatch%area, &
                      currentPatch%total_canopy_area -currentPatch%area,&
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -59,18 +59,18 @@ module EDCanopyStructureMod
 
   character(len=*), parameter, private :: sourcefile = &
        __FILE__
-  
+
   real(r8), parameter :: area_target_precision = 1.0E-11_r8  ! Area conservation
-                                                             ! will attempt to reduce errors
-                                                             ! below this level
-  
+  ! will attempt to reduce errors
+  ! below this level
+
   real(r8), parameter :: area_check_precision  = 1.0E-7_r8     ! Area conservation checks must 
-                                                               ! be within this absolute tolerance
+  ! be within this absolute tolerance
   real(r8), parameter :: area_check_rel_precision = 1.0E-4_r8  ! Area conservation checks must
-                                                               ! be within this relative tolerance
+  ! be within this relative tolerance
 
   real(r8), parameter :: similar_height_tol = 1.0E-3_r8    ! I think trees that differ by 1mm
-                                                           ! can be roughly considered the same right?
+  ! can be roughly considered the same right?
 
 
   ! 10/30/09: Created by Rosie Fisher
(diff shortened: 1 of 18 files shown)


main/FatesInterfaceMod.F90
```diff
@@ -1438,13 +1450,19 @@ subroutine set_fates_ctrlparms(tag,ival,rval,cval)
...
          end if
 
-        ! Future reduced complexity mode   
-        !if(hlm_use_nocomp.eq.unset_int) then
-        !      if(fates_global_verbose()) then
-        !     write(fates_log(), *) 'switch for no competition mode. '
-        !    end if
-        !   call endrun(msg=errMsg(sourcefile, __LINE__))
-        ! end if
+        if(hlm_use_nocomp.eq.unset_int) then
```
ckoven, line 1453: indenting

main/FatesInterfaceMod.F90
```diff
@@ -1438,13 +1450,19 @@ subroutine set_fates_ctrlparms(tag,ival,rval,cval)
...
+              if(fates_global_verbose()) then
+             write(fates_log(), *) 'switch for no competition mode. '
+            end if
+           call endrun(msg=errMsg(sourcefile, __LINE__))
+         end if
+
+         if(hlm_use_sp.eq.unset_int) then
```
ckoven, line 1460: indenting

main/FatesRestartInterfaceMod.F90
```diff
@@ -2579,7 +2588,7 @@ subroutine get_restart_vectors(this, nc, nsites, sites)
                       ccohort%prt%variables(i_var)%net_alloc(i_pos) = &
                             this%rvars(ir_prt_var)%r81d(io_idx_co)
 
-                      ir_prt_var = ir_prt_var + 1
+                     ir_prt_var = ir_prt_var + 1
```
ckoven, line 2591:
```suggestion
                      ir_prt_var = ir_prt_var + 1
```

parameter_files/fates_params_default.cdl
```diff
@@ -501,6 +503,9 @@ variables:
 	double fates_z0mr(fates_pft) ;
 		fates_z0mr:units = "unitless" ;
 		fates_z0mr:long_name = "Ratio of momentum roughness length to canopy top height" ;
+        double fates_hlm_pft_map(fates_hlm_pftno, fates_pft) ;
```
ckoven, line 506: indenting

parameter_files/fates_params_default.cdl
```diff
@@ -12,6 +12,8 @@ dimensions:
 	fates_prt_organs = 6 ;
 	fates_string_length = 60 ;
 	fates_variants = 2 ;
+        fates_hlm_pftno = 14 ;
```
ckoven, line 15: this forces an update to the trait sorting script.  i'll make a PR to your PR branch to do that.

ckoven (approved):
@rosiealice this all looks really great!  I made a bunch of minor comments, almost all of which are asking for either more standardization of indenting around the many new if clauses or a bit more documentation of the logic in a few places.  sorry about the indentation nitpickiness, I've been using python long enough now that I tend to actually think it means something, but in any case emacs should be able to take care of it all in about a second.

huitang-earth (commented):
Hi Rosie,
I just finished half of the code review. So far, there is only one place I suspect to be a bug.
Line 1416-1421,  EDPhysiologyMod.F90
Please check the detailed comments there. 
I will continue looking at the rest of the code tomorrow.
Best, Hui

main/FatesInterfaceTypesMod.F90
```diff
@@ -331,6 +336,10 @@ module FatesInterfaceTypesMod
                                                  ! be equivalent (ie integer ascending)
                                                  ! Or, all will be 1.
       
+     integer,allocatable   :: sp_patch_index(:)  ! in SP mode, we need to map the p values for each patch
```
huitang-earth, line 339: I do not find any usage of this variable.....

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1346,13 +1344,33 @@ subroutine canopy_summarization( nsites, sites, bc_in )
...
+
+             if(hlm_use_sp.eq.itrue)then
+
+               if(associated(currentPatch%tallest%shorter))then
+                 write(fates_log(),*) 'more than one cohort in SP mode',s,currentPatch%nocomp_pft_label
+                 call endrun(msg=errMsg(sourcefile, __LINE__))
+               end if
+
+               if(currentPatch%total_canopy_area-currentPatch%area.gt.1.0e-16)then
+                 write(fates_log(),*) 'too much canopy in summary',s,currentPatch%total_canopy_area-currentPatch%area
```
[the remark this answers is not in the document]
rosiealice: OK done...

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1330,6 +1336,238 @@ subroutine phenology_leafonoff(currentSite)
...
+
+    !------------------------------------------
+    !  Calculate dbh from input height, and c_area from dbh
+    !------------------------------------------
+    currentCohort%hite = htop
+
+    fates_pft = currentCohort%pft
+    call h2d_allom(currentCohort%hite,fates_pft,currentCohort%dbh)
+
+   currentCohort%n = 1.0_r8 ! make n=1 to get area of one tree.
```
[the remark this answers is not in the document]
rosiealice: Indeed I could...

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1902,8 +1921,9 @@ subroutine update_hlm_dynamics(nsites,sites,fcolumn,bc_out)
         currentPatch => sites(s)%oldest_patch
         c = fcolumn(s)
         do while(associated(currentPatch))
+          if(currentPatch%nocomp_pft_label.ne.0)then ! only increase ifp for veg patches not BG (in SP mode)
```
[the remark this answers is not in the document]
rosiealice: yep...

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2023,6 +2036,9 @@ subroutine zero_patch(cp_p)
     currentPatch%gnd_alb_dir(:)             = nan
     currentPatch%gnd_alb_dif(:)             = nan
 
+   ! special modes
```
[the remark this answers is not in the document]
rosiealice: This code seems to have gone away now..

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1330,6 +1336,238 @@ subroutine phenology_leafonoff(currentSite)
...
+      currentCohort => currentCohort%shorter
+    end do !cohort loop
+    currentPatch => currentPatch%younger
+  end do ! patch loop
+
+  end subroutine satellite_phenology
+
+! =====================================================================================
+
+  subroutine assign_cohort_SP_properties(currentCohort,htop,tlai,tsai,parea,init,leaf_c)
```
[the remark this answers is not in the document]
rosiealice: As I recall the reason it is there is because a value of c_leaf is needed to bt used in init_cohorts. Does that make sense?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1330,6 +1336,238 @@ subroutine phenology_leafonoff(currentSite)
...
+        write(fates_log(),*) 'error in validate treelai',currentCohort%treelai,check_treelai,currentCohort%treelai-check_treelai
+        call endrun(msg=errMsg(sourcefile, __LINE__))
+      end if
+      
+     ! the carea_allom routine sometimes generates precision-tolerance level errors in the canopy area 
+     ! these mean that the canopy area does not exactly add up to the patch area, which causes chaos in 
+     ! the radiation routines.  Correct both the area and the 'n' to remove error, and don't use 
+    !! carea_allom in SP mode after this point. 
+
+     if(abs(currentCohort%c_area-parea).gt.nearzero)then ! there is an error
```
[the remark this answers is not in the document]
rosiealice: Good question. I think the answer here is that the minute errors arise in the calculation of c_area from dbh, and we really want c_area, AND leaf_c ro be consistant with dbh. 
Either way, we are talking about precision-level errors here rather than ones with biological meaning.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1330,6 +1336,238 @@ subroutine phenology_leafonoff(currentSite)
...
+   do while (associated(currentPatch))
+
+     fates_pft = currentPatch%nocomp_pft_label
+     if(fates_pft.ne.0)then 
+
+       sumarea = 0.0_r8
+       do hlm_pft = 1,size( EDPftvarcon_inst%hlm_pft_map,2)
+
+         if(bc_in%pft_areafrac(hlm_pft) * EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft).gt.0.0_r8)then
+            sumarea = sumarea + bc_in%pft_areafrac(hlm_pft)*EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft)
```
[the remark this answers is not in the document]
rosiealice: good catch! I removed it...

biogeochem/FatesAllometryMod.F90
```diff
@@ -756,6 +757,98 @@ real(r8) function tree_sai( pft, dbh, canopy_trim, c_area, nplant, cl, &
...
+ 
+    if(cl>1)then
+      write(fates_log(),*) 'in sub-canopy layer in leafc_from_treelai'
+      write(fates_log(),*) 'this is not set up to work for lower canopy layers.'
+      write(fates_log(),*) 'problem in leafc_from_treelai',cl,pft
+      call endrun(msg=errMsg(sourcefile, __LINE__))
+    endif
+ 
+    slat = g_per_kg * prt_params%slatop(pft)
+    leafc_per_unitarea = leaf_c/(c_area/nplant) !KgC/m2
```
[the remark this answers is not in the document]
rosiealice: Indeed. I think it was a hangover from my copying and psting the treelai routine. Not sure why is wasn't creating an error given I hadn't actually defined a value for leaf_c by then.

main/EDInitMod.F90
```diff
@@ -298,14 +313,67 @@ subroutine set_site_properties( nsites, sites,bc_in )
...
+            do hlm_pft = 1,size( EDPftvarcon_inst%hlm_pft_map,2)
+               do fates_pft = 1,numpft ! loop round all fates pfts for all hlm pfts        
+                 sites(s)%area_pft(fates_pft) = sites(s)%area_pft(fates_pft) + &
+                     EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft) * bc_in(s)%pft_areafrac(hlm_pft) 
+               end do
+            end do !hlm_pft
+
+            sumarea = sum(sites(s)%area_pft(1:numpft))
+            do ft =  1,numpft
+              if(sites(s)%area_pft(ft).lt.0.01_r8.and.sites(s)%area_pft(ft).gt.0.0_r8)then
```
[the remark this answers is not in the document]
rosiealice: OK I put them in the same loop...

main/FatesInterfaceTypesMod.F90
rosiealice: It was a remnant of a previous attempt to solve the SP mode indexing issues. Luckily it can now be deleted. I think/hope :)

main/EDPftvarcon.F90
```diff
@@ -1667,7 +1693,20 @@ subroutine FatesCheckParams(is_master)
 
         end if
 
-     end do
+        ! check that the host-fates PFT map adds to one in both dimension
```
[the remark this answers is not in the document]
rosiealice: Indeed, they don't actually need to add to one alone the FATES PFT dimension, only the HLM PFT dimension. Changed the comment to reflect that.

main/EDInitMod.F90
```diff
@@ -389,51 +464,138 @@ subroutine init_patches( nsites, sites, bc_in)
...
+         end do
+
+         if(abs(tota-area).gt.nearzero*area)then
+           if(abs(tota-area).lt.1.0e-10_r8)then ! this is a precision error
+             if(sites(s)%oldest_patch%area.gt.(tota-area+nearzero))then 
+               ! remove or add extra area from bare ground patch
+                sites(s)%oldest_patch%area = sites(s)%oldest_patch%area - (tota-area)
+                write(*,*) 'fixing patch precision O',s, tota-area
+              else
+                sites(s)%youngest_patch%area = sites(s)%oldest_patch%area - (tota-area)
```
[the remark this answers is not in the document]
rosiealice: I added some mode details to the error check commnets. Sometimes you can't fix the area with the oldest patch as it is itself too small, so have to revert to the youngest patch...

main/EDMainMod.F90
```diff
@@ -276,15 +300,16 @@ subroutine ed_ecosystem_dynamics(currentSite, bc_in, bc_out)
        end if
     end if
 
+    ! SP has changes in leaf carbon but we don't expect them to be in balance.
```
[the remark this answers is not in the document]
rosiealice: At the moment, we can't run the host LM in SP mode because it thinks that FATES and SP mode are inconsistant (because FATES depends upon using 'use_cn' which is the opposite flag to 'use_sp'.  i got around this by using the code that allows some SP functionality -in- SP mode for the sake of n depositioin. However, we are currently making plans to break up the meaning of the 'use_cn' flag in CLM as it is an artifcact of past code and somewhat nonsensical under the present array of model configurations.   

That's all a bigger fish than this PR will deal with though. 

I am not, at least, triggering any balance check errors in the HLM. Maybe no carbon is getting through at all?

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1947,7 +1967,7 @@ end subroutine mortality_litter_fluxes
 
   ! ============================================================================
 
-  subroutine create_patch(currentSite, new_patch, age, areap, label)
+  subroutine create_patch(currentSite, new_patch, age, areap, label,nocomp_pft)
```
rgknox, line 1970: I noticed that none of the calls to create_patch() initialize the nocomp_pft with anything other than unset_int.  Is it needed as an argument here then?

rgknox: nevermind, forgot to expand the github diffs for the files with lots of changes!

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -150,7 +150,7 @@ subroutine canopy_structure( currentSite , bc_in )
       
 
       !----------------------------------------------------------------------
-
+    if(hlm_use_sp.eq.ifalse)then
```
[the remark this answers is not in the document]
rosiealice: I moved this to outside the canopy_structure routine.

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1330,10 +1328,10 @@ subroutine canopy_summarization( nsites, sites, bc_in )
              call coagetype_class_index(currentCohort%coage,currentCohort%pft, &
                   currentCohort%coage_class,currentCohort%coage_by_pft_class)
           end if
-          
+          if(hlm_use_sp.eq.ifalse)then
```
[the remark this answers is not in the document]
rosiealice: indented all of this file...

biogeochem/EDCanopyStructureMod.F90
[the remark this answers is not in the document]
rosiealice: Indented all of file in https://github.com/rosiealice/fates/commit/6a4dec9aa1bf4d61e301379e143f0b4311f5e77f

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -2004,8 +2026,14 @@ subroutine update_hlm_dynamics(nsites,sites,fcolumn,bc_out)
            currentPatch => sites(s)%oldest_patch
            ifp = 0
            do while(associated(currentPatch))
+            if(currentPatch%nocomp_pft_label.ne.0)then ! for vegetated patches only
```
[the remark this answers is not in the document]
rosiealice: Indented all of this file in https://github.com/rosiealice/fates/commit/6a4dec9aa1bf4d61e301379e143f0b4311f5e77f

main/EDInitMod.F90
```diff
@@ -298,14 +313,67 @@ subroutine set_site_properties( nsites, sites,bc_in )
...
+
+            sites(s)%area_pft(1:numpft) = 0._r8
+            do hlm_pft = 1,size( EDPftvarcon_inst%hlm_pft_map,2)
+               do fates_pft = 1,numpft ! loop round all fates pfts for all hlm pfts        
+                 sites(s)%area_pft(fates_pft) = sites(s)%area_pft(fates_pft) + &
+                     EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft) * bc_in(s)%pft_areafrac(hlm_pft) 
+               end do
+            end do !hlm_pft
+
+            sumarea = sum(sites(s)%area_pft(1:numpft))
```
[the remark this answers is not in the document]
rosiealice: OK I fixed that there...

main/EDInitMod.F90
```diff
@@ -298,14 +313,67 @@ subroutine set_site_properties( nsites, sites,bc_in )
...
+                else
+                  sites(s)%area_pft(ft) = area/numpft 
+                  ! in nocomp mode where there is only bare ground, we assign equal area to 
+                  ! all pfts and let the model figure out whether land should be bare or not. 
+                end if
+               end do !ft
+              else ! for sp mode, assert a bare ground patch
+                sumarea = sum(sites(s)%area_pft(1:numpft)) 
+
+                if(sumarea.lt.area)then !make some bare ground
```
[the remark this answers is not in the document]
rosiealice: It's because in all the other FATES modes, bareground is the area in which trees do not grow of their own accord. In SP mode, because we are asserting that there is a full canopy in each of the PFT patches, we also need to assert a bare ground area in order to not have all of the ground filled by leaves.

Further to that, one could imagine having bare ground being calculated as the remaining area when all fhe canopies are accounted for, but this means we don't pass balance check on canopy are inside FATES, and so in SP mode, we define the bare groud/no cohorts patch as having a PFT identifier as zero.  

I added comments to this effect to help people understand this a bit more.

main/EDInitMod.F90
```diff
@@ -389,51 +464,138 @@ subroutine init_patches( nsites, sites, bc_in)
...
+           end if 
+
+           if(hlm_use_nocomp.eq.itrue)then 
+              ! In no competition mode, if we are using the fixed_biogeog filter 
+              ! then each PFT has the area dictated  by the surface dataset.
+              ! If not, each PFT gets the same area. 
+              if(hlm_use_fixed_biogeog.eq.itrue)then
+                 newparea = sites(s)%area_pft(nocomp_pft)
+              else
+                 newparea = area / numpft
```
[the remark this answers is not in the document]
rosiealice: Added:
                ! If we are not using fixed biogeog model, each PFT gets the same area.                                  
      	      	! i.e. each grid cell is divided exactly into the number of FATES PFTs

biogeochem/FatesAllometryMod.F90
```diff
@@ -756,6 +757,98 @@ real(r8) function tree_sai( pft, dbh, canopy_trim, c_area, nplant, cl, &
     return
   end function tree_sai
   
+! =====================================================================================
+ 
+  real(r8) function leafc_from_treelai( treelai, pft, c_area, nplant, cl, vcmax25top)
+ 
+    ! -----------------------------------------------------------------------------------
+    ! LAI of individual trees is a function of the total leaf area and the total
```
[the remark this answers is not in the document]
rosiealice: Indeed. Changing it to something better.

biogeophys/EDSurfaceAlbedoMod.F90
```diff
@@ -89,6 +89,7 @@ subroutine ED_Norman_Radiation (nsites, sites, bc_in, bc_out )
            ifp = 0
            currentpatch => sites(s)%oldest_patch
            do while (associated(currentpatch))  
+             if(currentpatch%nocomp_pft_label.ne.0)then
```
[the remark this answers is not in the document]
rosiealice: added this:
              ! do not do albedo calculations for bare ground patch in SP mode                                           
      	      ! and (more impotantly) do not iterate ifp or it will mess up the indexing wherein                         
      	      ! ifp=1 is the first vegetated patch.

biogeophys/EDAccumulateFluxesMod.F90
```diff
@@ -64,6 +64,7 @@ subroutine AccumulateFluxes_ED(nsites, sites, bc_in, bc_out, dt_time)
 
        cpatch => sites(s)%oldest_patch
        do while (associated(cpatch))                 
+         if(cpatch%nocomp_pft_label.ne.0)then
```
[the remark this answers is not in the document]
rosiealice: done.

biogeophys/EDBtranMod.F90
```diff
@@ -133,6 +133,7 @@ subroutine btran_ed( nsites, sites, bc_in, bc_out)
            ifp = 0
            cpatch => sites(s)%oldest_patch
            do while (associated(cpatch))                 
+             if(cpatch%nocomp_pft_label.ne.0)then ! only for veg patches
```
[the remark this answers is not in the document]
rosiealice: done.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -2260,7 +2498,7 @@ real(r8) :: catanf_30             ! hyperbolic temperature function from CENTURY
     catanf_30 = catanf(30._r8)
     
     ifp = currentPatch%patchno 
-
+    if(currentPatch%nocomp_pft_label.gt.0)then
```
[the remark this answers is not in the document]
rosiealice: done

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -2332,6 +2332,7 @@ subroutine hydraulics_bc ( nsites, sites, bc_in, bc_out, dtime)
        ifp = 0
        cpatch => sites(s)%oldest_patch
        do while (associated(cpatch))
+         if(cpatch%nocomp_pft_label.ne.0)then
```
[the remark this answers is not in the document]
rosiealice: done.

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -279,7 +278,7 @@ real(r8) :: tcwood             ! Temperature response function for wood
          ifp = 0
          currentpatch => sites(s)%oldest_patch
          do while (associated(currentpatch))  
-
+            if(currentpatch%nocomp_pft_label.ne.0)then
```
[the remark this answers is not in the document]
rosiealice: done.

main/EDInitMod.F90
```diff
@@ -487,38 +651,80 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
+          temp_cohort%n           =  temp_cohort%n * sum(site_in%use_this_pft)
+       endif
 
        temp_cohort%canopy_trim = 1.0_r8
 
+       !  h,dbh,leafc,n from SP values or from small initial size.
+
+       if(hlm_use_sp.eq.itrue)then
+         init = itrue
+         call assign_cohort_SP_properties(temp_cohort, 0.5_r8,0.2_r8, 0.1_r8,patch_in%area,init,c_leaf)
```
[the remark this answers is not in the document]
rosiealice: At this point, we do not know the  bc_in values of tlai tsai and htop, so this is initializing to an arbitrary value for the very first timestep. Not sure if there's a way around this or not.

main/EDMainMod.F90
```diff
@@ -637,12 +662,15 @@ subroutine ed_update_site( currentSite, bc_in, bc_out )
     ! !LOCAL VARIABLES:
     type (ed_patch_type) , pointer :: currentPatch   
     !-----------------------------------------------------------------------
-
-    call canopy_spread(currentSite)
+    if(hlm_use_sp.eq.ifalse)then
```
ckoven: @rosiealice as I read this bit, by turning off the `canopy_spread` code, this locks in the initial value, which is being set by `init_spread_near_bare_ground` to 1.0 in EDTypesMod.F90.  Not sure whether or not the value of spread matters for SP mode, but likely will init and hold it at 0.0 for bigleaf mode.

rosiealice: I think that's fine. We should probably make sure we lock in the same value in both cases.  I don't think it matters what the value is for SP mode. 
 
I am setting it to 1.0 in assign_cohort_sp_properties but maybe I should actually use the patch spread and make sure it is invariant instead?

https://github.com/rosiealice/fates/blob/c49dc43e2427918eb4d100b6884859aa4cecee44/biogeochem/EDPhysiologyMod.F90#L1507

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1330,6 +1336,238 @@ subroutine phenology_leafonoff(currentSite)
...
+           currentSite%sp_htop(fates_pft) = currentSite%sp_htop(fates_pft) + &
+                 bc_in%hlm_sp_htop(hlm_pft) * bc_in%pft_areafrac(hlm_pft) &
+               * EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft)
+         end if ! there is some area in this patch
+       end do !hlm_pft
+
+       ! weight for total area in each patch/fates_pft
+       if(currentPatch%area.gt.0.0_r8)then 
+         currentSite%sp_tlai(fates_pft) = currentSite%sp_tlai(fates_pft) &
+             /(currentPatch%area/area)
```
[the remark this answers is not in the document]
rosiealice: It's because the areas by which the lai values etc are multiplied in the previous section don't necessarily add up to one.  

tlai  = sum( sp_tlai(hlmpft) * pftareafrac(hlmpft) * hlm_pft_map(fates_pft,hlm_pft)
          
say that a given fates PFT's patch is made up 50% of the grid cell, which was itself made up from 30% of one pft and 20% from another. The input lais are both 10, that would give
10*0.3*1. + 10*0.2+1.  = 2+3 = 5

the right answer here is 10, (for the weighted average lai) and so we need ti divide by 5000/10000.= 0.5 (or x2) 

To check this I just did a simulation where I commented out these lines and the LAI was too small compared with the CLM5-SP run LAI (more so in areas with mixed pfts like savanna and W Europe). 

Does that make sense.  I added another comment to help out here.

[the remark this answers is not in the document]
huitang-earth: Thanks for the clarification. From your example, I now understand that "currentPatch%area/area" actually equal to "sumarea". I suppose "currentPatch%area" has been defined according to "bc_in%pft_areafrac(hlm_pft)" and "EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft)" in SP mode, right?  :)

huitang-earth (approved):

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -2026,146 +2056,151 @@ end subroutine update_hlm_dynamics
...
+          end if
+          currentCohort => currentCohort%shorter
+       enddo
+
+       ! Does the bottom layer have more than a full canopy? 
+       ! If so we need to make another layer.
+       if(arealayer > currentPatch%area)then
+          z = z + 1
+          if(hlm_use_sp)then
+             write(fates_log(),*) 'SPmode, canopy_layer full:',arealayer,currentPatch%area
```
rgknox, line 2198: @rosiealice , is this an error? If so, should we abort?

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1851,173 +1871,183 @@ subroutine leaf_area_profile( currentSite , snow_depth_si, frac_sno_eff_si)
...
+          end if
+
+          if(debug) then
+             write(fates_log(),*) 'imprecise patch areas in update_hlm_dynamics',total_patch_area
+          end if
+
+          currentPatch => sites(s)%oldest_patch
+          ifp = 0
+          do while(associated(currentPatch))
+             if(currentPatch%nocomp_pft_label.ne.0)then ! for vegetated patches only
```
rgknox, line 2031: maybe we should give "0" a special named parameter constant value

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1272,6 +1276,22 @@ subroutine set_patchno( currentSite )
        currentPatch => currentPatch%younger
     enddo
 
+    if(hlm_use_sp)then
```
rgknox, line 1279: perhaps assign the non-sp type of patch number as an else condition so that it is not caculated unecessarily when not in sp mode.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -2321,14 +2342,20 @@ subroutine fuse_patches( csite, bc_in )
                             endif ! sum(biomass(:,:) .gt. force_patchfuse_min_biomass 
                          endif ! maxage
 
+
+                         ! Do not fuse patches that have different PFT labels in nocomp mode
```
rgknox, line 2346: should we have this fusion exclusion prior to the the block above?  It seems there is no reason to calculate and compare age profiles if the patches are in nocomp and have different pfts

https://github.com/NGEET/fates/pull/703/files?diff=unified&w=1#diff-28c679efbe804905b8b4fac3ccdbb04f6bb3bcd2794fe1a4556d5541e008a750R2289-R2343

rgknox: also, this whole routine would benefit from using named "if" "do" and "while"s.  Its starting to get hard to determine where things are opening and closing.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1118,122 +1124,125 @@ subroutine phenology_leafonoff(currentSite)
...
+                if (currentCohort%status_coh == leaves_off)then    
 
                    !is it the leaf-on day? Are the leaves currently off?    
 
-                  currentCohort%status_coh = leaves_on    ! Leaves are on, so change status to 
-                                                           ! stop flow of carbon out of bstore. 
+                   currentCohort%status_coh = leaves_on    ! Leaves are on, so change status to 
+                   ! stop flow of carbon out of bstore. 
+
+                   if(store_c>nearzero) then
```
rgknox, line 1243: This will overwrite a fix in a previous PR.

rgknox: #684

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1245,92 +1254,322 @@ subroutine phenology_leafonoff(currentSite)
...
+    ! hlm_pft_map is the area of that land in each FATES PFT (from param file)
+
+    ! 2. weight each fates PFT target for lai, sai and htop by the area of the 
+    ! contrbuting HLM PFTs.
+
+    currentPatch => currentSite%oldest_patch
+    do while (associated(currentPatch))
+
+       fates_pft = currentPatch%nocomp_pft_label
+       if(fates_pft.ne.0)then
```
rgknox, line 1395: This is impossible in sp right? I see a check on it below.  Can we remove this clause then?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1245,92 +1254,322 @@ subroutine phenology_leafonoff(currentSite)
...
+          oldcarea = currentCohort%c_area
+          !generate new cohort area
+          currentCohort%c_area = currentCohort%c_area - (currentCohort%c_area- parea)
+          currentCohort%n = currentCohort%n * (currentCohort%c_area/oldcarea)
+          if(abs(currentCohort%c_area-parea).gt.nearzero)then
+             write(fates_log(),*) 'SPassign, c_area still broken',currentCohort%c_area-parea,currentCohort%c_area-oldcarea
+             call endrun(msg=errMsg(sourcefile, __LINE__))
+          end if
+       else 
+          write(fates_log(),*) 'SPassign, big error in c_area',currentCohort%c_area-parea,currentCohort%pft
```
rgknox, line 1561: worried about logging that is not followed by an endrun.  Should we encapsulate this with a local debug flag that is default set to false?

main/EDTypesMod.F90
```diff
@@ -410,7 +410,7 @@ module EDTypesMod
      integer  ::  ncl_p                                            ! Number of occupied canopy layers
      integer  ::  anthro_disturbance_label                         ! patch label for anthropogenic disturbance classification
      real(r8) ::  age_since_anthro_disturbance                     ! average age for secondary forest since last anthropogenic disturbance
-
+     integer  ::  nocomp_pft_label                                 ! where nocomp is active, use this label for patch ID.
```
rgknox: add text here that describes the special case where = 0
Also, short description of how labels are determined.

main/FatesInterfaceMod.F90
```diff
@@ -297,7 +297,8 @@ subroutine zero_bcs(fates,s)
     fates%bc_out(s)%displa_pa(:) = 0.0_r8
     fates%bc_out(s)%z0m_pa(:)    = 0.0_r8
     fates%bc_out(s)%dleaf_pa(:)   = 0.0_r8
-    
+    fates%bc_out(s)%nocomp_pft_label_pa(:) = 0
```
rgknox: should this be initialized as an undefined?  0 is a special case right?  It might get confusing at some point of we use this to set the label and not have it set in an expected location.

rgknox:
@rosiealice, the setting of nocomp_pft gets a little confusing for me.
Sometimes we set it to zero, sometimes it seems to be set as unset_int, its also zero'd during the patch zero'ing process.  I think when no-comp is turned on, this should be set to the sole pft of the cohorts inabiting.
If it makes sense, could we create some global parameter constants that define the special modes (ie non no-comp) and pass those in as arguments where appropriate, instead of using 0 and unset_int?

biogeochem/EDPatchDynamicsMod.F90
rgknox: examples:
open: https://github.com/NGEET/fates/blob/master/parteh/PRTAllometricCNPMod.F90#L1399
close: https://github.com/NGEET/fates/blob/master/parteh/PRTAllometricCNPMod.F90#L1534

main/FatesInterfaceMod.F90
```diff
@@ -476,6 +478,12 @@ subroutine allocate_bcin(bc_in, nlevsoil_in, nlevdecomp_in, num_lu_harvest_cats)
 
       allocate(bc_in%pft_areafrac(maxpft))
```
huitang-earth: After testing the code myself, I noticed that this better to be `allocate(bc_in%pft_areafrac(**0:**maxpft))`. Then, it will be consistent with the indexing in the code in "https://github.com/rosiealice/ctsm/blob/SP_api/src/utils/clmfates_interfaceMod.F90" Line 630-639. The mismatch of the indexing can cause segmentation fault without showing where the problem happens (when DEGUB=false).

main/EDInitMod.F90
```diff
@@ -278,175 +293,334 @@ subroutine set_site_properties( nsites, sites,bc_in )
...
+
+    ! This sets the rhizosphere shells based on the plant initialization
+    ! The initialization of the plant-relevant hydraulics variables
+    ! were set from a call inside of the init_cohorts()->create_cohort() subroutine
+    if (hlm_use_planthydro.eq.itrue) then 
+       do s = 1, nsites
+          sitep => sites(s)
+          call updateSizeDepRhizHydProps(sitep, bc_in(s))
+       end do
+       deallocate(recall_older_patch)
```
huitang-earth: `deallocate(recall_older_patch)`
This line should be removed. This line is not existing in "https://github.com/NGEET/fates/blob/master/main/EDInitMod.F90". 

Deallocate will remove the patch that has been created, and cause problems in later steps.  In my case, it complains: 

forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable FRAGMENTATION_SCALER when it is not allocated

Image              PC                Routine            Line        Source
cesm.exe           000000000408FCA6  Unknown               Unknown  Unknown
cesm.exe           0000000001765F68  fateshistoryinter        2776  FatesHistoryInterfaceMod.F90
cesm.exe           00000000009C3387  clmfatesinterface        1581  clmfates_interfaceMod.F90
cesm.exe           0000000000913EB1  clm_initializemod         746  clm_initializeMod.F90
cesm.exe           0000000000891C4D  lnd_comp_mct_mp_l         234  lnd_comp_mct.F90
cesm.exe           000000000046605F  component_mod_mp_         257  component_mod.F90
cesm.exe           0000000000427AC2  cime_comp_mod_mp_        1350  cime_comp_mod.F90
cesm.exe           000000000045CC59  MAIN__                    122  cime_driver.F90
cesm.exe           0000000000415252  Unknown               Unknown  Unknown
libc-2.17.so       00002AAF381B2505  __libc_start_main     Unknown  Unknown
cesm.exe           0000000000415169  Unknown               Unknown  Unknown

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1493,167 +1514,166 @@ subroutine leaf_area_profile( currentSite , snow_depth_si, frac_sno_eff_si)
...
-                                           currentPatch%canopy_layer_tlai,currentCohort%vcmax25top )    
+             currentCohort%treelai = tree_lai(leaf_c, currentCohort%pft, currentCohort%c_area, &
+                  currentCohort%n, currentCohort%canopy_layer,               &
+                  currentPatch%canopy_layer_tlai,currentCohort%vcmax25top )    
 
-          currentCohort%treesai = tree_sai(currentCohort%pft, currentCohort%dbh, currentCohort%canopy_trim, &
-                                           currentCohort%c_area, currentCohort%n, currentCohort%canopy_layer, &
-                                           currentPatch%canopy_layer_tlai, currentCohort%treelai , &
-                                           currentCohort%vcmax25top,4)  
+             currentCohort%treesai = tree_sai(currentCohort%pft, currentCohort%dbh, currentCohort%canopy_trim, &
```
huitang-earth, line 1537: For SP model, should "treesai" be prescribed rather than being calculated from allometric function?  This will lead to very small sai value.

main/EDInitMod.F90
```diff
@@ -487,171 +662,209 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
+             temp_cohort%canopy_trim = 1.0_r8
+
+             !  h,dbh,leafc,n from SP values or from small initial size.
+
+             if(hlm_use_sp.eq.itrue)then
+                init = itrue
+                ! At this point, we do not know the bc_in values of tlai tsai and htop, 
+                ! so this is initializing to an arbitrary value for the very first timestep. 
+                ! Not sure if there's a way around this or not.
+                call assign_cohort_SP_properties(temp_cohort, 0.5_r8,0.2_r8, 0.1_r8,patch_in%area,init,c_leaf)
```
huitang-earth: This is indeed problematic. My test run shows that the arbitrary value of LAI and SAI at the first step, will give rise to balance error ("ERROR: ERROR in BalanceCheckMod.F90 at line 654") in the later time steps when the new LAI and SAI are read from satellite phenology. My solution to the initialization of SP mode is to add  satellite phenology in "clm_initializeMod.F90" before calling FATES initialization. Also, we need to add subroutine "satellite_phenology" into subroutine "wrap_update_hlmfates_dyn" in "CLMFatesInterfaceMod.F90". Then, FATES will be initializaed with proper values of LAI and SAI at the very beginning.

main/FatesInterfaceMod.F90
glemieux: Good catch @huitang-earth.  This is due to the hlm side code where the index for `pft_areafrac` is set to always index from zero regardless of the setting of `natpft_lb`: https://github.com/glemieux/ctsm/blob/6323556f1d1a4d4b9b40be993b814f7db2574de7/src/utils/clmfates_interfaceMod.F90#L630-L632

I also noticed that the `hlm_sp_` prefixed variables have the same indexing, so I changed those allocations to match as well: https://github.com/glemieux/fates/blob/3ac64584b5442791a16b119b265d827bad35d11f/main/FatesInterfaceMod.F90#L482-L486

glemieux: Actually I think I'm going to reverse course on this: I just double-checked and noticed that in EDInitMod and EDPhysiologyMod that the `pft_areafrac` and `hlm_pft_map` are using the same indexing variable which starts at index one to conform to the `hlm_pft_map` allocation: https://github.com/rosiealice/fates/blob/c7e4df8e01fe66a2dacd89527849c7c1d9a4629b/main/EDInitMod.F90#L324-L327

glemieux:
Running fates suite tests, after the most recent commits, `compare_base_rest` is failing.  Also @rosiealice is seeing an overall reduction in the TLAI output and believes the issue might be due to the indexing update:

To do List:
- [ ] Determine where compare_base_rest failure first appeared in regression test history
- [ ] Revert indexing updates to determine if it is the cause of TLAI reduction

rosiealice:
FWIW, these changes:
https://github.com/rosiealice/fates/tree/IFP_issue

make the differences in LAI go away, and I think should not try and set any output variables when IFP=0

glemieux:
Thanks to help from @rgknox the `COMPARE_base_rest` issue on the regression tests appears to be cleared up with e9a5da90fb74b1db8993b7b35cf2589c20d6e2e1.  The following errors are still TBR:

- [ ] `pft_areafrac` sum to 1.0 check fails in `ERP_D_P32x2_Ld3.f19_g17.I2000Clm50FatesCru.cheyenne_intel.clm-FatesColdDef`
- [ ] `currentPatch%disturbance_mode` check fails in `ERS_Ld60.f45_f45_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-FatesST3`
- [ ] `fragmentation_scaler` not allocated in `ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCru.cheyenne_intel`
- [ ] Unidentified RUN failure  `SMS_Lm6.f45_f45_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-Fates`

I believe the first of the above is due to the summation not being handled correctly for multiple threads.  Others are still under review.

glemieux:
> FWIW, these changes:
> https://github.com/rosiealice/fates/tree/IFP_issue
> 
> make the differences in LAI go away, and I think should not try and set any output variables when IFP=0

@rosiealice I'm thinking I'll manually incorporate those changes on

glemieux:
After reviewing the failures with @ckoven, we noted that the `fragmentation_scaler` failure and the unidentified RUN failure are only present for `FatesHydro` testmods.  In the first instance a patch does not seem to be allocated and is causing the RUN to fail during `init_coldstart`.

glemieux:
Removing the deallocation of the `recall_older_patch` with d901f9f36450d5b7f19dca257a668e865e5eb6dd as noted by  @huitang-earth [here](https://github.com/NGEET/fates/pull/703#discussion_r557441834) clears the fragmentation and unidentified errors.

glemieux:
The `pft_areafrac` check error for `ERP_D_P32x2_Ld3` was due to the summation occurring on a thread prior to the completion of a given site-level do loop.  Adding `ft` to the `PRIVATE` clause alleviated the error (see https://github.com/ESCOMP/CTSM/pull/1182/commits/6361c71eeaab75fdb4c5397bf078a4bcae4500a2).  I double checked that the outputs were summing to `1` as expected.

rgknox:
For some reason the carriage return is now showing up on this line:
https://github.com/NGEET/fates/blob/c7e4df8e01fe66a2dacd89527849c7c1d9a4629b/main/EDPftvarcon.F90#L632

rgknox:
@glemieux , I noticed that for other 2D parameters, when we register we pass in the lower bounds for both dimensions. Here is an example:

https://github.com/NGEET/fates/blob/master/parteh/PRTParamsFATESMod.F90#L589

For the hlm to fates pft parameter map, can you confirm that this second dimension is not required? (I believe that it is set at 1 in most cases, so maybe it is optional to have the second dimension in that argument?)

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1242,92 +1251,322 @@ subroutine phenology_leafonoff(currentSite)
...
+       write(fates_log(),*) "SP mode >1 cohort: CL",currentCohort%canopy_layer, currentCohort%shorter%canopy_layer
+       call endrun(msg=errMsg(sourcefile, __LINE__))
+    end if
+
+    !------------------------------------------
+    !  Calculate dbh from input height, and c_area from dbh
+    !------------------------------------------
+    currentCohort%hite = htop
+
+    fates_pft = currentCohort%pft
```
rgknox, line 1505: this variable is sometimes used sometimes not, I recommend removing this and just using the original variable

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1242,92 +1251,322 @@ subroutine phenology_leafonoff(currentSite)
...
+             write(fates_log(),*) 'SPassign, c_area still broken',currentCohort%c_area-parea,currentCohort%c_area-oldcarea
+             call endrun(msg=errMsg(sourcefile, __LINE__))
+          end if
+       else 
+          write(fates_log(),*) 'SPassign, big error in c_area',currentCohort%c_area-parea,currentCohort%pft
+       end if ! still broken
+    end if !small error
+
+    if(init.eq.ifalse)then
+       call SetState(currentCohort%prt,leaf_organ,1,leaf_c,1)
```
rgknox, line 1563: Do we have a check somewhere that gracefully complains and fails if someone tries to run SP with fates_parteh_mode > 1?  I'm realizing that SP is not compatable with N or P.

rosiealice: No, we don't. I think it might not actually break if you turned both things on, but you wouldn't interact with any of the nutrient cycling so it likely wouldn't be doing what people intended....

glemieux:
Merged up to sci.1.46.2_api.16.1.0 in coordination with [merging the ctsm side up to dev tag 046](https://github.com/ESCOMP/CTSM/pull/1182#issuecomment-898187884).  All fates suite test PASS b4b expect for two known failures and one unexpected failure for the FixedBioGeo testmod.  After talking with Ryan, we ultimately determined that this failure mode has more to do with the original regression test results being broken for the following reasons:

- The master branch assumed a 1:1 match of hlm pfts to fates pft.  This PR introduces the use of  `hlm_pft_map` from the parameter file in associating more the pfts between the too more approriately.
- There was a bug in the master branch on the ctsm-side that was [setting the bareground index to the first index](https://github.com/ESCOMP/CTSM/blob/3215d575a136262504520fa6a8da4fe5319a657e/src/utils/clmfates_interfaceMod.F90#L621-L624) of `pft_areafrac`.  This PR now corrects that indexing (and the allocation of the `pft_areafrac` accordingly).

As such, I think that the results for the fixedbiogeo testmod are actually more correct and the FAIL comparison is expected.

I have a few more items to clean up before merging the ctsm-side up to the most recent dev tag.  These should not affect the result of the test, but I will re-run the regression tests after coming up-to-date with the most recent ctsm dev tag:

- [ ] Address Ryan's comments/suggestions in this PR
- [ ] Fix `fates_update_hifrq_hist` time call error on the ctsm-side
- [ ] Refactor the use of `tlai_hist_patch` when running in sp-mode
- [ ] Refactor the `init_patch` linked list to match other fates linked list setups
- [ ] Refactor `do_patch_dynamics` in `ed_ecosystem_dynamics` to be more simplified

rgknox:
If my comments are stylistic, don't let them get in the way of meeting an integration time window. We can always clean that stuff up in a later PR.

rgknox (approved):

glemieux:
Re-tested post ctsm-side [merge up to dev tag 053](https://github.com/ESCOMP/CTSM/pull/1182#issuecomment-901269211).  An update to MOSART (likely during dev049) is causing an issue with all the `1x1_brazil` tests.  The short-term fix was to modify `usr_nl_mosart` to add `frivinp_rtm = '/dev/null'` (per @ekluzek) and rerun the affected tests.  In the long-term, Erik entered an issue on the mosart github board:  who entered an issue with mosart: https://github.com/ESCOMP/MOSART/issues/45.  That said, it was also noted that running a compset with mosart isn't necessary for single site grids.  As such I also update PR [#1275](https://github.com/ESCOMP/CTSM/pull/1275) to change the `1x1_brazil` grid tests to use the non-mosart version of the compset `I2000ClmFatesCruRsGs`.

All expected tests except one pass b4b now: the gnu version of the `SMS_Lm13.1x1_brazil.I2000Clm50FatesCru` test.  Since the intel version passes, this seems likely due to a compiler difference, that said I didn't see this when testing the [previous merge](https://github.com/NGEET/fates/pull/703#issuecomment-898193613) up to dev046 against the same baseline. 

Location of the test suite: 
```
/glade/u/home/glemieux/scratch/ctsm-tests/tests_spmode-fatessuite
```

glemieux:
Still waiting for the tests to finish, but running `ctsm5.1.dev053` with `sci.1.46.2_api.16.1.0` and comparing against the original dev046 baseline on Cheyenne results in a DIFF for `SMS_Lm13.1x1_brazil.I2000Clm50FatesCru.cheyenne_gnu.clm-FatesColdDef` suggesting that the issue is due to the ctsm dev tag update and not the sp mode branch changes.  Test location:
```
/glade/u/home/glemieux/scratch/ctsm-tests/tests_spmode-fatessuite-gendev053
```

rosiealice:
So, I ran my global SP case earlier and it resulted in the same pretty big diffs as before. Looking at it superficially, my suspicion is that there is a oft mapping issue introduced somewhere. The diffs exist in the LAI fields and are quite large. There is a big hole where the Amazon should be, for example. I don't think it should be too difficult to trace. Will add a link to my jupyter notebook and figs in a bit.

glemieux:
I systematically ran a few more tests with fates tag `sci.1.46.2_api.16.1.0` and different ctsm tags and found the same DIFFs for the `SMS` test with `ctsm5.1.dev049`.  Reviewing the ChangeLog, this leads me to believe that the issue is due to the externals update, probably for cime.  

Looking at the cprnc output, I think the issue is likely due to the calculation of `patch%area` since `AREA_PLANT` and `AREA_TREES` are both affected.

glemieux:
Introduced ctsm [issue 1464](https://github.com/ESCOMP/CTSM/issues/1464) per Erik's request.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1105,243 +1108,473 @@ subroutine phenology_leafonoff(currentSite)
...
+             write(fates_log(),*) 'SPassign, c_area still broken',currentCohort%c_area-parea,currentCohort%c_area-oldcarea
+             call endrun(msg=errMsg(sourcefile, __LINE__))
+          end if
+       else
+          write(fates_log(),*) 'SPassign, big error in c_area',currentCohort%c_area-parea,currentCohort%pft
+       end if ! still broken
+    end if !small error
+
+    if(init.eq.ifalse)then
+       call SetState(currentCohort%prt,leaf_organ,1,leaf_c,1)
```
ckoven, line 1558:
```suggestion
       call SetState(currentCohort%prt, leaf_organ, carbon12_element, leaf_c, 1)
```

rgknox:
This line: 

Where the loop index for patch x canopy layer (in restarts):

io_idx_pa_ncl   = io_idx_co_1st

https://github.com/NGEET/fates/pull/703/files?file-filters%5B%5D=.cdl&file-filters%5B%5D=.py&w=1#diff-ec5d556c13b821b952a02a94987333cadd8e3a165a1e7b8c65b14636db13f587R1990

This should not be inside the element loop.  It does not need to be reinitialized at all in that place.

This should initialized here next to the "_ib" index:

https://github.com/NGEET/fates/pull/703/files?file-filters%5B%5D=.cdl&file-filters%5B%5D=.py&w=1#diff-ec5d556c13b821b952a02a94987333cadd8e3a165a1e7b8c65b14636db13f587R1742

And, it should continue to be updated here at the end of the patch loop (no change):

https://github.com/NGEET/fates/pull/703/files?file-filters%5B%5D=.cdl&file-filters%5B%5D=.py&w=1#diff-ec5d556c13b821b952a02a94987333cadd8e3a165a1e7b8c65b14636db13f587R2050

This general logic should be true for both the read and the write.

rgknox:
Regarding the last comment, the logic looks good and clean in the  get_restart_vectors() subroutine, just not the set_restart_vectors() subroutine.

glemieux:
The `fates` suite test runs on cheyenne are complete and were tested against `fates-sci.1.46.2_api.16.1.0-ctsm5.1.dev046`.  All expected pass with one exception: `SMS_Lm13.1x1_brazil.I2000Clm50FatesCru.cheyenne_gnu.clm-FatesColdDef` fails the baseline comparison.  I don't see anything in the ctsm changelog that suggests that there has been a gnu-specific compiler update since ctsm5.1.dev046, but I have run an updated baseline (ctsm5.1.dev054) for this test and I'm currently waiting for the comparison test against this newer baseline to complete.  There is an `intel` version of this same test and it passes b4b against ctsm5.1.dev046.

As a reminder the following are expected failures:
- ReducedComplexFixedBiogeo testmod fails DIFF, but this is acceptable as it was determined that the previous baseline was erroneous.
- ReducedComplexSatellitePhenology fails COMPARE_base_rest.  This has been recorded for a future PR in https://github.com/ESCOMP/CTSM/issues/1485.

All NLCOMP and FIELDLIST errors are expected due to the ctsm dev tag update.

glemieux:
It looks like the `SMS_Lm13.1x1_brazil.I2000Clm50FatesCru.cheyenne_gnu.clm-FatesColdDef` passes b4b against ctsm5.1.dev054.  I'm working down the ctsm dev tag list to determine where the diff first disappears to gauge the possible issue.

glemieux:
I generated a new fates baseline using `ctsm5.1.dev055` and `sci.1.46.2_api.16.1.0` and then tested this PR and https://github.com/ESCOMP/CTSM/pull/1182 against that.  All expected pass B4B including `SMS_Lm13.1x1_brazil.I2000Clm50FatesCru.cheyenne_gnu.clm-FatesColdDef`.

rosiealice:
Hurray! Thank you Greg! :)

PR #717, opened 2020-12-12
Some fire-related fixes and clarifications

### LINKED ISSUE #473: Update FIRE_AREA long name units
@rosiealice noticed that the long name units for FIRE_AREA should be updated to fraction. Incorrectly classified as m2 in long name. 
https://github.com/NGEET/fates/blob/08bb1406bf4de99252cc14faa08a7d2eb651f140/main/FatesHistoryInterfaceMod.F90#L3336

### LINKED ISSUE #575: Clean up NaN and zeroing of fire variables in Patch and Cohort Dynamics
Need to clean up Nan and zeroing of fire variables in Patch and Cohort Dynamics per issue #231 
Opening this as a separate issue due to the number of fire variables that need to be assessed in https://github.com/NGEET/fates/blob/3f6749d471e8766a212dd0a2b0c4c95aacd97f65/biogeochem/EDPatchDynamicsMod.F90#L1960-L1988
tagging @jkshuman to do this

### LINKED ISSUE #713: Changes associated with patch fraction burnt
**Issue Summary:** 
The calculation for fraction burnt "frac_burnt" is incorrectly normalized to the grid cell. 
https://github.com/NGEET/fates/blob/ac2b479022f857f1c726c88ac2b6a7765d9c71f5/fire/SFMainMod.F90#L791
This change to frac_burnt also requires a change to summarizing frac_burnt across the site. 
https://github.com/NGEET/fates/blob/ac2b479022f857f1c726c88ac2b6a7765d9c71f5/fire/SFMainMod.F90#L823-L824
Frac_burnt is used in the fire_intensity calculation and should be removed. The units of W are per m2, and so do not need to account for the amount of patch that is burnt. 
https://github.com/NGEET/fates/blob/ac2b479022f857f1c726c88ac2b6a7765d9c71f5/fire/SFMainMod.F90#L802-L804

**Solution:**
Update fract_burnt calculation to get m2 burnt/m2 per day without grid cell normalization:
`currentPatch%frac_burnt = (min(0.99_r8, AB / km2_to_m2)) `

corrected fire intensity:
`currentPatch%FI = SF_val_fuel_energy * W * ROS !kj/m/s, or kW/m`

**Additional Issues:**
@ckoven indicated that there are additional considerations within the history logic for capturing disturbance. "But also, the order of operations is such that the history logic happens *after* the patch dynamics logic, which is when the disturbance rates update the prognostic patch areas by carving off the fraction that actually burns.  So the history logic (e.g., here: https://github.com/NGEET/fates/blob/master/main/FatesHistoryInterfaceMod.F90#L2768 but also in the lines before and after it for other variables) is using the cpatch%area that already has subtracted off the newly-disturbed-by-fire patch.  And the frac_burnt of the new patch is set to zero in the act of creating it (here: https://github.com/NGEET/fates/blob/master/biogeochem/EDPatchDynamicsMod.F90#L2036 ), so the total gridcell-integrated product of currentpatch%frac_burnt*currentpatch%area isn’t being conserved.  For small fires this shouldn’t be too big a deal, but for big fires it probably is."

@ckoven has diagnostic testing looking at these changes.

In discussion with @ckoven @rgknox @adrifoster on this issue

ckoven:
there were a bunch of issues with the fire code, which this PR fixes.

### Description:
There were a bunch of fire-related issues, which include:

1. fire area was wrongly using the patch%frac_burnt to modify the fire intensity, which because intensity is used via a threshold to determine whether or not a fire actually occurs, was leading to too-low burned area
2. the decay flux from the seed pool was being routed to the FATES leaf litter pool, where it could act as a fuel source.  this meant we were both double-counting the seed C stocks in the scheme and also meant that under the special cases where a plant allocates a large fraction of its NPP to seeds (this can particularly happen with grasses right now) it would build up huge amounts of fuel and thus fire area.  I've changed this to reroute the seed decay fluxes straight to the HLM decomposition cascade rather than the intermediate step of the FATES leaf litter pool.  some day we may want to directly couple the seed pool to the fire behavior and effects but for now this should make the model behavior more as expected.
3. re-added the fire ellipse length-to-breadth calculation in the fire code.  This had been hard-coded to one (thus all fires are circular rather than elliptical), which was making the fire areas too big.  also there were typos in the equations for the length-to-breadth code. One of the typos goes back to the original 1992 Canadian fire behavior model (which noted in a 2009 errata that there had been a typo) and thus possibly exists in all extant versions of SPITFIRE (not just FATES-SPITFIRE).  In addition to the typos, we reverted the length-to-breadth logic to be categorically distinct—rather than a continuous transition—between grasslands and forests.  we can revisit this later, as well as what value of tree cover should switch between them, but for changed it to the separate equation approach.
4. added some more comment tracing the provenance of various equations and more info on the units where helpful.
5. added a couple more history variables that are helpful (e.g. PFT-level NPP and GPP) for investigating things.  also fixed and simplified some unit descriptions.
6. the MEF (moisture extinction factor) equation didn't have a clear provenance; once that was sorted out we realized there was a typo there as well (log10->log).
7. nan (rather than zero) the fire-relevant patch variables when initializing them.

Fixes #713 
Fixes #473 
Fixes #575

### Collaborators:
Lots of discussion and code from @jkshuman and also discussion with @lmkueppers @pollybuotte @rosiealice @rgknox @glemieux @JunyanDing 

highly answer changing if fire on.  also answer-changing litter and belowground BGC dynamics for case where fire is off.

### Checklist:

still need to update the tech note to be consistent with the changes here.  will do so before merging this PR.

### Test Results:
not yet tested

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -2034,6 +2034,23 @@ contains
     new_patch%fabi_sha_z(:,:,:)  = 0._r8  
     new_patch%scorch_ht(:)       = 0._r8  
     new_patch%frac_burnt         = 0._r8  
+    new_patch%litter_moisture(:) = 0._r8
+    new_patch%fuel_eff_moist     = 0._r8
+    new_patch%livegrass          = 0._r8
+    new_patch%sum_fuel           = 0._r8
+    new_patch%fuel_bulkd         = 0._r8
+    new_patch%fuel_sav           = 0._r8
+    new_patch%fuel_mef           = 0._r8
+    new_patch%ros_front          = 0._r8
+    new_patch%effect_wspeed      = 0._r8
+    new_patch%tau_l              = 0._r8
+    new_patch%fuel_frac(:)       = 0._r8
+    new_patch%tfc_ros            = 0._r8
+    new_patch%fi                 = 0._r8
+    new_patch%fd                 = 0._r8
+    new_patch%ros_back           = 0._r8
+    new_patch%scorch_ht(:)       = 0._r8
+    new_patch%burnt_frac_litter(:) = 0._r8
     new_patch%total_tree_area    = 0.0_r8  
     new_patch%NCL_p              = 1
 
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -2108,31 +2108,31 @@ contains
 
 
     ! FIRE
-   currentPatch%litter_moisture(:)         = 0.0_r8 ! litter moisture
-    currentPatch%fuel_eff_moist             = 0.0_r8 ! average fuel moisture content of the ground fuel 
+    currentPatch%litter_moisture(:)         = nan    ! litter moisture
+    currentPatch%fuel_eff_moist             = nan    ! average fuel moisture content of the ground fuel 
     ! (incl. live grasses. omits 1000hr fuels)
-    currentPatch%livegrass                  = 0.0_r8 ! total ag grass biomass in patch. 1=c3 grass, 2=c4 grass. gc/m2
-    currentPatch%sum_fuel                   = 0.0_r8 ! total ground fuel related to ros (omits 1000hr fuels). gc/m2
-    currentPatch%fuel_bulkd                 = 0.0_r8 ! average fuel bulk density of the ground fuel 
+    currentPatch%livegrass                  = nan    ! total ag grass biomass in patch. 1=c3 grass, 2=c4 grass. gc/m2
+    currentPatch%sum_fuel                   = nan    ! total ground fuel related to ros (omits 1000hr fuels). gc/m2
+    currentPatch%fuel_bulkd                 = nan    ! average fuel bulk density of the ground fuel 
     ! (incl. live grasses. omits 1000hr fuels). kgc/m3
-    currentPatch%fuel_sav                   = 0.0_r8 ! average surface area to volume ratio of the ground fuel 
+    currentPatch%fuel_sav                   = nan    ! average surface area to volume ratio of the ground fuel 
     ! (incl. live grasses. omits 1000hr fuels).
-    currentPatch%fuel_mef                   = 0.0_r8 ! average moisture of extinction factor of the ground fuel
+    currentPatch%fuel_mef                   = nan    ! average moisture of extinction factor of the ground fuel
     ! (incl. live grasses. omits 1000hr fuels).
-    currentPatch%ros_front                  = 0.0_r8 ! average rate of forward spread of each fire in the patch. m/min.
-    currentPatch%effect_wspeed              = 0.0_r8 ! dailywind modified by fraction of relative grass and tree cover. m/min.
-    currentPatch%tau_l                      = 0.0_r8 ! mins p&r(1986)
-    currentPatch%fuel_frac(:)               = 0.0_r8 ! fraction of each litter class in the sum_fuel 
+    currentPatch%ros_front                  = nan    ! average rate of forward spread of each fire in the patch. m/min.
+    currentPatch%effect_wspeed              = nan    ! dailywind modified by fraction of relative grass and tree cover. m/min.
+    currentPatch%tau_l                      = nan    ! mins p&r(1986)
+    currentPatch%fuel_frac(:)               = nan    ! fraction of each litter class in the sum_fuel 
     !- for purposes of calculating weighted averages. 
-    currentPatch%tfc_ros                    = 0.0_r8 ! used in fi calc
-    currentPatch%fi                         = 0._r8  ! average fire intensity of flaming front during day.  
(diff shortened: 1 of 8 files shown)


rgknox:
some tests on cheyenne are failing at line 2213 in  main/FatesHistoryInterfaceMod.F90

Here is the log:
```
812:MPT: #1  0x00002ab78894e866 in mpi_sgi_system (
812:MPT: #2  MPI_SGI_stacktraceback (
812:MPT:     header=header@entry=0x7ffd4aa9d600 "MPT ERROR: Rank 812(g:812) received signal SIGFPE(8).\n\tProcess ID: 43396, Host: r12i4n34, Program: /glade/scratch/rgknox/ERP_D_Ld3.f19_g16.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesColdDef.C.202012"...) at sig.c:340
812:MPT: #3  0x00002ab78894ea62 in first_arriver_handler (signo=signo@entry=8, 
812:MPT:     stack_trace_sem=stack_trace_sem@entry=0x2ab792fc0080) at sig.c:489
812:MPT: #4  0x00002ab78894edfb in slave_sig_handler (signo=8, siginfo=<optimized out>, 
812:MPT:     extra=<optimized out>) at sig.c:565
812:MPT: #5  <signal handler called>
812:MPT: #6  0x00000000016dd595 in fateshistoryinterfacemod::update_history_dyn (
812:MPT:     this=0x5ac9970 <clm_instmod_mp_clm_fates_+144>, nc=1, nsites=4, sites=...)
812:MPT:     at /glade/u/home/rgknox/ctsm/src/fates/main/FatesHistoryInterfaceMod.F90:2213
812:MPT: #7  0x00000000009ba8a0 in clmfatesinterfacemod::L_clmfatesinterfacemod_mp_init_coldstart__1453__par_loop0_2_121 ()
812:MPT:     at /glade/u/home/rgknox/ctsm/src/utils/clmfates_interfaceMod.F90:1535
812:MPT: #8  0x00002ab787131d43 in __kmp_invoke_microtask ()
812:MPT:    from /glade/u/apps/opt/intel/2019u5/compilers_and_libraries/linux/lib/intel64/libiomp5.so
812:MPT: #9  0x00002ab7870c1d65 in __kmp_fork_call (loc=0x2ac7c41b7168, gtid=95221696, 
812:MPT:     call_context=(unknown: 3286322944), argc=1252673288, 
812:MPT:     microtask=0x4ab7d08 <__STRLITPACK_22>, invoker=0x38525f49, 
812:MPT:     ap=0x7ffd4aabe160) at ../../src/kmp_runtime.cpp:2085
812:MPT: #10 0x00002ab787081c60 in __kmpc_fork_call (loc=0x2ac7c41b7168, argc=95221696, 
812:MPT:     microtask=0x2ac7c3e14f00) at ../../src/kmp_csupport.cpp:360
812:MPT: #11 0x00000000009a6413 in clmfatesinterfacemod::init_coldstart (
812:MPT:     this=0x5ac98e0 <clm_instmod_mp_clm_fates_>, waterstatebulk_inst=..., 
1155:MPT:     waterdiagnosticbulk_inst=..., canopystate_inst=..., soilstate_inst=...)
1155:MPT:     at /glade/u/home/rgknox/ctsm/src/utils/clmfates_interfaceMod.F90:1453
1155:MPT: #12 0x000000000090e112 in clm_initializemod::initialize2 ()
1155:MPT:     at /glade/u/home/rgknox/ctsm/src/main/clm_initializeMod.F90:753
1155:MPT: #13 0x0000000000888eb9 in lnd_comp_mct::lnd_init_mct (eclock=..., cdata_l=..., 
1155:MPT:     x2l_l=..., l2x_l=..., nlfilename=..., .tmp.NLFILENAME.len_V$757a=6)
1155:MPT:     at /glade/u/home/rgknox/ctsm/src/cpl/mct/lnd_comp_mct.F90:238
```

This seems to me like an uninitialized variable issue, probably related to patch%frac_burnt.  Notice that with a cold-start, (see clmfates_interfaceMod.F90 line 1535) we do not call dynamics at time zero on that first day, but we do initialize the history variables, and thus we are trying to use variables like %frac_burnt (and some other fire variables as well) but they do not have a value yet.

Solutions:
1) short term: allow zero'ing of fire variables but only during the cold-start initialization
2) long-term: I feel this is yet another reason we should de-centralize the history variable filling process.  In this situation, the history fields would be initialized with the special value flag (like ignore/invalid), and then we would update the history variable immediately upon where the variable of interest is updated in the fire code

rgknox:
tests, not b4b changes, run tests pass: 
/glade/scratch/rgknox/clmed-tests/fates-sci.1.43.3_api.14.2.0-ctsm1.0.dev113-Ca7e020f2-Fc0f3957e.fates.cheyenne

jkshuman:
I can confirm that the last commit to re-add patch%frac_burnt=0 for below threshold intensity fires does repair the previous simulation fail. Tests across South America are successful out to 28 years, where previous runs failed in year 0 with a mass balance error. Thanks for @ckoven and @rgknox for discussions on this.

PR #727, opened 2021-03-22
mass conservation fix on hydro recruit limitations

rgknox:
### Description:

These changes fix mass conservation errors generated when soil water availability limits the number of new recruits when the hydro model is active.  Previously, the number of new recruits was reduced from what the CNP model had predicted, however we were not re-adjusting the amount of mass pulled from the seed germination pool.

### Collaborators:

@pnlfang 

This should change answers with hydro, but I would not be surprised if it is not triggered by a test (but maybe it will).

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeophys/FatesPlantHydraulicsMod.F90
+++ b/biogeophys/FatesPlantHydraulicsMod.F90
@@ -90,10 +90,12 @@ module FatesPlantHydraulicsMod
   use FatesHydraulicsMemMod, only: rwccap, rwcft
   use FatesHydraulicsMemMod, only: ignore_layer1
   
-  use PRTGenericMod,          only : all_carbon_elements
+  use PRTGenericMod,          only : carbon12_element
   use PRTGenericMod,          only : leaf_organ, fnrt_organ, sapw_organ
   use PRTGenericMod,          only : store_organ, repro_organ, struct_organ
-
+  use PRTGenericMod,          only : num_elements
+  use PRTGenericMod,          only : element_list
+  
   use clm_time_manager  , only : get_step_size, get_nstep
 
   use EDPftvarcon, only : EDPftvarcon_inst
@@ -832,10 +834,10 @@ contains
     ccohort_hydr => ccohort%co_hydr
     ft           = ccohort%pft
     nlevrhiz = site_hydr%nlevrhiz
-    leaf_c   = ccohort%prt%GetState(leaf_organ, all_carbon_elements)
-    sapw_c   = ccohort%prt%GetState(sapw_organ, all_carbon_elements)
-    fnrt_c   = ccohort%prt%GetState(fnrt_organ, all_carbon_elements)
-    struct_c = ccohort%prt%GetState(struct_organ, all_carbon_elements)
+    leaf_c   = ccohort%prt%GetState(leaf_organ, carbon12_element)
+    sapw_c   = ccohort%prt%GetState(sapw_organ, carbon12_element)
+    fnrt_c   = ccohort%prt%GetState(fnrt_organ, carbon12_element)
+    struct_c = ccohort%prt%GetState(struct_organ, carbon12_element)
 
     roota    = prt_params%fnrt_prof_a(ft)
     rootb    = prt_params%fnrt_prof_b(ft)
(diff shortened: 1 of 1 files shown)


biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1651,8 +1661,33 @@ subroutine ConstrainRecruitNumber(csite,ccohort, bc_in)
           nmin = min(n, nmin) 
        endif
     end do
-    ccohort%n = min (ccohort%n, nmin) 
 
+    ! If the minimum number of plants is less than what was dictated by the
+    ! carbon-nitrogen-phosphorus availability, then we appy a reduction.
```
ckoven, line 1666: isn't it the water availability that limits things in this logic, not the carbon-nitrogen-phosphorus availability?

ckoven (approved):
looks good, thanks @rgknox!  just a tiny comment on one of your code comments.

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: will clean that up, thanks

rgknox:
Changes in this branch, along with : https://github.com/ESCOMP/CTSM/pull/1334, are passing regression restart tests "FatesHydro" now.
Performing standard regression tests now via ctsmdev025, note that these are not expected to pass the hydro restart comparison portion of the tests.

rgknox:
Compared with fates-sci.1.44.0_api.14.2.0-ctsm5.1.dev025-C2094348b-F74433f48:
All expected tests PASS, and b4b
/glade/scratch/rgknox/clmed-tests/fates-sci.1.44.1_api.14.2.0-ctsm5.1.dev025-C2094348b-F83d55d24-hydro-fixes.fates.cheyenne

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -2010,15 +2007,26 @@ subroutine update_hlm_dynamics(nsites,sites,fcolumn,bc_out)
...
+        ! will demand the water for their initialization on the first hydraulics time-step
+        
+        if (hlm_use_planthydro.eq.itrue) then
+           call UpdateH2OVeg(sites(s),bc_out(s),bc_out(s)%plant_stored_h2o_si,1)
+        end if
         
      end do
 
-     ! If hydraulics is turned on, update the amount of water bound in vegetation
+     ! This call is purely for diagnostics and history variables
```
xuchongang, line 2025: It might be good to point out the real recruit water demand is handled by subroutine RecruitWUptake(nsites,sites,bc_in,dtime,recruitflag). This is only for diagnostics and history variables. I am wondering if we should delete RecruitWaterStorage as it is redundant and it is not functional at all.

xuchongang (commented):
@rgknox , I added a few comments on the code. Please let me know your thoughts.

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -2010,15 +2007,26 @@ subroutine update_hlm_dynamics(nsites,sites,fcolumn,bc_out)
...
      if (hlm_use_planthydro.eq.itrue) then
         call RecruitWaterStorage(nsites,sites,bc_out)
```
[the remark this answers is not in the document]
rgknox: @xuchongang , thanks for bringing this up.  After re-reading this, I'm realizing that we are not accounting for the water in the new recruits in the total mass balance. 
Did we once have code that pulled water from the soil and put it in the new-recruits, or have we thus far always treated it as an error term?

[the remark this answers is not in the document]
xuchongang: It is in subroutine RecruitWUptake(nsites,sites,bc_in,dtime,recruitflag)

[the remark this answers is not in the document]
rgknox: Ok, I see it now. I think we are accounting for the mass correctly then.   That routine is called inside the main hydraulics loop, and should only be relevant on the first call following a dynamics call. Routine RecruitWUptake()  updates the site x depth level data structure site_hydr%recruit_w_uptake(j).  This is then applied to the uptake flux, here:
https://github.com/rgknox/fates/blob/fix-recruit-hydro/biogeophys/FatesPlantHydraulicsMod.F90#L2586-L2588

However, I think we could potentially improve the order of operations.  Since the recruit water uptake is already pre-determined, I think we should remove it from the soil water (site_hydr%h2osoi_liqvol_shell(:,:)) before implementing the hydraulic fluxes that follow.

biogeochem/EDCanopyStructureMod.F90
rgknox: We could potentially delete RecruitWaterStorage(), but its not straightforward.  It is called near the end of the dynamics sequence, but before the call to update history variables.  Here is the variable it informs, which is updated with the rest of the dynamics history variables:
https://github.com/NGEET/fates/blob/master/main/FatesHistoryInterfaceMod.F90#L2125

If we combined RecruitWaterStorage() into RecruitWUptake(), the update of the history variable wouldn't happen at the right time.

rgknox:
re-tested:  /glade/scratch/rgknox/clmed-tests/fates-hydro-fixes-C61770ad0-F3bddb84b-v2

PR #733, opened 2021-03-25
remove artificial minimum LAI value

### LINKED ISSUE #261: LAI sent to host has a forced minimum
In EDCanopyStructureMod.F90:calc_areaindex()  we impose an artificial minimum on all area indices.  In other words, LAI/TAI/etc is not allowed to be lower than 0.1.

I think this is a pretty old artifact in the code, it has been left alone for this long just to retain B4B results in tests, but I think we really should remove this.

Its possible this was originally added to prevent a div0 somewhere, but we can find protections against this.

I could conceive that this minimum is generating some weird inconsistency that is affecting @jkshuman 's runs, but this is just a stab in the dark (might be worth testing its removal if no other ideas are generated).

ckoven:
Removes the artificial minimum LAI of 0.1 that is being used in the canopy code, which was there at some point to provide stability.  Hopefully we can get rid of it.

fixes #261

expected to be answer changing.

### Checklist:

### Test Results:

Not yet tested. Expectation is that it will not be bit for bit.

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -2040,9 +2040,6 @@ contains
 
      integer :: cl,ft
      real(r8) :: ai
-     ! TODO: THIS MIN LAI IS AN ARTIFACT FROM TESTING LONG-AGO AND SHOULD BE REMOVED
-     ! THIS HAS BEEN KEPT THUS FAR TO MAINTAIN B4B IN TESTING OTHER COMMITS
-     real(r8),parameter :: ai_min = 0.1_r8
      real(r8),pointer   :: ai_profile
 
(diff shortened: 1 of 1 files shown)


rgknox (approved):
Looks pretty straightforward.

rosiealice (approved):

rgknox:
All expected PASS

/glade/scratch/rgknox/clmed-tests/no_min_lai-ctsm5.1.dev025-C2094348b-Ff1842374.fates.cheyenne

(had to re-run one test:)
/glade/scratch/rgknox/ERP_Ld3.f09_g17.I2000Clm50Fates.cheyenne_intel.clm-FatesColdDef.retry-string-format

Issue #726, opened 2021-03-18
Plant hydraulics updates


ckoven:
Following up on today's FATES call, several developments are either proposed or underway for plant hydraulics in FATES.  Just want to list some of them here, for reference and tracking of their evolving status. Feel free to add to and/or update this thread as things proceed.  cc: @JunyanDing @xuchongang @pnlfang

- reworking some of the root-layering so as to generalize away from the requirement that every soil layer has a corresponding root layer.
- updates to P-V and PLC curves.
- removing the assumption that leaf internal relative humidity equals 100% when leaf water potentials become dry.
- incorporation of FETCH2 dynamics to more finely resolve canopy water potential gradients.
- possibly adding lower-complexity hydraulic models, e.g. similar to CLM5 PHS, to FATES.
- @JunyanDing didn't get to this today but is also adding size-resolved cohort-specific rooting depths.

glemieux:
Closing this as stale per discussion with @rgknox and @ckoven.  There is nothing currently actionable here.

PR #737, opened 2021-04-08
Adding the van Genuchten formula for plant P-V and P-K curves

JunyanDing:
### Description:

see attached file 
[ForFATES_Hydro_VG.pptx](https://github.com/NGEET/fates/files/6282643/ForFATES_Hydro_VG.pptx)

issue number: #726

Fixes: #812

### Collaborators:

Ryan Knox, Chonggang Xu, Charlie Koven

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeophys/FatesHydroWTFMod.F90
+++ b/biogeophys/FatesHydroWTFMod.F90
@@ -34,6 +34,8 @@ module FatesHydroWTFMod
   ! In this context, the saturated fraction is defined by the volumetric WC "th"
   ! and the volumetric residual and saturation "th_res" and "th_sat": (th-th_r)/(th_sat-th_res)
 
+  real(r8), parameter :: min_sf_interp = 0.01 ! Linear interpolation below this saturated frac
+  real(r8), parameter :: max_sf_interp = 0.98 ! Linear interpolation above this saturated frac
   real(r8), parameter :: min_sf_interp = 0.02 ! Linear interpolation below this saturated frac
   real(r8), parameter :: max_sf_interp = 0.99 ! Linear interpolation above this saturated frac
 
@@ -112,7 +114,8 @@ module FatesHydroWTFMod
   ! Water Retention Function
   type, public, extends(wrf_type) :: wrf_type_vg
      real(r8) :: alpha   ! Inverse air entry parameter         [m3/Mpa]
-     real(r8) :: psd     ! Inverse width of pore size distribution parameter
+     real(r8) :: n_vg    ! pore size distribution parameter, psd in original code
+     real(r8) :: m_vg    ! m in van Genuchten 1980, also a pore size distribtion parameter , 1-m in original code   
      real(r8) :: th_sat  ! Saturation volumetric water content [m3/m3]
      real(r8) :: th_res  ! Residual volumetric water content   [m3/m3]
    contains
@@ -126,7 +129,8 @@ module FatesHydroWTFMod
   ! Water Conductivity Function
   type, public, extends(wkf_type) :: wkf_type_vg
      real(r8) :: alpha   ! Inverse air entry parameter         [m3/Mpa]
-     real(r8) :: psd     ! Inverse width of pore size distribution parameter
+     real(r8) :: n_vg    ! pore size distribution parameter
+     real(r8) :: m_vg    ! m in van Genuchten 1980, also a pore size distribtion parameter ,      
      real(r8) :: tort    ! Tortuosity parameter (sometimes "l")
      real(r8) :: th_sat  ! Saturation volumetric water content [m3/m3]
      real(r8) :: th_res  ! Residual volumetric water content   [m3/m3]
@@ -176,6 +180,7 @@ module FatesHydroWTFMod
      real(r8) :: pinot    ! osmotic potential at full turger            [MPa]
      real(r8) :: epsil    ! bulk elastic modulus                        [MPa]
      real(r8) :: rwc_ft   ! RWC @ full turgor, (elastic drainage begins)[-]
+     real(r8) :: psicap   ! xylem water potential where capilary source is exhausted [Mpa] 
      real(r8) :: cap_corr ! correction for nonzero psi0x
      real(r8) :: cap_int  ! intercept of capillary region of curve
      real(r8) :: cap_slp  ! slope of capillary region of curve
@@ -378,9 +383,14 @@ contains
     real(r8), intent(in) :: params_in(:)
 
     this%alpha  = params_in(1)
-    this%psd    = params_in(2)
-    this%th_sat = params_in(3)
-    this%th_res = params_in(4)
+    this%n_vg    = params_in(2)
+    this%m_vg    = params_in(3)    
+    this%th_sat = params_in(4)
+    this%th_res = params_in(5)
+
+    !write(fates_log(),*) 'set_wrf_param_vg'
+    !write(fates_log(),*) 'th_sat:',this%th_sat, 'th_res: ', this%th_res
+    !write(fates_log(),*) 'alpha:',     this%alpha, 'm: ',  this%m_vg, 'n: ',  this%n_vg
 
     call this%set_min_max(this%th_res,this%th_sat)
     
@@ -395,11 +405,15 @@ contains
     real(r8), intent(in) :: params_in(:)
 
     this%alpha  = params_in(1)
-    this%psd    = params_in(2)
-    this%th_sat = params_in(3)
-    this%th_res = params_in(4)
-    this%tort   = params_in(5)
+    this%n_vg   = params_in(2)
+    this%m_vg   = params_in(3)
+    this%th_sat = params_in(4)
+    this%th_res = params_in(5)
+    this%tort   = params_in(6)
 
+    !write(fates_log(),*) 'set_wkf_param_vg'
+    !write(fates_log(),*) 'th_sat:',this%th_sat, 'th_res: ', this%th_res
+    !write(fates_log(),*) 'alpha:',     this%alpha, 'm: ',  this%m_vg, 'n: ',     this%n_vg
     return
   end subroutine set_wkf_param_vg
 
@@ -426,30 +440,38 @@ contains
     real(r8)             :: th            ! Volumetric Water Cont [m3/m3]
 
     real(r8)             :: dpsidth_interp  ! change in psi during lin interp (slope)
-    real(r8)             :: m               ! pore size distribution param (1/n)
+    real(r8)             :: m               ! pore size distribution param 1
+    real(r8)             :: n               ! pore size distribution param 2    
 
-    m   = 1._r8/this%psd
+    m = this%m_vg
+    n = this%n_vg
 
-    if(psi>this%psi_max) then
+    ! pressure above which we use a linear function, 
+    ! psi_interp = -(1._r8/this%alpha)*(max_sf_interp**(1._r8/(-m)) - 1._r8 )**(1/n)
 
-        ! Linear range for extreme values
-        th = this%th_linear_sat(psi)
-        
-    elseif(psi<this%psi_min) then
-        
-        ! Linear range for extreme values
-        th = this%th_linear_res(psi)
+    ! Junyan modified to get rid of the linear interperation
+      psi_interp = 0
 
     else
 
        ! Saturation fraction
-       satfrac = (1._r8 + (-this%alpha*psi)**this%psd)**(-1._r8+m)
+       satfrac = (1._r8 + (-this%alpha*psi)**n)**(-m)
 
        ! convert to volumetric water content
        th = satfrac*(this%th_sat-this%th_res) + this%th_res
 
+    else
+       !th_interp = max_sf_interp * (this%th_sat-this%th_res) + this%th_res
+       ! dpsidth_interp = this%dpsidth_from_th(th_interp)
 
+       !th = th_interp + (psi-psi_interp)/dpsidth_interp
+       th = this%th_sat
+       !write(fates_log(),*) 'linear interpreation ', 'th_interp: ', th_interp   
+       !write(fates_log(),*) 'linear interpreation ', 'psi_interp: ', psi_interp            
     end if
+
+    !write(fates_log(),*) 'm:',m, 'n: ',n 
+    !write(fates_log(),*) 'th_from_psi_vg ','th:', th, 'psi:',psi
 
   end function th_from_psi_vg
 
@@ -463,7 +485,8 @@ contains
     class(wrf_type_vg)   :: this
     real(r8),intent(in)  :: th
     real(r8)             :: psi            ! matric potential [MPa]
-    real(r8)             :: m              ! inverse of psd
+    real(r8)             :: m              ! 
+    real(r8)             :: n              ! 
     real(r8)             :: satfrac        ! saturated fraction
     real(r8)             :: th_interp      ! theta where we start interpolation
     real(r8)             :: psi_interp     ! psi at interpolation point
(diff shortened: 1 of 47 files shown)


biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -121,6 +121,11 @@ module FatesPlantHydraulicsMod
   ! 1 => use BC hydraulics; 
   ! 2 => use CX hydraulics
 
+  ! use a control to prevent model from quiting when encounter the water balance error
+  ! instead put it into an endless loop   (JD)
+  integer, public :: debug_JD = 1                        ! set to 0 to disable the
```
xuchongang, line 126: why endless loop?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -121,6 +121,11 @@ module FatesPlantHydraulicsMod
   ! 1 => use BC hydraulics; 
   ! 2 => use CX hydraulics
 
+  ! use a control to prevent model from quiting when encounter the water balance error
+  ! instead put it into an endless loop   (JD)
+  integer, public :: debug_JD = 1                        ! set to 0 to disable the 
+  integer, public :: RP = 1                              !
```
xuchongang, line 127: what is this?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -192,8 +197,8 @@ module FatesPlantHydraulicsMod
   integer, public, parameter :: campbell_type           = 2
   integer, public, parameter :: tfs_type                = 3
   
-  integer, parameter :: plant_wrf_type = tfs_type
-  integer, parameter :: plant_wkf_type = tfs_type
+  integer, parameter :: plant_wrf_type = van_genuchten_type
```
xuchongang, line 200: Is this set the default as van_genuchten? I would prefer to keep the default as TFS

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1651,7 +1663,13 @@ subroutine ConstrainRecruitNumber(csite,ccohort, bc_in)
           nmin = min(n, nmin) 
        endif
     end do
-    ccohort%n = min (ccohort%n, nmin) 
+    ! Junyan added the constrian of temperature on recuitment number
```
xuchongang, line 1666: This is fine and make sense

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -425,32 +427,42 @@ function th_from_psi_vg(this,psi) result(th)
...
 
-        ! Linear range for extreme values
-        th = this%th_linear_sat(psi)
-        
-    elseif(psi<this%psi_min) then
-        
-        ! Linear range for extreme values
-        th = this%th_linear_res(psi)
+    ! Junyan modified to get rid of the linear interperation
+      psi_interp = 0
```
xuchongang, line 443: why get rid of the linear interpolation?

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -475,29 +488,32 @@ function psi_from_th_vg(this,th) result(psi)
...
     
-    if(th>this%th_max)then
-
-        psi = this%psi_linear_sat(th)
-        
-    elseif(th<this%th_min)then
-        
-        psi = this%psi_linear_res(th)
-
+    ! Junyan disable the linear interperation below
```
xuchongang, line 500: why disable the linear interoperation?

xuchongang (commented):
@JunyanDing , it seem that you set the default as the van_genuchten for vegetation P-V/P-K curves. Can we set this as a switch to turn it on or off? It is also not clear to me why you turned off the linear interpolation. I would also suggest that @rgknox review the code as well.

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: @xuchongang 
Yes, there is a variable in the code to control which model to use for plants. 
The endless loop is used to avoid the water balance errors that may cause model end in large ensemble I am running. As for the pull request, I will delete this section for the codes to be meagered.

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: I will change default to be TFS

biogeophys/FatesHydroWTFMod.F90
JunyanDing: The linear interpretation will result in the > the_sat. I think, instead of disable it, I should leave it as an option.

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -99,22 +99,27 @@ module FatesHydroWTFMod
 
   type, public :: wrf_arr_type
      class(wrf_type), pointer :: p
+     real(r8) :: th_sat
```
rgknox, line 102: I think these can be removed, I don't think these are used

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -800,22 +831,32 @@ subroutine set_wrf_param_tfs(this,params_in)
...
     this%th_sat   = params_in(1)
     this%th_res   = params_in(2)
     this%pinot    = params_in(3)
     this%epsil    = params_in(4)
     this%rwc_ft   = params_in(5)
-    this%cap_corr = params_in(6)
-    this%cap_int  = params_in(7)
-    this%cap_slp  = params_in(8)
-    this%pmedia   = int(params_in(9))
+    this%psicap   = params_in(6)
```
rgknox, line 842: I think this is also unused, right? can this be removed?

rgknox: Also, the arguments initializing TFS parameterizations are not changed, so I don't think this call would work:

https://github.com/NGEET/fates/pull/737/files#diff-e5e2459b995e00503077bce86074962e2acc4780ca00f248242b153527655c42R5346-R5353

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -800,22 +831,32 @@ subroutine set_wrf_param_tfs(this,params_in)
...
+    ! used for interpolating in extreme ranges
+    th_max=max_sf_interp*(this%th_sat-this%th_res)+this%th_res-1.e-9_r8
+    th_min=min_sf_interp*(this%th_sat-this%th_res)+this%th_res+1.e-9_r8
+    this%psi_max     = this%psi_from_th(th_max)
+    this%dpsidth_max = this%dpsidth_from_th(th_max)
+    this%psi_min     = this%psi_from_th(th_min)
+    this%dpsidth_min = this%dpsidth_from_th(th_min)
+
 
-    call this%set_min_max(this%th_res,this%th_sat)
```
rgknox, line 814: I agree with @xuchongang , lets keep the ability to use extrapolations, and use a switch to turn that option on/off.  So lets keep this call to set those boundaries

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -847,21 +888,21 @@ function th_from_psi_tfs(this,psi) result(th)
     if(psi>this%psi_max) then
 
         ! Linear range for extreme values
-
-        th = this%th_linear_sat(psi)
+        th = this%th_res+max_sf_interp*(this%th_sat-this%th_res) + &
```
rgknox, line 891: this looks like an alternative to the current linear extrapolation scheme.  I think this alternative should be encapsulated in a switch, and probably writtten into the th_linear_sat() routine.

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -925,7 +935,8 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
     ! The transporting root donates some of its volume
     ! to the layer-by-layer absorbing root (which is now a hybrid compartment)
     ! ------------------------------------------------------------------------------
-    ccohort_hydr%v_troot = (1._r8-t2aroot_vol_donate_frac) * v_troot
+    ! ccohort_hydr%v_troot = (1._r8-t2aroot_vol_donate_frac) * v_troot
+    ccohort_hydr%v_troot = ( v_troot + v_aroot_tot) / 2
```
rgknox, line 939: please change 2 to 2._r8  and other instances as well please

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -216,11 +220,13 @@ module FatesPlantHydraulicsMod
...
   real(r8), parameter :: th_res_vg = 0.15_r8
   real(r8), parameter :: psd_vg    = 2.7_r8
+  real(r8), parameter :: m_vg      = 0.62963_r8
   real(r8), parameter :: tort_vg   = 0.5_r8
+  real(r8), parameter :: plant_tort = 0.0_r8
 
   ! The maximum allowable water balance error over a plant-soil continuum
   ! for a given step [kgs] (0.1 mg)
-  real(r8), parameter :: max_wb_step_err = 1.e-7_r8 
+  real(r8), parameter :: max_wb_step_err = 2.e-6_r8
```
rgknox, line 229: I feel the tolerances that we've specified up to this point have been fairly arbitrary, and some of them were high enough to obscure errors. 1 e-7 is a tenth of a milligram per plant, now it is 2 milligrams per plant.  I honestly don't know what the right answer is here.  Plants can be small and the total water flux in a 30 minute time-step could be around a milligram, no? The intent of the tolerances are to ensure that the numerical integration's truncation error (solution error) is small within reason.  When its large it could potentially allow errors from other sources to pass by undetected.  
@JunyanDing , is this increased tolerance still required for runs to complete?  If it is, I suppose we should move forward for now, but create an issue to further evaluate why the errors are large.

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1651,7 +1662,13 @@ subroutine ConstrainRecruitNumber(csite,ccohort, bc_in)
           nmin = min(n, nmin) 
        endif
     end do
-    ccohort%n = min (ccohort%n, nmin) 
+    ! Junyan added the constrian of temperature on recuitment number
+    if (bc_in%t_veg_pa(1) > 273.15_r8) then
```
rgknox, line 1666: This will use the vegetation temperature from patch 1 for all plants. Another alternative is to use the site-level soil temperature (which is indexed by layer).  See here:

https://github.com/NGEET/fates/blob/master/main/FatesInterfaceTypesMod.F90#L431

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -2490,8 +2511,8 @@ subroutine hydraulics_bc ( nsites, sites, bc_in, bc_out, dtime)
              call UpdatePlantPsiFTCFromTheta(ccohort,site_hydr)
 
              ccohort_hydr%btran = wkf_plant(stomata_p_media,ft)%p%ftc_from_psi(ccohort_hydr%psi_ag(1))
-             
-
+             ! write(fates_log(),*) 'cohort(dbh):', ccohort%dbh, 'btran', ccohort_hydr%btran
```
rgknox, line 2514: please remove commented out write statements before we get to final integration phase

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -3221,9 +3245,16 @@ subroutine ImTaylorSolve1D(site_hydr,cohort,cohort_hydr,dtime,q_top, &
             if(iter>max_iter)then
                 call Report1DError(cohort,site_hydr,ilayer,z_node,v_node, & 
                       th_node_init,q_top_eff,dt_step,w_tot_beg,w_tot_end,& 
-                      rootfr_scaler,aroot_frac_plant,error_code,error_arr)
-
-                call endrun(msg=errMsg(sourcefile, __LINE__))
+                      rootfr_scaler,aroot_frac_plant,error_code,error_arr,slat,slon,recruitflag)
+              !  if (debug_JD>0) then
```
rgknox, line 3249: please remove unused write statements before final integration phase

main/EDTypesMod.F90
```diff
@@ -171,7 +171,8 @@ module EDTypesMod
                                                                           ! if the fusion area is less than min_patch_area_forced
 
   real(r8), parameter, public :: min_nppatch    = min_npm2*min_patch_area ! minimum number of cohorts per patch (min_npm2*min_patch_area)
-  real(r8), parameter, public :: min_n_safemath = 1.0E-12_r8              ! in some cases, we want to immediately remove super small
+  ! Junyan changed the min_n_safemath below , original is 1.0E-12_r8
+  real(r8), parameter, public :: min_n_safemath = 1.0E-9_r8               ! in some cases, we want to immediately remove super small,
```
rgknox, line 175: please revert this back to e-12

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -321,7 +323,6 @@ function th_from_psi_base(this,psi) result(th)
     class(wrf_type)     :: this
     real(r8),intent(in) :: psi
     real(r8)            :: th
-    th = 0._r8
```
rgknox, line 324: These should not be removed, this will cause compile problems for XLF if the base does have  a result.

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -425,32 +428,49 @@ function th_from_psi_vg(this,psi) result(th)
...
-        
-    elseif(psi<this%psi_min) then
-        
-        ! Linear range for extreme values
-        th = this%th_linear_res(psi)
+    m = this%m_vg
+    n = this%n_vg
 
+    if (use_linear_interp) then
+       pressure above which we use a linear function,
```
rgknox, line 441: please comment out, this should not compile

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: I can confirm that the soil P-V P-K curves still default to Campbell, and that the plants are controled by the parameter "hydr_htftype_node" in the parameter file.  The default parameter file still uses TFS P-V P-K for plants.
Resolving

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: removed, resolved

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: done, confirmed

biogeophys/FatesHydroWTFMod.F90
rgknox: the linear interpolation code was left in, an issue was created, see #811

biogeophys/FatesHydroWTFMod.F90
rgknox: modifications to linear interpolation were omitted, resolving

biogeophys/FatesHydroWTFMod.F90
rgknox: removed psicap, resolved

biogeophys/FatesHydroWTFMod.F90
rgknox: omitted removal of extrapolations, resolved

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: resolved

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: There are a few options here going forward, we can 1) continue to use the patch vegetation temperature, but pass in the correct patch index, 2) use the soon to be implemented running average feature and use an EMA type vegetation temperature or 3) use the soil temperature (instantaneous or time averaged)
Will create an issue

rgknox:
test nominal: /glade/scratch/rgknox/tests_1117-133121ch

PR #738, opened 2021-04-08
Logistic model of root growth

JunyanDing:
### Description:

1. simulates the rooting depth of cohorts using a logistic model. See attached file for detail
[FATESHydroDynamicRoots.pdf](https://github.com/NGEET/fates/files/6281612/FATESHydroDynamicRoots.pdf)
2. modified the corresponding routines so that the soil layers can have zero roots so that 1) that layer will be bypassed in 1D solver, 2) fine root water content in soil layer deeper than maximum rooting depth will be set to zero so, 3) handle zero (and infinite) situation in calculating rhizosphere shell conductance

 
issue #726
### Collaborators:

Ryan Knox, Chonggang Xu, Charlie Koven

### Checklist:

- [ x] My change requires a change to the documentation.
- [x ] I have read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/master/CONTRIBUTING.md) document.

### Test Results:

FATES baseline hash-tag:

Test Output:


biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -818,6 +825,21 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
...
+    real(r8) :: dbh_max                      ! the maximum dbh a PFT can have as observed                            [m]
+    real(r8) :: dbh_rev                      ! ratio, similar to RWC*   
+
+    real(r8) :: z_fr                         ! rooting depth of a cohort                                             [m]
+    real(r8) :: z_fr_0                       ! the rooting depth of of the sappling, corresponding to dbh_0          [m]
+    real(r8) :: z_fr_max                     ! the maximum rooting depth of a PFT, currently set to the soil depth, but can be a PFT based parameter  
+    real(r8) :: frk                          ! the exponent parameter of the cohort rooting depth function, a PFT based parameter 
+ 
+  
+    ! end of Junyan's addition
```
rosiealice, line 841:
```suggestion
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -818,6 +825,21 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
     real(r8) :: norm                         ! total root fraction used <1
     integer  :: nlevrhiz                     ! number of rhizosphere levels
 
+    ! added by Junyan May 29, 2020
```
rosiealice, line 828:
```suggestion
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -818,6 +825,21 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
...
+    real(r8) :: dbh                          ! the dbh of current cohort                                             [m]   
+    real(r8) :: dbh_0                        ! the dbh of the sappling at recuitment                                 [m]
+
+    real(r8) :: dbh_max                      ! the maximum dbh a PFT can have as observed                            [m]
```
rosiealice, line 832: Do we always want to have a dbh_max? I'm curious about the consequences of imposing this for all possilbe allometries.

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -818,6 +825,21 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
...
+    real(r8) :: dbh_rev                      ! ratio, similar to RWC*
```
rosiealice, line 833: what is rwc*?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -895,7 +928,7 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
     ! alternative cross section calculation
     ! a_sapwood    = a_leaf_tot / ( 0.001_r8 + 0.025_r8 * ccohort%hite ) * 1.e-4_r8
 
-    call CrownDepth(ccohort%hite,crown_depth)
```
rosiealice, line 898: Why remove the call to crown depth here?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -930,22 +963,40 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
     ! Partition the total absorbing root lengths and volumes into the active soil layers
     ! We have a condition, where we may ignore the first layer
     ! ------------------------------------------------------------------------------
-    
+    ! modified by Junyan May 29, 2020
```
rosiealice, line 966:
```suggestion
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -930,22 +963,40 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
     ! Partition the total absorbing root lengths and volumes into the active soil layers
     ! We have a condition, where we may ignore the first layer
     ! ------------------------------------------------------------------------------
-    
+    ! modified by Junyan May 29, 2020 
+    ! norm = 1._r8 - &
```
rosiealice, line 967:
```suggestion
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -930,22 +963,40 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
...
+    !      zeng2001_crootfr(roota, rootb,site_hydr%zi_rhiz(1)-site_hydr%dz_rhiz(1), site_hydr%zi_rhiz(nlevrhiz))
```
rosiealice, line 968:
```suggestion
    !      zeng2001_crootfr(roota, rootb,site_hydr%zi_rhiz(1)-site_hydr%dz_rhiz(1), site_hydr%zi_rhiz(nlevrhiz))
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -930,22 +963,40 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
...
     
     do j=1,nlevrhiz
         
-        rootfr = norm*(zeng2001_crootfr(roota, rootb, site_hydr%zi_rhiz(j),site_hydr%zi_rhiz(nlevrhiz)) - &
-              zeng2001_crootfr(roota, rootb, site_hydr%zi_rhiz(j)-site_hydr%dz_rhiz(j),site_hydr%zi_rhiz(nlevrhiz)))
+        rootfr = norm * (zeng2001_crootfr(roota, rootb, site_hydr%zi_rhiz(j),z_fr) - &
+              zeng2001_crootfr(roota, rootb, site_hydr%zi_rhiz(j)-site_hydr%dz_rhiz(j),z_fr))
         
+        if(JD_debug)then
+            write(fates_log(),*) 'check rooting depth of cohort - Junyan, line 987'
```
rosiealice, line 985:
```suggestion
            write(fates_log(),*) 'check rooting depth of cohort '
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -164,26 +164,23 @@ module FatesPlantHydraulicsMod
...
-  
-
   real(r8), parameter :: thsat_buff = 0.001_r8 ! Ensure that this amount of buffer
                                                ! is left between soil moisture and saturation [m3/m3]
                                                ! (if we are going to help purge super-saturation)
                                              
   logical,parameter :: debug = .false.          ! flag to report warning in hydro
-
-
+  logical,public, parameter :: JD_debug = .false.       ! Junyan added to debug my modifications
```
rosiealice, line 182: Probably change these to using normal debug switch for PR

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1009,11 +1060,16 @@ subroutine UpdateSizeDepPlantHydStates(currentSite,ccohort)
...
     ccohort_hydr%errh2o_growturn_aroot = 0._r8
     do j=1,currentSite%si_hydr%nlevrhiz
-       th_aroot_uncorr(j) = ccohort_hydr%th_aroot(j) * &
-            ccohort_hydr%v_aroot_layer_init(j)/ccohort_hydr%v_aroot_layer(j)
-       ccohort_hydr%th_aroot(j) = constrain_water_contents(th_aroot_uncorr(j), small_theta_num, ft, pm_node(4))
-       ccohort_hydr%errh2o_growturn_aroot = ccohort_hydr%errh2o_growturn_aroot + & 
-            denh2o*cCohort%n*AREA_INV*(ccohort_hydr%th_aroot(j)-th_aroot_uncorr(j))*ccohort_hydr%v_aroot_layer(j)
+      ! check v_aroot >0
+      if (ccohort_hydr%v_aroot_layer(j) > 0) then 
+        th_aroot_uncorr(j) = ccohort_hydr%th_aroot(j) * &
```
rosiealice, line 1065: Maybe add some comments here to describe what is going on.

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1009,11 +1060,16 @@ subroutine UpdateSizeDepPlantHydStates(currentSite,ccohort)
...
+             ccohort_hydr%v_aroot_layer_init(j)/ccohort_hydr%v_aroot_layer(j)
+        ccohort_hydr%th_aroot(j) = constrain_water_contents(th_aroot_uncorr(j), small_theta_num, ft, pm_node(4))
+        ccohort_hydr%errh2o_growturn_aroot = ccohort_hydr%errh2o_growturn_aroot + & 
+             denh2o*cCohort%n*AREA_INV*(ccohort_hydr%th_aroot(j)-th_aroot_uncorr(j))*ccohort_hydr%v_aroot_layer(j)
+      else
+
+      endif ! end checking v_arrot
```
rosiealice, line 1072:
```suggestion
      endif ! end checking v_aroot 
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1460,7 +1522,7 @@ subroutine UpdateH2OVeg(nsites,sites,bc_out)
     type(ed_patch_type), pointer :: currentPatch
     type(ed_cohort_hydr_type), pointer :: ccohort_hydr
     type(ed_site_hydr_type), pointer :: csite_hydr
-    integer :: s
+    integer :: s,ily
```
rosiealice, line 1525:
```suggestion
    integer :: s
    integer :: ily      ! ** add description**
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1483,12 +1545,26 @@ subroutine UpdateH2OVeg(nsites,sites,bc_out)
           do while(associated(currentCohort))
              ccohort_hydr => currentCohort%co_hydr
              !only account for the water for not newly recruit for mass balance
-             if(.not.ccohort_hydr%is_newly_recruited) then 
+             if(.not.ccohort_hydr%is_newly_recruited) then
+               ! check for nan value , Junyan
+               do ily = 1,csite_hydr%nlevrhiz
```
rosiealice, line 1550: There should be an 'isnan' function somewhere that makes this more obvious. Is it likely to generate nan's still?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1483,12 +1545,26 @@ subroutine UpdateH2OVeg(nsites,sites,bc_out)
...
+                  endif                 
+               end do  ! end checking nan
+
+
                 csite_hydr%h2oveg = csite_hydr%h2oveg + &
                      (sum(ccohort_hydr%th_ag(:)*ccohort_hydr%v_ag(:)) + &
                      ccohort_hydr%th_troot*ccohort_hydr%v_troot + &
                      sum(ccohort_hydr%th_aroot(:)*ccohort_hydr%v_aroot_layer(:)))* &
                      denh2o*currentCohort%n
+                if (JD_debug) then
```
rosiealice, line 1562: Do you need this still?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1501,6 +1577,14 @@ subroutine UpdateH2OVeg(nsites,sites,bc_out)
        ! Note that h2oveg_dead is incremented wherever we have litter fluxes
        ! and it will be reduced via an evaporation term
        ! growturn_err is a term to accomodate error in growth or turnover. need to be improved for future(CX) 
+       if (JD_debug) then
```
rosiealice, line 1580: Also here. Is this likely to be triggered?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1522,7 +1606,10 @@ subroutine RecruitWUptake(nsites,sites,bc_in,dtime,recruitflag)
     ! After the root water uptake, is_newly_recruited flag is set to false.
     ! Note, this routine is not accounting for the normal water uptake of new plants
     ! going forward, this routine accounts for the water that needs to be accounted for
-    ! as the plants pop into existance.    
+    ! as the plants pop into existance.  
+    ! Notes by Junyan,  July 16. 2020
```
rosiealice, line 1610:
```suggestion
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1522,7 +1606,10 @@ subroutine RecruitWUptake(nsites,sites,bc_in,dtime,recruitflag)
...
+    !   modify the accessable soil layer equal to z_fr_0
```
rosiealice, line 1611: Is that still tbd?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1539,6 +1626,8 @@ subroutine RecruitWUptake(nsites,sites,bc_in,dtime,recruitflag)
     type(ed_site_hydr_type), pointer :: csite_hydr
     integer :: s, j, ft
     integer :: nstep !number of time steps 
+    real(r8) :: roota !root distriubiton parameter a
+    real(r8) :: rootb !root distriubiton parameter b
```
rosiealice, line 1630:
```suggestion
    real(r8) :: rootb !root distribution parameter b
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1634,14 +1735,16 @@ subroutine ConstrainRecruitNumber(csite,ccohort, bc_in)
     end do
 
     do j=1,csite_hydr%nlevrhiz
+     ! check there is roots in the layer, only proceed when there is roots
```
rosiealice, line 1738:
```suggestion
     ! check there is roots in the layer, only proceed when there are roots 
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -3167,8 +3366,15 @@ subroutine ImTaylorSolve1D(site_hydr,cohort,cohort_hydr,dtime,q_top, &
         ! plant for this cohort takes up, relative to ALL cohorts at the site. Note:
         ! cohort_hydr%l_aroot_layer(ilayer) is units [m/plant]
         ! site_hydr%l_aroot_layer(ilayer) is units [m/site]
-        
-        aroot_frac_plant = cohort_hydr%l_aroot_layer(ilayer)/site_hydr%l_aroot_layer(ilayer)
+ 
+        ! Junyan added if statement to handle zero l_aroot_layer condition
```
rosiealice, line 3370:
```suggestion
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -3167,8 +3366,15 @@ subroutine ImTaylorSolve1D(site_hydr,cohort,cohort_hydr,dtime,q_top, &
...
+        if (site_hydr%l_aroot_layer(ilayer)<nearzero) then
+           l_aroot_layer = 2*nearzero
+        else 
+           l_aroot_layer = site_hydr%l_aroot_layer(ilayer)
+        end if ! end Junyan addition
```
rosiealice, line 3375:
```suggestion
        end if !
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -4153,6 +4387,10 @@ function zeng2001_crootfr(a, b, z, z_max) result(crootfr)
     ! root fraction.
 
     if(present(z_max))then
+       ! If the soil depth is larger than the maximum rooting depth of the cohort, 
+       ! then the cumulative root frection of that layer equals that of the maximum rooting depth
```
rosiealice, line 4391:
```suggestion
       ! then the cumulative root fraction of that layer equals that of the maximum rooting depth  
```

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -4199,35 +4437,49 @@ subroutine shellGeom(l_aroot, rs1, area_site, dz, r_out_shell, r_node_shell, v_s
...
+      do k = 1,nshells
+         if(k == 1) then
+            v_shell(k) = pi_const*l_aroot*(r_out_shell(k)**2._r8 - rs1**2._r8)
+         else
+            v_shell(k) = pi_const*l_aroot*(r_out_shell(k)**2._r8 - r_out_shell(k-1)**2._r8)
+         end if
+      enddo
+
+    else
+    ! set values for zero roots case
```
rosiealice, line 4477: delete this commented out section

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -4199,35 +4437,49 @@ subroutine shellGeom(l_aroot, rs1, area_site, dz, r_out_shell, r_node_shell, v_s
...
+     ! r_out_shell(1:nshells) = 0
+     ! r_node_shell(1:nshells) = 0
+     ! v_shell(1:k) = 0 
+
+    end if ! end line 4439
```
rosiealice, line 4482:
```suggestion
    end if 
```

biogeophys/FatesPlantHydraulicsMod.F90
rosiealice: Also, following up on this, I can't find an instance where dbh_max is used in this PR?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -840,6 +862,16 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
...
     rootb    = prt_params%fnrt_prof_b(ft)
 
+    dbh_max = EDPftvarcon_inst%allom_dbh_max(ft)
+    dbh_0   = EDPftvarcon_inst%allom_dbh_0(ft)
+    z_fr_max = EDPftvarcon_inst%allom_zfr_max(ft)
+
+    z_fr_0 = EDPftvarcon_inst%allom_zfr_0(ft)
+    frk    = EDPftvarcon_inst%allom_frk(ft)
+    dbh      =  ccohort%dbh
+    dbh_rev  = (dbh - dbh_0)/(dbh_max - dbh_0)
```
rosiealice, line 872: Now I see here how dbh_max and dbh_0 are used, to calculate a sort of 'relative size', I wonder whether these should be estimated using the allometry functions that we already have. I worry that adding this many allometric parameters might mean that people tie themselves up in terrible knots when trying to parameterize the model. Further, we are only using dbh_max to calculate the rooting depth, when people might reasonably think that it also actually affects the size of the trees...  Do you think it's reasonable to calculate this metric from the allometry functions, @rgknox ?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -818,6 +825,21 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
...
     integer  :: nlevrhiz                     ! number of rhizosphere levels
 
+    ! added by Junyan May 29, 2020
+    real(r8) :: dbh                          ! the dbh of current cohort                                             [m]   
+    real(r8) :: dbh_0                        ! the dbh of the sappling at recuitment                                 [m]
+
+    real(r8) :: dbh_max                      ! the maximum dbh a PFT can have as observed                            [m]
+    real(r8) :: dbh_rev                      ! ratio, similar to RWC*   
+
+    real(r8) :: z_fr                         ! rooting depth of a cohort                                             [m]
```
rosiealice, line 835: Why is this called z_fr?

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: @rosiealice , I was thinking something similar.  We do have a routine that calculates "derived" parameters (somewhat of an oxymoron).  But they are essentially constants that are derived from user controlled constants.  So dbh_0 would be calculated from the initial cohort height.  Is it possible that dbh_max is similar to the dbh at maximum height parameter?

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: rwc* is the relative water content, the way dbh_rev is calculated in the similar way as RWC* is calculated.

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: I'm thinking through this too.  "fr" must mean fine-root right?  The rooting depth should reflect the maximum depth of all roots (suberized/structural/transporting as well) right?
z is pretty common in refering to a vertical coordinate, maybe "depth" or something else might be more descriptive.  Although we use "z_" do describe the soil depths, which is also consistent with CLM/ELM.  Although, in those cases those are depth arrays of the soil layers.

rosiealice: I get the 'z'. I was being throw off by my mental association of 'fr' with fraction.  Could it be called f_root?

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: The root model hasn't be extensively tested. I think to keep this for a little while in case something goes wrong.

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: @rgknox , you are right fr stands for fine roots. Yes, the rooting depth should include both fine roots and coarse roots. Currently, hydro does not explicitly model the coarse roots fraction and depth, they are treated the same as fine roots. 
I think maybe just call it z_root ?

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: I think dbh_max and dbh at maximum height are different concepts. dbh_max here is the maximum dbh a PFT can possible have in reality without any constrains, while dbh at maximum height, by my understanding, is the dbh the PFT will no longer increase height, but they can still increase dbh, right?

biogeophys/FatesPlantHydraulicsMod.F90
rosiealice: z_root is good too.

biogeophys/FatesPlantHydraulicsMod.F90
rosiealice: Is there a paper that addresses the conversion from dbh to relative size for this purpose? Maybe that would help us elucidate the way in which it should be used. Note that in FATES, there is in fact no maximum DBH. Plants can just keep getting fatter, but generally not taller. So, whatever you set this variable as, it it will, I think, be possible to exceed it. Given the lack of data on rooting depths in general, my suspicion is that we do not know exactly how this scaling should work, and probably can't distinguish between the merits of different formats of relative size. (if not, then, great!)

JunyanDing: @rosiealice , good point. I was just thinking there should be a up limit when the size can no longer increases e.g. from basic biological processes and/or physics. So I borrow the idea of the logistic population growth model, so the dbh_max is analog to the carrying capacity. I should look into the literatures see if there is any studies on that. 
Yes, different formulas of relative size can generate exact the same shape, which is determined by the values of the parameters. 

I am hoping to have a root growth model that best represents the biological and physical mechanisms, in the way the formula represent the processes/interactions and each parameter has biological or physical meaning. Do you have any suggestions/ideas?

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -491,15 +490,20 @@ subroutine InitPlantHydStates(site, cohort)
 !       h_aroot_mean = 0._r8
 
        do j=1, site_hydr%nlevrhiz
-          
-          ! Match the potential of the absorbing root to the inner rhizosphere shell
-          cohort_hydr%psi_aroot(j) = site_hydr%wrf_soil(j)%p%psi_from_th(site_hydr%h2osoi_liqvol_shell(j,1))
+        ! Checking apperance of roots. Only proceed if there is roots in that layer 
+         if(cohort_hydr%l_aroot_layer(j) > 0) then
```
rgknox, line 494: @JunyanDing this change seems reasonable

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: @JunyanDing , could we integrate your changes into the CrownDepth subroutine?
https://github.com/NGEET/fates/blob/master/biogeochem/FatesAllometryMod.F90#L1894-L1914

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -930,22 +963,40 @@ subroutine UpdatePlantHydrLenVol(ccohort,site_hydr)
...
+    ! modified by Junyan May 29, 2020 
+    ! norm = 1._r8 - &
+    !      zeng2001_crootfr(roota, rootb,site_hydr%zi_rhiz(1)-site_hydr%dz_rhiz(1), site_hydr%zi_rhiz(nlevrhiz))
+ 
+   
+    ! set the rooting depth of the cohort, using the logistic functionbelow:   
+    !    z_fr_max/(1 + ((z_fr_max-z_fr_0)/z_fr_0)*exp(-frk*dbh_rev))
+    !  which is constrained by the maximum soil depth:  site_hydr%zi_rhiz(nlevrhiz)
+
+    z_fr = min(site_hydr%zi_rhiz(nlevrhiz), z_fr_max/(1 + ((z_fr_max-z_fr_0)/z_fr_0)*exp(-frk*dbh_rev)))
```
rgknox, line 975: This line here is the key component of the PR, right @JunyanDing ? 
It would be good to put your name and date by this line so folks in the future will know the author.

rgknox:
In summary I think these changes all make sense to me.  The only things I'm wondering right now:
1) Allometry parameters are migrated to the parteh/ folder as per #716, so we might want to move these params with them (we can wait until that PR is integrated)
2) The zeng2001 rooting profile is not currently integrated with the other rooting profile types.  So there is an inconsistency between the root profile for hydraulics and the root profile for biogeochemistry, decomposition and respiration. This is an existing issue, and is not the responsibility of this PR to fix this. But once this PR is integrated, we will need to follow up and integrate some of the functions added here, into the generalzed routines we have for routing depth (found in FatesAllometryMod.F90).

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: @rgknox  yes, do you want me to change that in this pull request or do that after this PR is merged?

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: you are right @rgknox . Thanks for the suggestion. I will do that.

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: I wanted to check in on this particular point. @JunyanDing.  As @rosiealice mentions, the growth equations in the core fates scaling algorithm don't really understand the concept of a maximum diameter.   I see two paths that are viable going forward.  1) We use "the diameter at maximum height" (parameter already exists)  instead of new parameter that defines the maximum diameter for relative root depth scaling.  Alternatively, 2) if we do want to have a separate parameter that defines the maximum carrying capacity,  I suppose it would be ok as long as we use a naming convention and use of metadata to explain its for the rooting depth and not used to actually constrain plant diameter. Part of my job is reduce the overall build-up of unused or duplicative parameters in the fates model (which is inevitable), so I'm always going to push for alternative 2.  :)

Also, @JunyanDing : How about dbh_0.  Can we get rid of this parameter and just use the diameter the cohorts start with when they are recruited (ie their smallest possible diameter)?

main/EDPftvarcon.F90
```diff
@@ -199,6 +199,16 @@ module EDPftvarcon
      real(r8), allocatable :: hydr_avuln_gs(:)      ! shape parameter for stomatal control of water vapor exiting leaf 
      real(r8), allocatable :: hydr_p50_gs(:)        ! water potential at 50% loss of stomatal conductance
 
+ !  PFT specific parameters for hydro dynamic roots
```
rgknox, line 202: comments with descriptions of each parameter following their declarations please
include units as well if possible

biogeophys/FatesPlantHydraulicsMod.F90
JunyanDing: @rgknox I like your idea of alternative 2 for max dbh,  :-) .maybe call it dbh_max_zroot?  As for dbh_0, I agree, we can use the smallest possible diameter here.

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: Working on resolving conflicts with master.  I'm going to remove this in my pass through @JunyanDing .  Let me know if you want to re-address this.

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: the naming convention was resolved via #759

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: several writes converted to debug, several removed

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: As far as I know, this is the dbh when the roots reach their maximum rooting depth.  It looks like these local variables need their definitions cleaned up a little bit.

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: As far as I can tell, this should be implicitly true, since rooting depth has already informed each cohorts array of fine absorbing roots with depth: ccohort_hydr%v_aroot_layer(:)

rgknox:
Site level comparison between this branch and master, similar settings, BCI. 
[rootdepth-v2_plots.pdf](https://github.com/NGEET/fates/files/7318912/rootdepth-v2_plots.pdf)
Note, while these results are similar, I do not expect b4b results against base in regression testing. We have made small modifications to error caps.

PR #742, opened 2021-05-12
Two new command-line scripts for comparing parameter files

ckoven:
Added two scripts: one to compare the trait values between two PFTs in a parameter file, and one to diff entire parameter files (or really any pair of netcdf files).  I wrote these because I got tired of the multistep process of converting .nc to .cdl files for comparisons, and the resulting directory clutter from that. These scripts just make, compare, and then delete a bunch of temporary files so as to avoid that.  Since these make working with multiple FATES runs slightly easier, I thought others might want as well.

### Description:
no changes to actual FATES code.

### Collaborators:
just me so far.

doesn't touch compile or runtime code, so no answer changes.

### Checklist:

### Test Results:
n/a

FATES baseline hash-tag:

Test Output:
n/a


tools/pftdiff
```diff
@@ -0,0 +1,33 @@
...
+	   echo "script to compare two PFTs in a FATES parameter file.  takes three arguments: "
+	   echo "first argument is the parameter file name"
+	   echo "second argument is the first pft number (PFT numbering starts with 1)"
+	   echo "third argument is the second pft number (PFT numbering starts with 1)"
+         exit;;
+   esac
+done
+
+
+tempfile1=$(mktemp)
```
rgknox, line 15: @ckoven , is mktemp something the user should have set in their environment?  Do these four tempfiles just need some nominal name? It seems like they should have a different name though, right?

rgknox: ok, I see, mktemp is a system command.  I'm surprised thought that it works with this syntax. Have you printed out the temporary file names to make sure that they are unique and defined?

ckoven: yep.  they look something like `tmp.cGZMIIuLIH` and get put in the local directory.

PR #751, opened 2021-06-17
Fix decay restarts

rgknox:
### Description:

This is a bugfix. These changes are required to ensure b4b comparisons between a base and its restart.  This is due to a recent regression from a very recent tag.

This will change the answers following a restart, particularly those answers related to soil carbon, as this will modify the litter flux slightly.  

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/FatesSoilBGCFluxMod.F90
+++ b/biogeochem/FatesSoilBGCFluxMod.F90
@@ -1022,15 +1022,34 @@ contains
           
        end select
 
-       ! Add efflux to the litter pool. kg/ha/day -> kg/m2/day                                                                             
-       do id = 1,nlev_eff_decomp
-          flux_lab_si(id) = flux_lab_si(id) + &
-               sum(csite%flux_diags(el)%nutrient_efflux_scpf)*surface_prof(id)*area_inv
-       end do
-
-       
        currentPatch => csite%oldest_patch
        do while (associated(currentPatch))
+
+          ! If there is any efflux (from stores overflowing)
+          ! than pass that to the labile litter pool
+          
+          currentCohort => currentPatch%tallest
+          do while(associated(currentCohort))
+             if(.not.currentCohort%isnew)then
+                if(element_list(el).eq.carbon12_element) then
+                   efflux_ptr => currentCohort%daily_c_efflux
+                elseif(element_list(el).eq.nitrogen_element) then
+                   efflux_ptr => currentCohort%daily_n_efflux
+                elseif(element_list(el).eq.phosphorus_element) then
+                   efflux_ptr => currentCohort%daily_p_efflux
+                end if
+                
+                ! Unit conversion
+                ! kg/plant/day * plant/ha * ha/m2 -> kg/m2/day
+                
+                do id = 1,nlev_eff_decomp
+                   flux_lab_si(id) = flux_lab_si(id) + &
+                        efflux_ptr * currentCohort%n* AREA_INV * surface_prof(id)
+                end do
+             end if
+             currentCohort => currentCohort%shorter
+          end do
+
 
           ! Set a pointer to the litter object
           ! for the current element on the current
--- a/biogeochem/FatesSoilBGCFluxMod.F90
+++ b/biogeochem/FatesSoilBGCFluxMod.F90
@@ -997,9 +997,9 @@ contains
        
        select case (element_list(el))
        case (carbon12_element)
-          bc_out%litt_flux_cel_c_si(:) = 0._r8
-          bc_out%litt_flux_lig_c_si(:) = 0._r8
-          bc_out%litt_flux_lab_c_si(:) = 0._r8
+          bc_out%litt_flux_cel_c_si(:) = 0.0_r8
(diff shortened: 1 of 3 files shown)


ckoven (approved):
this looks good to me, thanks @rgknox for catching.

rgknox:
Tests on cheyenne:

/glade/scratch/rgknox/clmed-tests/fix-decay-ctsm5.1.dev041

One noted difference, the PRT2 test generated very small differences with the base, see here:

/glade/scratch/rgknox/ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-FatesPRT2.GC.20210618_075109_ll9h81/run/ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-FatesPRT2.GC.20210618_075109_ll9h81.clm2.h0.2000-01-01-00000.nc.cprnc.out

The largest rms difference was on the order of e-21.  This might be fine, but I'd like to think through this a little more.

biogeochem/FatesSoilBGCFluxMod.F90
```diff
@@ -997,9 +997,9 @@ subroutine FluxIntoLitterPools(csite, bc_in, bc_out)
        
        select case (element_list(el))
        case (carbon12_element)
-          bc_out%litt_flux_cel_c_si(:) = 0._r8
-          bc_out%litt_flux_lig_c_si(:) = 0._r8
-          bc_out%litt_flux_lab_c_si(:) = 0._r8
+          bc_out%litt_flux_cel_c_si(:) = 0.0_r8
+          bc_out%litt_flux_lig_c_si(:) = 0.0_r8
+          bc_out%litt_flux_lab_c_si(:) = 0.0_r8
```
rosiealice, line 1002: Embarrassing question I've never got my head around. Does this matter? Like, are there circumstances where you get a different answer with a 0.0_r8 than a 0._r8?

main/FatesRestartInterfaceMod.F90
```diff
@@ -944,6 +946,17 @@ subroutine define_restart_vars(this, initialize_variables)
            units='kg/m2', veclength=num_elements, flushval = flushzero, &
            hlms='CLM:ALM', initialize=initialize_variables, ivar=ivar, index = ir_seedgerm_litt)
 
+
+    call this%RegisterCohortVector(symbol_base='fates_seed_decay', vtype=cohort_r8, &
+            long_name_base='seed bank (non-germinated)',  &
```
rosiealice, line 951: Is this 'long name' accurate? Like, if it's the seed decay rate, it's not actually in the seed bank any more?

main/FatesRestartInterfaceMod.F90
```diff
@@ -944,6 +946,17 @@ subroutine define_restart_vars(this, initialize_variables)
...
+            units='kg/m2', veclength=num_elements, flushval = flushzero, &
+            hlms='CLM:ALM', initialize=initialize_variables, ivar=ivar, index = ir_seed_decay_litt)
+
+    call this%RegisterCohortVector(symbol_base='fates_seedgerm_decay', vtype=cohort_r8, &
+           long_name_base='seed bank (germinated)',  &
+           units='kg/m2', veclength=num_elements, flushval = flushzero, &
```
rosiealice, line 957: Relatedly, while I know this isn't really related to this bug fix, without delving into the code a bit more, I don't think it's super clear from the names how these variables differ.

rosiealice (approved):
Comments here entirely related to my own slight incomprehension of how this code has evolved. Otherwise it looks fine.

biogeochem/FatesSoilBGCFluxMod.F90
rgknox: nope, I don't think it matters.  This was just an artifact too, I didn't make the change for stylistic reasons.  This is what happened.  I was testing the code by forcing a non zero decomposition flux into E3SM with a value of 1.0. I wanted to see if E3SM was closing its carbon budget with a constant and known input flux from FATES. When I reverted it, I just added a zero.

main/FatesRestartInterfaceMod.F90
rgknox: good catch, that is not the right long name. Its a flux rate. I stole the long_name from the pool variable that was already there and forgot to update the name.  Will do.

main/FatesRestartInterfaceMod.F90
rgknox: will update this as well, thanks

rgknox:
I reverted a portion of the code where the labile fluxes to the HLM were calculated. The diagnostic site level array with effluxes is populated after the restart when we set the fragmentation fluxes, so it should work. I reverted it to use the cohort level effluxes, to preserve b4b with the test which seems safer.  I think the difference could be explained by order of operations rounding errors (which are acceptable).

rgknox:
retesting

rgknox:
tests pass: /glade/scratch/rgknox/clmed-tests/fix-decay-ctsm5.1.dev041-v4

Issue #747, opened 2021-05-21
Leaf layer resolution impacts total absorbed radiation


rosiealice:
Looking at the SP mode simulations, it appears that FATES has a low bias in the absorbed radiation fluxes (and a correspondingly high bias in reflected radiation), compared to CLM5-SP runs with the same LAI forcing, no stem area index (for the sake of avoiding #744)  and optical properties. 

The dinc_ed parameter is hard-wired, here:
https://github.com/NGEET/fates/blob/516c1845d5fe18eacbb7e8e7d70596b4eff364aa/main/EDTypesMod.F90#L123

Modifying it from 1.0 to 0.2 reduces the bias from ˜0.15 (version 4 below) to <0.06  (version 5 below). 

Obviously, this is non-ideal.  The potential fixes include:

1. Having a dinc_ed so small that we resolve the radiation environment perfectly 
2. Adding in a scheme whereby dinc changes (gets larger) as a function of canopy depth, to better resolve the rapid changes in light in the upper layers. 
3. We change to/implement a 2-stream scheme within cohorts in FATES. 

'1' is just really for the sake of completeness and/or testing. Making dinc uniformly <0.2 through the canopy would be far too expensive.   
'2' is less of a time sink than '3'., and is probably what we need to do in the interim. 
'3' would allow one to replicate 'big-leaf' physiology in the FATES code and thus ultimately reduce duplication. 

˜![SP_2rad_SABV_5 (3)](https://user-images.githubusercontent.com/12954691/119110982-ecd57a80-ba22-11eb-91ea-bb8af054f4f5.png)

![SP_2rad_SABV_4 (4)](https://user-images.githubusercontent.com/12954691/119110482-7e90b800-ba22-11eb-9bbc-7f9712f0a294.png)

walkeranthonyp:
There's a further hybrid option of those you list above, but it's going to add complexity. 

4. Hybrid option(s): Different dinc_ed for RT and for calculation of photosynthesis. e.g. calculating RT with dinc_ed of 0.2 or smaller, and photosynthesis with dinc_d of 1 ... or perhaps variable by canopy layer. If RT calculation is still too expensive Gordon's textbook has a tridiagonal solver for the Norman scheme (tho I recognise that's not trivial to add with with this awesome fates RT code!). 

[...]

Issue #272, opened 2017-09-15
large share of memory footprint in two variables


rgknox:
We have two cohort arrays that are dimensioned by nlevleaf:
```
     real(r8) ::  ts_net_uptake(nlevleaf)              ! Net uptake of leaf layers: kgC/m2/s
     real(r8) ::  year_net_uptake(nlevleaf)            ! Net uptake of leaf layers: kgC/m2/year
```
nlevleaf is quite large, default = 40

This is the equivalent of 80 cohort level variables that are scalar values.  These two variables may be taking up more than half of the total memory.  We should investigate how to reduce/remove these arrays.

rgknox:
Sorry for the late response @rosiealice , this probably made my brain hurt trying to figure this out.

@rosiealice, regarding your point on the bottom layer tracking issue.  Thinking out loud here.  I guess the uptake rate we would track would  be for whatever the lowest layer is for the fully flushed plant at its target (or carrying capacity), at the current trimming rate.  If a plant is constantly at a leaf area/biomass that does not reach this layer, than I suppose we would not be updating the uptake rate in that lowest layer, and wouldn't need to because the plant doesn't even fill out and wouldnt be relevent anyway.

Now, I guess if the plant grows in stature, and the index of the bottom leaf layer gets larger as it grows;   I'm now asking myself if the uptakes that have been tracked at the lower layers, are still relevant/meaningful/accurate to influence the decision on how to trim with respect to the new and larger lowest leaf layer index.

If the plant was not productive in lower leaf layers as a smaller plant, I would suppose that information would still be relevant as the plant gets larger, and would be meaningful as we decide to trim leaf layers at its larger stature. 

This assumes that it has not jumped into a new canopy layer (but the vectorized uptake array would not catch this either).

ckoven:
following up on @rgknox's update and responses to @rosiealice's comment https://github.com/NGEET/fates/issues/272#issuecomment-332398625: in general the bottom layer shouldn't change with growth, via the same logic laid out in https://github.com/NGEET/fates/issues/420#issue-364630134, which says that, so long as the bleaf and canopy area use the same exponent, then LAI should be invariant with tree size.  This isn't strictly true in the understory as the SLA scaling could change due to things happening above, even in that case, the inexactness arising from only tracking the carbon balance of the lowermost expressed leaf layer shouldn't be very large.

So I support trying to see if only tracking the lowermost layer works.

rgknox:
[...]

PR #752, opened 2021-06-17
Variable VAI bin widths

### LINKED ISSUE #492: dynamic leaf layer depth system
Following up on a discussion started in PR: #487; going off an idea proposed by @walkeranthonyp :

The proposal is to change our leaf layering system to one that does not have static leaf layer depths.  By having constant leaf layer depth, we are forced to maintain some large arrays and scratch arrays that hold leaf-layer quantities.

Proposal:

We will have a more reasonable maximum number of leaf layers; lets say 10.  We will also set a minimum bin size, say 0.05 m2/m2.  The leaf layer widths will be the maximum 1) of the total crown LAI / number of layers (10), and 2) the minimum we set.   The actual number of layers will then be dynamic and range somewhere between 0 and the maximum (10 in this example).  We will also maintain the layer leaf layer width as a single cohort level variable.

ckoven:
This code allows for an exponential spacing to bin widths in the VAI (LAI+SAI) calculations.  This should allow for higher resolution of the radiative transfer code at the top of the canopy where light levels are higher, but wider bins at the bottom of the canopy where light levels are lower.  In principle this should do a few different things: 
- improve the accuracy of the model for a given number of leaf layers by putting the resolution where the light intensities and intensity gradients are highest (which should hopefully ameliorate #747)
- improve the stability of the model with respect to parameter perturbations by allowing larger total maximum LAI for a given number of layers (this should help with #420)
- reduce the memory load by not having to carry around a lot of leaf layers that are rarely used but are needed to prevent VAI from exceeding the total size of the array (this should help with #272)

In discussions with @walkeranthonyp @rgknox @rosiealice we were looking at some different layering options in MAAT.  @walkeranthonyp did some initial analyses in MAAT, and more needs to be done on that front to find a good layering that balances cost and accuracy, and to ensure that the same dynamics are at work in FATES. 

In the current code, the number of leaf layers is hard-coded in the fortran, but the width of the top VAI bin, as well as the amount by which each subsequent bin is wider than the previous are both moved to the parameter file in order to facilitate experiments to determine what works best.  The current default values basically keep the old layering scheme, by setting the width of the top bin width to 1 VAI unit and no change in bin thickness with depth.

fixes #492, since we've decided that a vertically-varying leaf layering system is preferable to a dynamic leaf-layering system.

This has some necessary but minimal HLM-side changes.  Currently the CTSM branch is https://github.com/ESCOMP/CTSM/compare/master...ckoven:variable_dinc_2?expand=1 and it mainly adds the VAI bin edges to the history file metadata.

Will update this PR to include CTSM and ELM PRs once it is tested and gets closer to readiness.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


rgknox:
Since we plan on adding a Leaf photosynthetic temperature acclimation timescale parameter as well, it would be nice if these parameter file additions are all grouped together.  Makes it easier on existing users to update their files..
Maybe we could migrate the new parameters defined here, as well as the API changes, to PR #724.  Then, this PR would not require an API change and synchronization with ctsm/e3sm.

main/FatesInterfaceMod.F90
```diff
@@ -795,7 +799,16 @@ subroutine SetFatesGlobalElements(use_fates)
             max_comp_per_site = 1
          end if
             
+         ! calculate the bin edges for radiative transfer calculations
```
rgknox, line 802: @ckoven , while I feel this is as good a place as any, why did you decide to fill these arrays here? (maybe the answer is, "its as good a place as any...")

ckoven: Because I couldn't think of  a better place to do it?

rgknox:
@ckoven @rosiealice  I see that these methods update currentPatch%elai_profile(L,ft,iv) and the like, which are the key arrays that drive the radiative transfer model.  It seems, at a glance at least, that the radiative transfer model should be able to handle variable bin widths.  Is this a definite, a probably, or something we should test further?  I note that in the default parameter file, we maintain constant bin widths which seems like a good place to start.

glemieux:
Note: the ctsm-side update to remove the `dinc_ed` variable that will be made obsolete by this PR will be handled as part of ctsm5.1.dev062 tag: https://github.com/ESCOMP/CTSM/pull/1540#issuecomment-959687078

glemieux:
All expected tests PASS initial fates suite regression tests.  There are four cases with DIFFs when compared against the baseline (all others are b4b):

- `SMS_Lm13.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_gnu.clm-FatesColdDef`
- `SMS_Lm13.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdDef`
- `SMS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdDefReducedComplexSatPhen`
- `SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdDefHydro`

The hydro run is resulting in only very small `FATES_ERRH2O_SCPF` errors.  Both compiler versions of the 13 month sms tests start out with `CBALANCE_ERROR_FATES` and `ERROR_FATES` early in the run on the order of `10E-06` and then eventual start reporting other errors about 1.5 months into the run.  The SatPhen sms test starts to diverge in `C_STOMATA` with modest differences about halfway through the first month and then starts to diverge by small differences in other variables soon after.

Results can be found here: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr752-vaibins`

ckoven:
Updating this to note that I've compared the model dynamics with this and the current main branch and, while not bit-for-bit, the differences appear to be roundoff-level, since the new parameters that control for bin spacing are the same as in the old code.  If I update the code to decrease the bin spacing, then differences do emerge as expected.  So I think we should integrate this now, which would then allow us to explore the costs and benefits of different bin spacings as a future step.

rosiealice:
I am getting to the point where integrating this would be useful, fwiw...

glemieux:
This PR now corresponds with ctsm PR https://github.com/ESCOMP/CTSM/pull/1562.

glemieux:
Update: I've retested this after merging ctsm master into https://github.com/ESCOMP/CTSM/pull/1562.  The same subset of regression test are not b4b and have similar results as discussed in https://github.com/NGEET/fates/pull/752#issuecomment-960309178.  As such, I think we will be good to integrate this once the `aux_clm` test for github.com/ESCOMP/CTSM/pull/1562 is complete (which is underway).

glemieux:
Final re-test after merging #821 into this PR:

Izumi:
All expected `fates` suite tests pass B4B except for:
`SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruRsGs.izumi_nag.clm-FatesColdDefHydro` 
which is an expected DIFF per the comment discussions above.  
Test location: `/scratch/cluster/glemieux/ctsm-tests/tests_vai-bin-fates-ersfix`

Cheyenne: 
The set of DIFFs has expanded slightly to include the other hydro tests with the same order of magnitude errors, so this is acceptable in my view.  Additionally, now that the ERS sp mode test is running that is showing similar diffs to the SMS sp mode test.  All other expected `fates` suite test pass B4B.
Test lcoation: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_vaibin-ctsmdev070-fates-ersfix`

Issue #755, opened 2021-06-22
reduce loop iterations in history hifreq


rgknox:
See here, we iterate through all possible canopy and leaf layers during calculations for radiation diagnostics:

https://github.com/rgknox/fates/commit/2105774598250b86e7948cf1dd3f8c9d4524593a

This normally wouldn't be such a big deal, but we do call this routine on the high frequency (30 minute) step.

I will make a formal PR on this commit once we integrate #751

PR #758, opened 2021-06-24
reduce loop bounds on hi-frequency radiation history updates

rgknox:
### Description:

Changed loop bounds in history writing to use actual rather than maximum canopy layers and leaf layers.

fixes: #755 

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -3682,8 +3682,8 @@ end subroutine flush_hvars
 
             ! summarize radiation profiles through the canopy
             do ipft=1,numpft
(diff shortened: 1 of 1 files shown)


rgknox:
Note: I did check through the call sequence, particularly during restarts, to make sure that cpatch%ncl_p and
cpatch%ncan(ican,ipft) had been updated and set before this call.
Also, in earlier tests, this gave me b4b answers with the base.

ckoven (approved):
Looks good to me. Thanks @rgknox!

rgknox:
Tests pass, b4b:
/glade/scratch/rgknox/tests_1010-151719ch

Issue #753, opened 2021-06-21
New Parameters Inbound


rgknox:
There are few new parameters that are imminent. 

See the following PRs:

#752 
 - scalar:  ED_val_vai_top_bin_width
 - scalar: ED_val_vai_width_increase_factor

#724
 - Leaf photosynthetic temperature acclimation timescale

#738 
     real(r8), allocatable :: allom_dbh_max(:)  
     real(r8), allocatable :: allom_dbh_0(:)
     real(r8), allocatable :: allom_zfr_max(:)
     real(r8), allocatable :: allom_zfr_0(:)
     real(r8), allocatable :: allom_frk(:)

#737
    real(r8), allocatable :: hydr_alpha_vg(:)      ! capilary length parameter in van Genuchten model 
     real(r8), allocatable :: hydr_m_vg(:)          ! pore size distribution, m in van Genuchten 1980 model range (0,1)
     real(r8), allocatable :: hydr_n_vg(:)          ! pore size distribution, n in van Genuchten 1980 model range >2
     real(r8), allocatable :: hydr_k_lwp(:)         ! inner leaf humidity scaling coefficient 

#736  
real(r8), allocatable :: hydr_k_lwp(:)         ! inner leaf humidity scaling coefficient

#703
real(r8), allocatable :: hlm_pft_map(:,:)           ! Mapping from HLM PFTs to FATES PFTs in fixed biogeog mode. 

It would be nice to consolidate these into one PR to help people with updating their code.  I will work with authors to finalize the list of parameters. Will roll this out before the referenced PRs (except for SP, if SP is ready first, I think we should just let SP do its thing)

@ckoven
@JunyanDing 
@rgknox 
@rosiealice 
@glemieux

ckoven:
As we will want to update the leaf maintenance respiration model shortly to also allow acclimation (PR #729), we should at a minimum figure out what the parameters for that will be and include them as well here.  Also which of these updates will require switches to turn on or off?  I imagine at a minimum, a switch to choose acclimating/non-acclimating photosynthesis for #724, a switch to turn on/off the size-dependent rooting depth for #738, anything else?

rgknox:
As an option, for the acclimation timescale, -1 or 0 could be a special value flagging that its not used.

For the rooting depths, by setting the dbh at maximum rooting depth to zero and the maximum rooting depth to the bottom of the soil column it should force cohorts to always have deep roots (current hypothesis).

ckoven:
[...]

Issue #729, opened 2021-03-22
Update FATES respiration model


ckoven:
Hi All-
in various threads and discussions, the topic of the (autotrophic) respiration model in FATES has come up a bunch recently.  right now there are a few distinct issues:
1. sapwood respiration scaling continues to have problems related to sapwood allometry (e.g. #392); Xinyuan Wei is looking into this right now so I won't get into it further here.
2. we'd like to include respiration acclimation, as in CLM5 and other models, following Atkin 2008 model or one of its descendants.
3. Possibly we should move away from using a Q10 model for short-term respiration responses to temperature to something more data-constrained.
4. FATES currently uses a relatively low default value of the growth respiration parameter (0.11 as here: https://github.com/NGEET/fates/blob/master/parameter_files/fates_params_default.cdl#L862-L863), the value of that was lowered during the CLM4->CLM5 development pathway (from a prior value of 0.3 in CLM4, according to the CLM4 tech note) when the Atkins respiration model was added, but in FATES we hadn't done the second part.  This may be a reason why FATES's NPP tends to be on the high side in initial uncalibrated runs (and possibly LAI as well because of leaf optimization logic).

I'd like to propose that, for 2-4 above, we add an option to include other respiration forms besides the current one based on Ryan et al 1991.  In doing a cursory scan through the literature, it seems like the potential top candidate is this model from an Atkin 2017 book chapter (https://link.springer.com/chapter/10.1007%2F978-3-319-68703-2_6), which combines the long-term acclimation model that CLM5 and others use, with a short-term temperature sensitivity model based on Heskel et al 2016 (https://doi.org/10.1073/pnas.1520282113).  I think this is also the formulation that JULES is now using (https://www.nature.com/articles/s41467-017-01774-z).  It is however different from the acclimating respiration model in CLM5, which still uses a Q10 for the short-term temperature sensitivity, per @danicalombardozzi's paper https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2015GL065934

A few questions though: 
1. Are there other respiration models that we should be considering using as well?
2. All the above applies to leaves only, should we continue to use the existing scheme for sapwood and fine roots (as CLM5 does)?
3. Is anyone else working on this at present?  If not, I was thinking of trying to put in a basic representation of this so that we can start to play around with it in our global calibration efforts.

danicalombardozzi:
[...]

PR #759, opened 2021-06-25
new parameters

rgknox:
### Description:

This set of changes responds to #753 and intends to group a bunch of planned additions to the parameter file into one merge.

Accompanies: https://github.com/ESCOMP/CTSM/pull/1414

### Collaborators:

@JunyanDing 
@ckoven
@rosiealice 
@glemieux 
@xuchongang 

no changes expected, this should be b4b

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeophys/FatesPlantHydraulicsMod.F90
+++ b/biogeophys/FatesPlantHydraulicsMod.F90
@@ -47,6 +47,7 @@ module FatesPlantHydraulicsMod
   use EDParamsMod       , only : hydr_kmax_rsurf2
   use EDParamsMod       , only : hydr_psi0
   use EDParamsMod       , only : hydr_psicap
+  use EDParamsMod       , only : hydr_wtftype_node
   
   use EDTypesMod        , only : ed_site_type
   use EDTypesMod        , only : ed_patch_type
@@ -190,12 +191,12 @@ module FatesPlantHydraulicsMod
        __FILE__
 
 
-  integer, public, parameter :: van_genuchten_type      = 1
-  integer, public, parameter :: campbell_type           = 2
-  integer, public, parameter :: tfs_type                = 3
+  integer, public, parameter :: van_genuchten_type      = 2
+  integer, public, parameter :: campbell_type           = 3
+  integer, public, parameter :: tfs_type                = 1
   
-  integer, parameter :: plant_wrf_type = tfs_type
-  integer, parameter :: plant_wkf_type = tfs_type
+  !integer, parameter :: plant_wrf_type = tfs_type
+  !integer, parameter :: plant_wkf_type = tfs_type
   integer, parameter :: soil_wrf_type  = campbell_type
   integer, parameter :: soil_wkf_type  = campbell_type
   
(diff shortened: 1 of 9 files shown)


JunyanDing (commented):
@rgknox These changes all look good to me.

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -190,12 +191,12 @@ module FatesPlantHydraulicsMod
        __FILE__
 
 
-  integer, public, parameter :: van_genuchten_type      = 1
-  integer, public, parameter :: campbell_type           = 2
-  integer, public, parameter :: tfs_type                = 3
+  integer, public, parameter :: van_genuchten_type      = 2
+  integer, public, parameter :: campbell_type           = 3
+  integer, public, parameter :: tfs_type                = 1
```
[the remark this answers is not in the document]
ckoven: also, should we list here which of these are valid for which types of media?

parameter_files/fates_params_default.cdl
```diff
@@ -124,6 +132,21 @@ variables:
 		fates_allom_stmode:units = "index" ;
 		fates_allom_stmode:long_name = "storage allometry function index." ;
 		fates_allom_stmode:possible_values = "1: target storage proportional to trimmed maximum leaf biomass." ;
+	double fates_allom_zroot_k(fates_pft) ;
+		fates_allom_zroot_k:units = "unitless" ;
+		fates_allom_zroot_k:long_name = "scale coefficient of logistic rooting depth model (NOT USED)" ;
+	double fates_allom_zroot_max_dbh(fates_pft) ;
+		fates_allom_zroot_max_dbh:units = "cm" ;
+		fates_allom_zroot_max_dbh:long_name = "dbh at which maximum rooting depth is defined (NOT USED)" ;
```
ckoven, line 140:
```suggestion
		fates_allom_zroot_max_dbh:long_name = "dbh at which a plant reaches the maximum value for its maximum rooting depth (NOT USED)" ;
```

parameter_files/fates_params_default.cdl
```diff
@@ -124,6 +132,21 @@ variables:
...
+	double fates_allom_zroot_max_z(fates_pft) ;
+		fates_allom_zroot_max_z:units = "m" ;
+		fates_allom_zroot_max_z:long_name = "the maximum rooting depth defined at dbh = fates_allom_zroot_max_dbh (NOT USED). note: max_z=min_z=large, sets rooting depth to soil depth" ;
+	double fates_allom_zroot_min_dbh(fates_pft) ;
+		fates_allom_zroot_min_dbh:units = "cm" ;
+		fates_allom_zroot_min_dbh:long_name = "dbh at which the rooting depth for a recruit is defined (NOT USED)" ;
```
ckoven, line 146:
```suggestion
		fates_allom_zroot_min_dbh:long_name = "dbh at which the maximum rooting depth for a recruit is defined (NOT USED)" ;
```

ckoven (commented):
looks good, thanks @rgknox.

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -190,12 +191,12 @@ module FatesPlantHydraulicsMod
...
   
-  integer, parameter :: plant_wrf_type = tfs_type
-  integer, parameter :: plant_wkf_type = tfs_type
+  !integer, parameter :: plant_wrf_type = tfs_type
+  !integer, parameter :: plant_wkf_type = tfs_type
```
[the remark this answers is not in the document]
rgknox: yes

main/EDParamsMod.F90
```diff
@@ -54,11 +64,21 @@ module EDParamsMod
...
+
+   ! Switch that defines the current pressure-volume and pressure-conductivity model
+   ! to be used at each node (compartment/organ)
+   ! 1  = Christofferson et al. 2016 (TFS),   2 = Van Genuchten 1980
+   integer, protected,allocatable,public :: hydr_wtftype_node(:) 
+   
+   character(len=param_string_length),parameter,public :: ED_name_vai_top_bin_width = "fates_vai_top_bin_width"
+   character(len=param_string_length),parameter,public :: ED_name_vai_width_increase_factor = "fates_vai_width_increase_factor"
+   character(len=param_string_length),parameter,public :: ED_name_photo_temp_acclim_timescale = "fates_photo_temp_acclim_timescale"
+   character(len=param_string_length),parameter,public :: ED_name_hydr_wtftype_node = "fates_hydr_wtftype_node"
```
rosiealice, line 81: But why is it called 'wtf' type?

parameter_files/fates_params_default.cdl
```diff
@@ -11,6 +11,7 @@ dimensions:
 	fates_pft = 12 ;
 	fates_prt_organs = 4 ;
 	fates_string_length = 60 ;
+	fates_hlm_pftno = 14 ;
```
rosiealice, line 14: Can one add comments to this part of the parameter file? If so, I think it would be useful to make clear what this is, lest people are confused bout  its meaning  in relation to the FATES pft number above

parameter_files/fates_params_default.cdl
```diff
@@ -244,6 +271,15 @@ variables:
 	double fates_hydr_thetas_node(fates_hydr_organs, fates_pft) ;
 		fates_hydr_thetas_node:units = "cm3/cm3" ;
 		fates_hydr_thetas_node:long_name = "saturated water content" ;
+	double fates_hydr_vg_alpha_node(fates_hydr_organs, fates_pft) ;
```
rosiealice, line 274: Do we have a way of indicating in that you only need to define these if using van Genuchten mode? I can imagine it freaking out plant trait measuring folk innocently browsing through the file!

rosiealice (approved):
Looks good... When we bring these new hydraulic and rooting variables online (i.e. when they function)  we should be careful to add lots of notes to the parameter file about when they -aren't- needed!

biogeophys/FatesPlantHydraulicsMod.F90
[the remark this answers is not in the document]
rgknox: I updated the text here.  But the numbers were re-ordered to match the description in the parameter file.  Also, we only support vg and tfs pv/pk hypothesese for plants, campbel could be done but it would require its own parameters and we don't have those in the parameter file.  So as of now, we only use campbel for soil, and it is hard-coded to be used by default there (since the HLMs use it also).

main/EDParamsMod.F90
rgknox: water transfer function types, a generalized way of saying pressure volume and pressure conductance relationships. It has no precedent.

parameter_files/fates_params_default.cdl
rgknox: I'm not exactly sure how one would go about associating metadata with a dimension size, or to add a comment that would preserve its location next to the dimension.  It might be possible though.

parameter_files/fates_params_default.cdl
rgknox: updated text to this effect

main/EDParamsMod.F90
ekluzek: Probably best to not use WTF in our culture. Unless you want to be teased a bit. I know of an organization that changed their acronym away from it.

rgknox: ok, yeah, I see it now.  I will change to HTF.

rgknox:
Added the following parameters (see in-code documentation):

dev_arbitrary_pft
dev_arbitrary
theta_cj_c3
theta_cj_c4

The former allows developers to use the parameter for whatever they want without having to create a new parameter. If they wish to formaly introduce a new parameter after testing, they can do that later during the integration process. This should make hypothesis testing a little easier.

theta_cj is the smoothing parameter in the photosynthesis code.  After testing for b4b I will make this parameter over-ride and remove the hard-coded smoothing parameters.

rgknox:
passes B4B on FATES tests when reverting the shrub respiration parameter to 1, and not activating the theta_cj parameters. 

/glade/scratch/rgknox/clmed-tests/test-b4b-new-params-2

ckoven:
@rgknox -- I was thinking about this more and I actually think that now is the time to put switches in for choosing between different models of both the leaf photosynthetic temperature dependence and the leaf maintenance respiration.  since both of these will lead to big changes for existing simulations having these in place will enable proceeding while enabling backwards compatibility. for now they can have only the one valid value (1), but we'll be adding to them shortly.

not sure if its easier for me to do a PR to your branch or you to add them yourself.  proposed names and metadata are:

```
	double fates_leaf_maintresp_model ;
		fates_leaf_maintresp_model:units = "unitless" ;
		fates_leaf_maintresp_model:long_name = "switch for choosing between leaf maintenance respiration model. 1=Ryan (1991)" ;
	double fates_leaf_photosynth_tempsens_model ;
		fates_leaf_photosynth_tempsens_model:units = "unitless" ;
		fates_leaf_photosynth_tempsens_model:long_name = "switch for choosing the model that defines the temperature sensitivity of photosynthetic parameters (vcmax, jmax). 1=non-acclimating" ;

```

both would just be scalar (not PFT-indexed), following how we already have a global switch for the stomatal conductance.

rgknox:
@ckoven and I debated off-line about the new parameters. To summarize, we considered if there would indeed be different hypotheses deployed for these switches in some reasonably soon timeline to merit their implementation now. It seems reasonable because we do have work slated to update respiration and photo temperature acclimation.  I will move forward and add these.

rgknox:
@ckoven , can you confirm that the maintenance respiration hypotheses that we will add a switch for are specific to leaves?

Also, the acclimation timescale parameter has this form:  fates_photo_temp_acclim_timescale   If the switch more closely matches its name the two will appear next to each other in the file, which is nice for users.

ckoven:
@rgknox apologies for slow response. (1) I guess we should make the switch open-ended for maintenance respiration, particularly as per point 2 of https://github.com/NGEET/fates/issues/729#issuecomment-804385518.  So let's call the first variable `fates_maintresp_model`. (2) I agree -- we should group the parameters together in the file.  So let's call the second one `fates_photo_tempsens_model`.  Thanks!

Issue #763, opened 2021-07-08
FatesPFTIndexSwapper.py needs updating to api 16.1


rgknox:
FatesPFTIndexSwapper.py will not work on the new parameter file as per 16.1.  This is due to new dimension combinations. Thanks for bringing this up @JessicaNeedham 

I will patch, but feel free to use this in the mean time:

https://drive.google.com/file/d/132HEMeZPv0OJKPkFXiMxYByKR7cLc7jK/view?usp=sharing

adrifoster:
I am using this new FatesPFTIndexSwapper.py, which works, but I am getting some runtime warnings.

From FatesPFTIndexSwapper.py I get this warning seemingly when it is just about finished:

```
/home/afoster/env/lib/python3.7/site-packages/scipy/io/netcdf.py:312: RuntimeWarning: Cannot close a netcdf_file opened with mmap=True, when netcdf_variables or arrays referring to its data still exist. All data arrays obtained from such files refer directly to data on disk, and must be copied before the file can be cleanly closed. (See netcdf_file docstring for more information on mmap.)
  ), category=RuntimeWarning)
Cloneing complete!
```

From the modify_fates_paramfile.py, when I try to change the pft names I get this error:

```
/project/tss/afoster/ctsm_fates/src/fates/tools/modify_fates_paramfile.py:65: DeprecationWarning: string or file could not be read to its end due to unmatched data; this will raise a ValueError in the future.
  outputval = np.fromstring(args.val, sep=',', dtype=np.float32)
replacing prior value of pft name for PFT 1, which was "needleleaf_evergreen_extratrop_tree", with new value of "pice_mari"
replacing file: /project/tss/afoster/parameter_files/param_file_canada.nc
```

I'm not totally convinced this one isn't an error on my end, but it still seems to work. It also just seems like `np.fromstring()` is deprecated in its current use. Things still seem to go okay, so just wanted users to be aware of potential warnings.

PR #764, opened 2021-07-16
restart fix for efflux variables

rgknox:
### Description:

Efflux is defined at the cohort scale. Efflux is passed from the cohort to the output flux to the HLMs litter pool before we perform disturbance and mortality. Thus, in order for restarts to work, this flux at the site level must be preserved prior to mortality and disturbance.   Luckily, we can use the flux_diags structure which already exists to generate the output flux, this same flux will be preserved after a restart read. 

Fixes: #763

This should not change results, because we don't yet generate efflux in any tests.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/FatesSoilBGCFluxMod.F90
+++ b/biogeochem/FatesSoilBGCFluxMod.F90
@@ -1022,34 +1022,18 @@ contains
           
        end select
 
+       
+       ! If there is any efflux (from stores overflowing)
+       ! than pass that to the labile litter pool
+       
+       do id = 1,nlev_eff_decomp
+          flux_lab_si(id) = flux_lab_si(id) + &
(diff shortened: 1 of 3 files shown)


rgknox:
Including update to the pft index swapper tool.

rgknox:
tested against baseline (temporary modification to ctsm5.1.dev056 that removes mosart tests from 1x1_brazil) :

fates-sci.1.47.1_api.17.0.0-ctsm5.1.dev056-RsFatesFix
 
Test results:
/glade/scratch/rgknox/tests_1010-122445ch

Expected DIFF against the base in: ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCru.cheyenne_intel.clm-FatesPRT2

Issue #443, opened 2018-11-12
optimizing photosynthesis


rgknox:
@walkeranthonyp, Nathan Collier and I met up at the DOE Modeling PI meeting and had a fun discussion about improving the photosynthesis code.  @walkeranthonyp has some more fundamental ways to improve how we go about this process regarding which schemes we use and how the equations themselves are cast, and Nate was interested in how we actual perform out solve for C_i (interstitial CO2), which is an iterative loop converging on a solution to non-linear algebraic equations.

See here where we perform the iteration: https://github.com/NGEET/fates/blob/master/biogeophys/FatesPlantRespPhotosynthMod.F90#L992

An important note here, is that the photosynthesis scheme calculates stomatal conductance (among other things), which in-turn effects latent-heat transfer from the surface, which effects land-atmospheric stability, which should tightly feed-back on these fluxes again.  Thus, photosynthesis is called inside this broader iteration loop  (CLM/ELM side of code)  that seeks to converge on a stability parameter and land-fluxes with a low residual (change since last iteration).  Note: These coupled iteration loops make the search for C_i (and the quadratic root finders) the most often called subroutines in the model.

I wanted to diagnose how efficient our code is currently performing the solve for C_I, and see if there is any low-hanging fruit that can be picked to help speed things up and make it more efficient.  

The first thing I did was do some counting statistics to see the distributions of how many iterations are performed.  The following plot shows results from a 1 year simulation at BCI, using a forest inventory initialization. 

Top Left Panel: Histogram of counts needed to converge on the outer loop (stability parameter)
Top Right Panel: Histogram of the final residuals in the C_i loop, using the existing formulation. The red-line is the residual error criteria for success
Bottom Left Panel: Histogram of an ALTERNATIVE way to formulate the residual in the C_i loop, assuming that the residual is the change in C_i from previous time-point, normalized by the canopy CO2 partial pressure.
Bottom Right Panel: Histogram of the number of inner loop counts for C_i (which is capped at 5. Spoiler alert, almost every single iteration attempt used up 5).  

![quad-bci-v1](https://user-images.githubusercontent.com/5891980/48369174-a1c89880-e66a-11e8-8e74-54fed1a08a62.png)

[...]

PR #768, opened 2021-08-05
Nutrient controls on leaf level photosynthetic processes 

jenniferholm:
This code takes into account leaf-level nutrient effects on photosynthesis when PARTEH is used, particularly updating how the vcmax25 and jmax25 terms are calculated. These new relationships between leaf N, leaf P, and regression coefficients estimated from a photosynthesis data set will allow for a dynamic approach to represent nutrient constraints on photosynthesis.

Connected to the issue discussion at #597 , #443 , and could later be important for the acclimation work being done by others. 

### Description:
Following the work that was implemented in ELMv1 by Qing Zhu 2019 (https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2018MS001571), the code has been updated so that "photosynthesis is a function of leaf N (lnc: gN/m2 leaf area) and leaf P (lpc: gP/m2 leaf area) contents, which are commonly measured leaf properties, and their fundamental controls on plant photosynthesis rates are widely supported by observations (Kattge et al., 2009; Reich et al., 2009)." 

The equations for vcmax25top and jmax25top have now been updated to be a function of leaf N and leaf P, based on the results from (Walker et al., 2014), instead of the static values provided in the parameter file. 
These changes only occur when PARTEH is turned in (in c-only mode and CNP mode in PARTEH). The default, non-PARTEH mode, still uses the vcmax25top from the parameter file, and jmax25top being derived from the parameter file vcmax25top within "FatesParameterDerivedMod". 

Now that there is a varying vcmax25top term, additional changes need to be made where other terms are based on vcmax25top. For example, the "kn" term (the decay coefficient) is now based on the newly updated vcmax25top, which in turn produces a new "nscaler" term (leaf nitrogen scaling coefficient).  Two new variables were introduced to keep track of these new terms: kn_prt, and nscaler_prt (standing for PARTEH). For now a newly updated "kn_prt" term is only being implemented in the photosynthesis code when using PARTEH, not in other parts of the code where "kn" is used, such as in FatesAllometryMod when using kn to calculate things like LAI an SLA. The default, parameter value of vcmax25top is still being used. 

The "kp25top" term (initial slope of CO2 response curve) has also been updated to account for the varying vcmax25top, but only when using PARTEH. Otherwise kp25top is still being derived in "FatesParameterDerivedMod" using this%kp25top(ft,iage)   = 20000._r8 * vcmax25top(ft,iage).   
Note - I would appreciate any feedback on this change, and also make sure I'm not double charging kp25top * 20000 twice. In the LEAFN figure below the initial jump in leaf N in the "w/ photosyn. CO2 response" test is due to this sudden change in how kp25top is calculated, which wasn't used during the spinup procedure. 

### Collaborators:
@rgknox , @walkeranthonyp , @ckoven , @Qing Zhu, @rosiealice , @alistairrogers 

These code changes will definitely change answers. Below are some output examples from different scenarios testing the changes to different forest states and fluxes. 
All simulations began after running an AD spinup case (with supplemental N and P), a 200-year regular spinup (with AD turned off, and no supplemental N and P, until all carbon pools were stable in equilibrium). Then code changes to photosynthesis were turned on. 
Legend key for figures below:
"Regular Spinup" - continuing the regular spinup mode, with no leaf level photosynthesis changes (as a pseudo default case)
"w/ photosyn." = all leaf level photosynthesis code changes, including adding the multiplier of "day length factor", which allows for the consideration of light limitation. Testing at a boreal forest site, so large impacts. 
"w/ photosyn. no day-length" = removing the day length multiplying factor to test the impacts of only the leaf level nutrient additions. 
"w/ photosyn. CO2 response" = all leaf level photosynthesis code changes, including adding the multiplier of day length factor, and slope of CO2 response curve based on updated vcmax25top. 

"C-only PARTEH" = test of if case is (prt_carbon_allom_hyp), and using the new approach of vcmax25top_prt = lnc_top * flnr(ft) * fnr * act25. 

### Checklist:

Yes, the code changes requires these checklist items and they still need to be done. 

### Test Results:

E3SM = Ryan's PARTEH branch = "rgknox/lnd/fates-cnp      cdbeea0 Made modifications to supplementation of NP during FATES-ECA."

FATES baseline hash-tag:

Test Output:
Single site test, in Alaska boreal forest - 

image

image

image

image

image

image

### DIFF
--- a/biogeochem/FatesSoilBGCFluxMod.F90
+++ b/biogeochem/FatesSoilBGCFluxMod.F90
@@ -1182,7 +1182,7 @@ contains
     integer, parameter :: downreg_logi   = 2
     integer, parameter :: downreg_CN_logi = 3
     
-    integer, parameter :: downreg_type = downreg_linear
+    integer, parameter :: downreg_type = downreg_CN_logi
 
     
     real(r8), parameter :: logi_k   = 25.0_r8         ! logistic function k
--- a/biogeophys/FatesPlantRespPhotosynthMod.F90
+++ b/biogeophys/FatesPlantRespPhotosynthMod.F90
@@ -40,6 +40,7 @@ module FATESPlantRespPhotosynthMod
    use PRTGenericMod,     only : prt_cnp_flex_allom_hyp 
    use PRTGenericMod,     only : all_carbon_elements
    use PRTGenericMod,     only : nitrogen_element
+   use PRTGenericMod,     only : phosphorus_element
    use PRTGenericMod,     only : leaf_organ
    use PRTGenericMod,     only : fnrt_organ
    use PRTGenericMod,     only : sapw_organ
@@ -161,6 +162,10 @@ contains
                                    ! (umol electrons/m**2/s)
     real(r8) :: kp_z               ! leaf layer initial slope of CO2 response 
                                    ! curve (C4 plants)
+    real(r8) :: vcmax25top_prt     ! canopy top: variable maximum rate of carboxylation at 25C for parteh
+                                   ! (umol CO2/m**2/s)
+    real(r8) :: jmax25top_prt      ! canopy top: variable maximum electron transport rate at 25C for parteh
+                                   ! (umol electrons/m**2/s)
     real(r8) :: c13disc_z(nclmax,maxpft,nlevleaf) ! carbon 13 in newly assimilated carbon at leaf level
    
     real(r8) :: mm_kco2            ! Michaelis-Menten constant for CO2 (Pa)
@@ -169,10 +174,12 @@ contains
     real(r8) :: btran_eff          ! effective transpiration wetness factor (0 to 1)
     real(r8) :: stomatal_intercept_btran   ! water-stressed minimum stomatal conductance (umol H2O/m**2/s)
     real(r8) :: kn                 ! leaf nitrogen decay coefficient
+    real(r8) :: kn_prt             ! leaf nitrogen decay coefficient based on variable vcmax25top_prt
     real(r8) :: cf                 ! s m**2/umol -> s/m (ideal gas conversion) [umol/m3]
     real(r8) :: gb_mol             ! leaf boundary layer conductance (molar form: [umol /m**2/s])
     real(r8) :: ceair              ! vapor pressure of air, constrained (Pa)
     real(r8) :: nscaler            ! leaf nitrogen scaling coefficient
+    real(r8) :: nscaler_prt        ! leaf nitrogen scaling coefficient based on variable vcmax25top_prt
     real(r8) :: leaf_frac          ! ratio of to leaf biomass to total alive biomass
     real(r8) :: tcsoi              ! Temperature response function for root respiration. 
     real(r8) :: tcwood             ! Temperature response function for wood
@@ -188,6 +195,7 @@ contains
     real(r8) :: fnrt_n             ! Fine root nitrogen content (kgN/plant)
     real(r8) :: leaf_c             ! Leaf carbon (kgC/plant)
     real(r8) :: leaf_n             ! leaf nitrogen content (kgN/plant)
+    real(r8) :: leaf_p             ! leaf phosphorus (kgP/plant)
     real(r8) :: g_sb_leaves        ! Mean combined (stomata+boundary layer) leaf conductance [m/s]
                                    ! over all of the patch's leaves.  The "sb" refers to the combined
                                    ! "s"tomatal and "b"oundary layer.
@@ -206,6 +214,10 @@ contains
                                    ! over each cohort x layer.
     real(r8) :: cohort_eleaf_area  ! This is the effective leaf area [m2] reported by each cohort
     real(r8) :: lnc_top            ! Leaf nitrogen content per unit area at canopy top [gN/m2]
+    real(r8) :: lpc_top            ! Leaf phosphorus content per unit area at canopy top [gP/m2]
+    real(r8) :: jmax_np1           ! jmax~np relationship coefficient
+    real(r8) :: jmax_np2           ! jmax~np relationship coefficient
+    real(r8) :: jmax_np3           ! jmax~np relationship coefficient
     real(r8) :: lmr25top           ! canopy top leaf maint resp rate at 25C 
                                    ! for this plant or pft (umol CO2/m**2/s)
     real(r8) :: leaf_inc           ! LAI-only portion of the vegetation increment of dinc_ed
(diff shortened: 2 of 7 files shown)


biogeochem/FatesSoilBGCFluxMod.F90
```diff
@@ -1182,7 +1182,7 @@ function ECACScalar(ccohort, element_id) result(c_scalar)
     integer, parameter :: downreg_logi   = 2
     integer, parameter :: downreg_CN_logi = 3
     
-    integer, parameter :: downreg_type = downreg_linear
+    integer, parameter :: downreg_type = downreg_CN_logi
```
rgknox, line 1185: @jenniferholm have you been getting better results with the logistic feedback mechanism? 
If so, lets submit this change in an other PR to maintain bit-for-bit answers with baseline tests

jenniferholm: @rgknox - I didn't find the differences between linear and CN_logistic to be that large, or at least not make the results any "better".  I think this was left here from my last test, and I can revert it back to linear to maintain bit-for-bit.  

The same goes for the N and P efflux code changes that I've been testing with. Since we've been supplying such a large amount of nutrients during testing, I turned on the n_efflux and p_efflux, so that it does not go back into storage.  I'm not sure what the default should be for this, and I can defer to you (or Charlie or others).  My only two-cents is that maybe it does seem like a good idea to have defaulted "turned on" so that we won't get this really high accumulation of nutrients into storage by accident, when testing.....

rgknox: Regarding your second point.  If the stores of N and P are at or above the maximum holding capacity, the need function that drives uptake on the following model steps should drop to 0 in all cases.  This includes RD, ECA and prescribed.  So stores should not start growing without control (should being the key word, hehe).  I think the only time I've seen stores get way above the target maximum is when I've been using poorly calibrated ECA parameters, which will overshoot uptake from soils with high availability by orders of magnitude. But in that situation, we are so off the mark in terms of what we would expect from nature, I don't think it matters if we hold the excess in storage or dump it to a labile pool.

jenniferholm: @rgknox - I went ahead and reverted these back to the default values and settings to be easier, and b4b.  So back to "downreg_linear" and also the n_efflux and p_efflux back to storage.  That totally makes sense that the plant need function will drive the uptake, and thus stop any excess uptake of the stores of N and P are at it's holding capacity, so limit any really high uptake and excess.

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -434,48 +464,70 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
...
                               end select
 
                               lmr25top = 2.525e-6_r8 * (1.5_r8 ** ((25._r8 - 20._r8)/10._r8))
                               lmr25top = lmr25top * lnc_top / (umolC_to_kgC * g_per_kg)
                               
+                              ! Bonan et al (2011) JGR, 116, doi:10.1029/2010JG001593 used
+                              ! kn = 0.11. Here, derive kn from vcmax25 as in Lloyd et al 
+                              ! (2010) Biogeosciences, 7, 1833-1859
+
+                              kn = decay_coeff_kn(ft,currentCohort%vcmax25top)
```
rgknox, line 519: or, alternative to my above comment, should we remove these and pass nscaler_prt into LeafLayerMaintenanceRespiration() ?

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -434,48 +464,70 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
...
                                  ! use the stoichiometry parameter
                                  ! lnc_top  = prt_params%nitr_stoich_p1(ft,prt_params%organ_param_id(leaf_organ))/slatop(ft)
                                  
+                                 vcmax25top_prt = exp(vcmax_np1(ft) + vcmax_np2(ft)*log(lnc_top) + &
+                                          vcmax_np3(ft)*log(lpc_top) + vcmax_np4(ft)*log(lnc_top)*log(lpc_top))
+                                 jmax25top_prt = exp(jmax_np1 + jmax_np2*log(vcmax25top_prt) + jmax_np3*log(lpc_top))
+                                 vcmax25top_prt = min(max(vcmax25top_prt, 10.0_r8), 150.0_r8)
+                                 jmax25top_prt  = min(max(jmax25top_prt, 10.0_r8), 250.0_r8)
+
+                                 kn_prt = decay_coeff_kn(ft,vcmax25top_prt)
```
rgknox, line 507: are the kn_prt and nscaler_prt variables used @jenniferholm ?  Should we remove these?

rgknox:
@jenniferholm I noticed that there the leaf maintenance respiration calculations use the old or existing nscalar value, which is based on the stochiometry parameter constant:

https://github.com/NGEET/fates/pull/768/files#diff-f62a39ba3dc682263ce52ac5f1bec68ac12bc849e5987c1a793938bb2a1339e8R534

But the effective biophysical rates, vcmax, jmax etc, take arguments for both the old (nscalar) and new (nscalar_prt) values, which gets confusing. 


I'm wondering if it would be worth it to determine the lnc, kn and nscalar values before both subroutines (which you do), but only arrive at a singluar lnc, nscalar and kn value (not both nscalar and nscalar_prt, and kn/kn_prt, etc), and pass in those just the single value to the routines, consistently.  Likewise, I think we need to add a control flag from the parameter file that allows us to override the prognostic N and P concentrations in favor of the values derived from the stoichiometry parameter constants.

Sorry if I just don't get it, maybe I just dont understand why a different nscalar value would be used for maintenance respiration and the vcmax.

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -251,8 +265,24 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
...
-         stomatal_intercept   => EDPftvarcon_inst%stomatal_intercept ) !Unstressed minimum stomatal conductance
-
+         stomatal_intercept   => EDPftvarcon_inst%stomatal_intercept, &  !Unstressed minimum stomatal conductance
+         flnr        => EDPftvarcon_inst%flnr, &       ! Input: fraction of leaf N in the Rubisco enzyme (gN Rubisco / gN leaf)
+         vcmax_np1   => EDPftvarcon_inst%vcmax_np1, &  !vcmax~np relationship coefficient
+         vcmax_np2   => EDPftvarcon_inst%vcmax_np2, &  !vcmax~np relationship coefficient
+         vcmax_np3   => EDPftvarcon_inst%vcmax_np3, &  !vcmax~np relationship coefficient
+         vcmax_np4   => EDPftvarcon_inst%vcmax_np4 )   !vcmax~np relationship coefficient
+
+         ! vcmax25 parameters from Bonan et al (2011) JGR, 116, doi:10.1029/2010JG001593
```
qzhu-lbl, line 275: It will be cleaner if fnr, act25, jmax_np parameters are all moved to parameter file, just like vcmax_np parameters. It will also benefit parameter tuning later.

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -434,48 +464,70 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
...
                                  leaf_c  = currentCohort%prt%GetState(leaf_organ, all_carbon_elements)
                                  if( (leaf_c*slatop(ft)) > nearzero) then
                                     leaf_n  = currentCohort%prt%GetState(leaf_organ, nitrogen_element)
+                                    leaf_p  = currentCohort%prt%GetState(leaf_organ, phosphorus_element)
                                     lnc_top = leaf_n / (slatop(ft) * leaf_c )
+                                    lpc_top = leaf_p / (slatop(ft) * leaf_c )
+                                    ! lnc_top = leaf_n / currentPatch%tlai_profile(cl,ft,iv) !testing with LAI method
+                                    ! lpc_top = leaf_p / currentPatch%tlai_profile(cl,ft,iv) !testing with LAI method
+                                    lnc_top = min(max(lnc_top,0.25_r8),3.0_r8) !based on doi: 10.1002/ece3.1173
+                                    lpc_top = min(max(lpc_top,0.014_r8),0.85_r8) !based on doi: 10.1002/ece3.1173
```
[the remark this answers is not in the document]
walkeranthonyp: Broadly speaking I would say the dataset in http://dx.doi.org/10.1002/ece3.1173 is not robust, it's a relatively small dataset n = ~100. And a general issue is that the Vcmax values are low-biased, likely due to a low-bias in the dataset, see https://www.nature.com/articles/s41467-021-25163-9 for details.   

I'm a co-author on a similar analysis but with a much larger dataset n >1000, and I'm allowed to share the relationships with the team but they're all mass-based. I need to get hold of the dataset to calculate area-based relationships.

Issue #644, opened 2020-05-07
Refactor yearly net uptake


glemieux:
This issue came up in the context of #623 review.  The `trim_canopy` routine uses this variable to check to see if there has been activity in a leaf layer for a given year:
https://github.com/NGEET/fates/blob/50e3c74f0afdeb2bcbbf19697f69e28440b4799d/biogeochem/EDPhysiologyMod.F90#L456
As noted by @rgknox this type of equivalence check against real numbers should be avoided.  This type of check is also present in `DeallocateCohort`:
https://github.com/NGEET/fates/blob/50e3c74f0afdeb2bcbbf19697f69e28440b4799d/biogeochem/EDCohortDynamicsMod.F90#L1349-L1357

Given that this is slightly beyond the scope of #623, I'm breaking this out into a separate issue to be address later.

PR #769, opened 2021-08-09
shorten cohort memory arrays to nlevleafmem

rgknox:
### Description:

There are two arrays used to track the carbon budget of leaf layers in the plant:  ts_net_uptake and year_net_uptake.  A few notes about these arrays:
1) These two arrays are attached to the cohort structure
2) These two arrays were as large as nlevleaf, which is the maximum possible number of leaf layers, which due to the nature of our leaf layering scheme, had to be large as the system is unbounded (ie 30)
3) To track and fill these arrays, we were performing loops of ~30 iterations inside various cohort loops (expensive)

This PR changes the way we track leaf layer carbon balance, by remembering only the 4 lowest layers in the plant's crown, instead of all 30.  The memory of these layers is used solely for trimming, and trimming only happens at the bottom, so we need not remember all of the layers, just the bottom.  We track 4 layers, because that is 1 larger than the number of layer memory needed to perform a regression on the trimming calculation. 

Various changes to the code are in response to the change in this layering, such as how we fuse cohorts with memory in different layers.  Also, there is a difference between the maximum number of layers a plant could have when on allometry, versus the number of layers it currently has. The latter may be lower than the maximum if a plant is not replacing leaf tissues lost to turnover.

Fixes: #272
Addresses #644 

### Collaborators:

@ckoven @glemieux @rosiealice 

Answers should change, but its unlikely any of our tests will catch the change because trimming only occurs annually.

### Checklist:

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:


rgknox:
To Do:
1) Evaluate long term simulations to see how different results are with base (b4b not expected)
2) Benchmark memory and timing changes

rgknox:
This branch is now producing nigh indistinguishable results compared to base at a single test site.

leafmem-v14-base-p1

Now running long term gridded smoke tests on cheyenne.

PR #801, opened 2021-10-30
Multiple updates to drought deciduous phenology

mpaiao:
Added new options and controls for drought deciduous phenology

### Description:

First set of changes in the drought deciduous phenology to improve its predictability.
1. Implemented option to drive leaves on / leaves off using soil matric potential instead of soil water content, which is more intuitive for setting up thresholds.  The choice is controlled by fates_phen_drought_threshold: when it is positive, it continues to use the soil volumetric water content (m3/m3), and when it is negative, it uses soil matric potential (in mm).
2. When drought-deciduous leaves are off, the plant now stops allocating carbon to fine roots to reduce carbon losses when carbon balance is necessarily non-positive.
3. Revised variable sapw_memory and struct_memory. The name referred to the target value, but the variable was storing the losses, and thus inconsistent with leafmemory (formerly laimemory, see #865). All of these, along with the new fnrtmemory now store the carbon pool prior to leaf shedding.
4. The drought-deciduous status is still done at the site level but now it is PFT-dependent because the root distribution is PFT dependent. This required turning a few site-level variables into site x PFT.
5.  Sub-routine phenology_leafonoff (EDPhysiologyMod.F90) was rewritten to reduce duplicated code. Most of the steps for cold deciduous and drought deciduous were the same.  The revised routine first checks whether or not it is time o flush or shed leaves, then it uses a common set of commands.

### Collaborators:

Discussions with @rgknox @ckoven @rosiealice @jkshuman @XiulinGao

These changes should affect the code when running with drought deciduous phenology. The code now uses a different approach to find moisture threshold (rooting zone and optionally, matric potential), and the carbon allocation strategies changed during the leaf-off periods.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


biogeochem/EDPhysiologyMod.F90
```diff
@@ -713,46 +716,41 @@ subroutine phenology( currentSite, bc_in )
...
     ! continues monotonically, indefinitely
     model_day_int = nint(hlm_model_day)
 
 
-    ! Use the following layer index to calculate drought conditions
-    ilayer_swater = minloc(abs(bc_in%z_sisl(:)-dphen_soil_depth),dim=1)
-
-
-    ! Parameter of drought decid leaf loss in mm in top layer...FIX(RF,032414)
+    ! Parameter of drought decid leaf loss in mm in top layer...FIX(RF,032414)
```
rosiealice, line 753: I believe that this might indeed be the fix to this, a mere 8 years later...

rosiealice, line 753:
```suggestion
    ! Parameter of drought decid leaf loss in mm in top layer...FIX(RF,032414) 
```

biogeochem/EDPhysiologyMod.F90
```diff
@@ -948,117 +948,164 @@ subroutine phenology( currentSite, bc_in )
...
-    enddo
-    currentSite%water_memory(1) = bc_in%h2o_liqvol_sl(ilayer_swater)
+    ! Why don't the drought deciduous trees grow in the North? 
+    ! Is cold decidousness maybe even the same as drought deciduosness there (and so does this 
+    ! distinction actually matter??).... 
+    !     MLO. They are probably not the same: unlike cold deciduous temperatures are high when
+    !     drought deciduous leaves are off, which means that their maintenance respiration is high
+    !     during the leaf-off season.  For them to be viable, we may need different allocation
+    !     strategies so they don't exhaust their storage to maintain fine roots and sapwood.
+
```
rosiealice, line 958:
```suggestion
    ! Why don't the drought deciduous trees grow in the North? 
    ! Is cold decidousness maybe even the same as drought deciduosness there (and so does this 
    ! distinction actually matter??).... 
    !     MLO. They are probably not the same: unlike cold deciduous temperatures are high when
    !     drought deciduous leaves are off, which means that their maintenance respiration is high
    !     during the leaf-off season.  For them to be viable, we may need different allocation
    !     strategies so they don't exhaust their storage to maintain fine roots and sapwood.

```

rosiealice, line 958: Can probably delete this musing...

biogeochem/EDPhysiologyMod.F90
```diff
@@ -948,117 +948,164 @@ subroutine phenology( currentSite, bc_in )
...
+       ! Calculate the mean soil moisture ( liquid volume (m3/m3) and matric potential (mm))
+       !    over the last 10 days
+       mean_10day_liqvol = sum(currentSite%liqvol_memory(1:numWaterMem,ft))/real(numWaterMem,r8)
+       mean_10day_smp    = sum(currentSite%smp_memory   (1:numWaterMem,ft))/real(numWaterMem,r8)
+
+       ! Compare the moisture with the threshold.
+       if ( ED_val_phen_drought_threshold >= 0. ) then
+          ! Liquid volume in reference layer (m3/m3)
+          smoist_below_threshold = mean_10day_liqvol < ED_val_phen_drought_threshold
```
rosiealice, line 1001: how does Fortran parse the < as different from a - ?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -948,117 +948,164 @@ subroutine phenology( currentSite, bc_in )
...
+             currentSite%dndaysleafon(ft) = 0
+          end if
        end if
-    end if
 
-    ! LEAF OFF: DROUGHT DECIDUOUS LIFESPAN - if the leaf gets to
-    ! the end of its useful life. A*, E*
-    ! i.e. Are the leaves rouhgly at the end of their lives?
+       ! LEAF OFF: DROUGHT DECIDUOUS LIFESPAN - if the leaf gets to
+       ! the end of its useful life. A*, E*
```
rosiealice, line 1084:
```suggestion
       ! the end of its useful life. 
```

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1106,229 +1165,159 @@ subroutine phenology_leafonoff(currentSite)
...
+             ! B. Is this the time for COLD LEAVES to switch to OFF?
+             is_shedding_time = ( currentSite%cstatus == phen_cstat_nevercold  .or.  & ! Past leaf drop day or...
+                                  currentSite%cstatus == phen_cstat_iscold   ) .and. & ! Too cold?
+                                currentCohort%status_coh == leaves_on          .and. & ! Leaves have not dropped yet
+                                ( currentCohort%dbh > EDPftvarcon_inst%phen_cold_size_threshold(ipft) .or. & ! Grasses are big enough or...
+                                  prt_params%woody(ipft) == itrue                                     )      ! this is a woody PFT.
+
+          elseif (prt_params%stress_decid(ipft) == itrue ) then ! Drought deciduous
+             ! A. Is this the time for DROUGHT LEAVES to switch to ON?
+             is_flushing_time = ( currentSite%dstatus(ipft) == phen_dstat_moiston  .or.  & ! Conditions are sufficiently moist
```
rosiealice, line 1221: While this is being overhauled, is there an argument for making leaves come on over some period of time rather than just on one day? Not that it's 100% necessary, just thought I'd mention it...

biogeochem/EDPhysiologyMod.F90
mpaiao: agree

biogeochem/EDPhysiologyMod.F90
mpaiao: I set this to be a logical variable, so it's `smoist_below_threshold = .true.` when the 10-day average is less the threshold.

biogeochem/EDPhysiologyMod.F90
mpaiao: I think this will be one of the options I am going to try (along with other approaches to decide abscission and flushing times).  One way of doing it would be similar to ED-2.2, where we have two moisture thresholds. The maximum leaf biomass goes from zero at the lower threshold to one at the upper threshold.

glemieux:
@mpaiao are you going to open up a new version of this?

mpaiao:
@glemieux Ops, I didn't notice that my branch clean up deleted the PR. But yes, I re-implemented my drought deciduous changes on a separate branch, in a way that it _should_ be easier to merge into the mainline.

I think this new branch still needs some testing, and I will probably make some additional edits (a few global parameters should become PFT-specific), but let me create a new PR using that branch instead, so the transition is documented.

Issue #781, opened 2021-09-27
Non-zero frost mortality at the beginning of the simulation


mpaiao:
I noticed that all my simulations have non-zero frost mortality at the first time step. All of them are for lowland sites near the Equator, where the temperature never gets near freezing. Below is an example for a simulation at Paracou, but it happens in other places and settings too.

[m8-tspft-D0006_ParacouEGHD_CLM_FATES.pdf](https://github.com/NGEET/fates/files/7238740/m8-tspft-D0006_ParacouEGHD_CLM_FATES.pdf)

Potentially a some variable that is not correctly initialised?

glemieux:
@mpaiao, now that the running means PR #724 has been integrated with [sci.1.52.0_api.20.0.0](https://github.com/NGEET/fates/releases/tag/sci.1.52.0_api.20.0.0), can you check to see if this issue presents itself (without using the fix in #816)?  Note that this update requires [ctsm5.1.dev066](https://github.com/ESCOMP/CTSM/releases/tag/ctsm5.1.dev066).

Per @rgknox, the relevant fixes is here in the same place as your frost mortality fix:
https://github.com/NGEET/fates/blob/b27149270b459fc2518dc2a5bacfc48ccd256528/biogeochem/EDMortalityFunctionsMod.F90#L174

The initialization is here;
https://github.com/NGEET/fates/blob/b27149270b459fc2518dc2a5bacfc48ccd256528/biogeochem/EDPatchDynamicsMod.F90#L2043

Note that it uses `temp_init_veg`, which is updated here:
https://github.com/NGEET/fates/blob/b27149270b459fc2518dc2a5bacfc48ccd256528/biogeochem/EDPatchDynamicsMod.F90#L2024

mpaiao:
@glemieux @rgknox I think initializing the temperature with 15°C should fix the issue. I guess a quick test is to see if any frost mortality occurs on the first day of simulation for a tropical place with [sci.1.52.0_api.20.0.0](https://github.com/NGEET/fates/releases/tag/sci.1.52.0_api.20.0.0). If frost mortality is zero, then #816 is not needed.

PR #804, opened 2021-11-04
Frost mortality fix

mpaiao:
Bug fix for non-zero mortality at the beginning of every FATES simulation.

### Description:

Simple bug fix for the non-zero frost mortality during the first day of the simulation. The problem is that patch variable `bc_in%t_veg24_pa` is set to zero at the beginning of the simulation, and this is interpreted as zero Kelvin. My solution was to simply bypass frost mortality during the first simulation day. This fix addresses issue #781 and should have minimal impact on long-term simulations.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


rgknox:
@mpaiao , this fix might help elucidate differences I'm seeing when migrating to tracking running means on the fates side, ie #724

mpaiao:
@rgknox I think the initial assignment for many of these "memory" variables may be worth thinking more. I see that relative humidity also starts at 0, which could exaggerate dryness for SPITFIRE too (though short-lived, just the first day).

mpaiao:
By the way, I noticed that my 3 pull requests may need to be implemented sequentially, because I am seeing many more changes in this PR than the actual fix for mortality. I will try to organize my branches a bit better next time.

jkshuman:
@mpaiao there is sometimes a high amount of fire in the first year. It would be great if this cleared that up. Thanks for finding and fixing this, but yes we should rethink initialization of these "memory" variables. Perhaps open a separate issue with a few examples, like the fire dependance on RH?

glemieux:
Closing per #816

PR #805, opened 2021-11-05
Some logic fixes to nocomp and nocomp + fixed_biogeog

ckoven:
This fixes things so that nocomp + fixed_biogeog works properly (it was crashing before during init) and also a fix for nocomp (it was allowing recruits on supposedly-disallowed patches).

I think this is why nocomp was super-slow before, because it was actually allowing and PFT to recruit on any patch.  Now it doesn't.

In my head, I had previously thought that nocomp also implied fixed_biogeog.  I realize now that it doesn't necessarily, and that the logic differs slightly between nocomp versus nocomp + fixed_biogeog.  Anyway, all the various modes should work as intended now:

- fixed_biogeog: only lets a given PFT exist on a gridcell if the map says it can
- fixed_biogeog + nocomp: only lets a given PFT exist on a gridcell if the map says it can, and separates the gridcell into patch areas that each PFT can solely exist on based on the map areas
- nocomp: splits up the map to give every possible PFT the same amount of space on every gridcell in which that PFT can solely exist on.

I think, in general, fixed_biogeog + nocomp is what people will want to use, not nocomp on its own, so we should probably add a test for a fixed_biogeog + nocomp configuration.

This should be answer-changing in nocomp, and allow the model to pass a nocomp + fixed_biogeog test (it wouldn't have before).  All other modes should be identical.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -16,6 +16,7 @@ module EDPhysiologyMod
   use FatesInterfaceTypesMod, only    : hlm_use_planthydro
   use FatesInterfaceTypesMod, only    : hlm_parteh_mode
   use FatesInterfaceTypesMod, only    : hlm_use_fixed_biogeog
+  use FatesInterfaceTypesMod, only    : hlm_use_nocomp
   use FatesInterfaceTypesMod, only    : hlm_nitrogen_spec
   use FatesInterfaceTypesMod, only    : hlm_phosphorus_spec
   use FatesConstantsMod, only    : r8 => fates_r8
@@ -1853,8 +1854,10 @@ contains
 
 
     do ft = 1,numpft
-       if(currentSite%use_this_pft(ft).eq.itrue)then
-       temp_cohort%canopy_trim = init_recruit_trim
+       if(currentSite%use_this_pft(ft).eq.itrue &
+            .and. ((hlm_use_nocomp .eq. ifalse) .or. (ft .eq. currentPatch%nocomp_pft_label)))then
+
+          temp_cohort%canopy_trim = init_recruit_trim
           temp_cohort%pft         = ft
           temp_cohort%hite        = EDPftvarcon_inst%hgt_min(ft)
           temp_cohort%coage       = 0.0_r8
--- a/main/EDInitMod.F90
+++ b/main/EDInitMod.F90
@@ -487,7 +487,7 @@ contains
           start_patch = 1   ! start at the first vegetated patch
           if(hlm_use_nocomp.eq.itrue)then
              num_new_patches = numpft
-             if(hlm_use_sp.eq.itrue)then
+             if( (hlm_use_sp.eq.itrue) .or. (hlm_use_fixed_biogeog .eq.itrue) )then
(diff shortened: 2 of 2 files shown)


main/EDInitMod.F90
```diff
@@ -487,7 +487,7 @@ subroutine init_patches( nsites, sites, bc_in)
           start_patch = 1   ! start at the first vegetated patch
           if(hlm_use_nocomp.eq.itrue)then
              num_new_patches = numpft
-             if(hlm_use_sp.eq.itrue)then
+             if( (hlm_use_sp.eq.itrue) .or. (hlm_use_fixed_biogeog .eq.itrue) )then
```
glemieux, line 490: We should be able to simplify this to just check fixed biogeog here since sp mode requires fixed biogeog:
https://github.com/NGEET/fates/blob/e1cf819ec3158b3a02cd693c73680dbca3cb4924/main/FatesInterfaceMod.F90#L1560-L1563

glemieux (approved):
I suggested one minor refactor for simplicity, but this looks good as is.

main/EDInitMod.F90
ckoven: Right.  I suspected that but wasn't completely sure that the code enforced it, so thanks for checking.  I'll update the code accordingly.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1853,8 +1854,10 @@ subroutine recruitment( currentSite, currentPatch, bc_in )
 
 
     do ft = 1,numpft
-       if(currentSite%use_this_pft(ft).eq.itrue)then
-       temp_cohort%canopy_trim = init_recruit_trim
+       if(currentSite%use_this_pft(ft).eq.itrue &
+            .and. ((hlm_use_nocomp .eq. ifalse) .or. (ft .eq. currentPatch%nocomp_pft_label)))then
+
+          temp_cohort%canopy_trim = init_recruit_trim
```
rosiealice, line 1860: I guess the crux of this is that the use_this_pft flag is a site level and not patch level quantity.  I wonder 
a) whether use_this_pft should be a patch level quantity. 
b) if not, could we put a comment where it is calculated (or here) to explain why this patch-level logic is required in nocomp mode?

glemieux:
All expected tests pass b4b, with the exception of an expected DIFF for the one no comp testmod.  Test results are here:
`/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr805-nocompfixes`

glemieux:
@ckoven am I remembering correctly that you had suggested we add a regression test covering the use of both nocomp + fixed biogeography modes?  If so, I'll add an issue on the ctsm-side and roll that into one of the ctsm-side cleanup PRs I have going.

ckoven:
@glemieux yes please do add a nocomp+fixed_biogeog test into one of the CTSM PRs

biogeochem/EDPhysiologyMod.F90
ckoven: @rosiealice I added the comment you requested above.

PR #817, opened 2021-12-06
Facilitate crops in SP mode. 

### LINKED ISSUE #782: SP mode: model crash with only 1 PFT
I was running FATES SP mode with single PFT. By using updated python scripts under tools folder to modify my paramter file, the fates_hlm_pft_map looks like this: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0](reversed in the dimension, so would be 14x1).
This then causes a problem in reading the fates_hlm_pft_map and leads to following error:

The distribution of this host land model PFT :           1
 into FATES PFTs, does not add up to 1.0.
 Error is:  -1.0000000000000000
 and the hlm_pft_map is:   0.0000000000000000
 Aborting
 ENDRUN:ERROR in /home/xiulingao/var-test-runs/CTSM/src/fates/main/EDPftvarcon.F90 at line 1762                                                                                                    $
 ERROR: Unknown error submitted to shr_abort_abort.

Apprarently, hlm_pft_map is not readed during the run. And then I tried to run SP mode with all 12 PFTs it then did work. So I think there's somehting wrong in either FatesInterfaceMod.F90 or EDPftvarcon.F90 which leads to failure in reading fates_hlm_pft_map when there is only one PFT or the output pft mapping matrix is wrong (so problem with FatesPFTIndexSwapper.py?)

rosiealice:
### Description:
To allow FATES-SP mode to generate crop areas, this PR updates the default PFT file to extend the map to HLM_PFTs 15 and 16. This means that for grid cells with crops, area is allocated.  In SP mode, the HLM actually shifts the crops onto the natural vegetation land unit, hence the previous solution (running over crop land units as well as soil land units) does not in itself solve the problem. 

Further, this PR also changes the order of operations logic related to the precision checking of the total patch area. Previously, this happened AFTER the call to init_cohorts, which appeared to result in cohorts that were larger than the patch area with an error the same size and that later trimmed off by the precision checking in EDInit (see changes).  This PR makes the call to init_cohorts occur after the change in patch areas to account for precision-level errors in the inputs.

Adding the crop PFTs appeared to trigger this error. It is possible that this change will fix  #782 which threw a similar error. 

This PR addresses issue #760

 

yes, this will change answers in SP and NOCOMP modes. 

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/main/EDInitMod.F90
+++ b/main/EDInitMod.F90
@@ -279,6 +279,7 @@ contains
     integer  :: dleafon    ! DOY for drought-decid leaf-on, initial guess
     integer  :: ft         ! PFT loop
     real(r8) :: sumarea    ! area of PFTs in nocomp mode.
+    integer  :: nhlmpft    ! number of hlm pfts. 
     integer  :: hlm_pft    ! used in fixed biogeog mode
     integer  :: fates_pft  ! used in fixed biogeog mode
     !----------------------------------------------------------------------
@@ -300,6 +301,7 @@ contains
        dleafoff = 300
        dleafon  = 100
        watermem = 0.5_r8
+       nhlmpft = size( EDPftvarcon_inst%hlm_pft_map,2)
 
        do s = 1,nsites
           sites(s)%nchilldays    = 0
@@ -331,7 +333,7 @@ contains
              ! hlm_pft_map is the area of that land in each FATES PFT (from param file)
 
              sites(s)%area_pft(1:numpft) = 0._r8
-             do hlm_pft = 1,size( EDPftvarcon_inst%hlm_pft_map,2)
+             do hlm_pft = 1,nhlmpft
                 do fates_pft = 1,numpft ! loop round all fates pfts for all hlm pfts
                    sites(s)%area_pft(fates_pft) = sites(s)%area_pft(fates_pft) + &
                         EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft) * bc_in(s)%pft_areafrac(hlm_pft)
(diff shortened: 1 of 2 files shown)


rosiealice:
TLAI comparison with CLM5 BEFORE change

SP_2rad_TLAI_41 (4)

TLAI comparison with CLM5 AFTER change. 
SP_2rad_TLAI_43 (4)

rosiealice:
As you can see, there is still something up with the high arctic and the sahara, but that's likely nothing to do with crops...

billsacks:
Thanks for your work on this @rosiealice ! I haven't gotten my head back into all of the discussions in #760 but can you clarify how much of #760 is done here?

Two specific questions I have are:

(1) Regarding this:

> In SP mode, the HLM actually shifts the crops onto the natural vegetation land unit, hence the previous solution (running over crop land units as well as soil land units) does not in itself solve the problem.

My understanding is that, at least eventually, we want crops still handled on their own landunits when running with FATES. Is that right? My understanding is also that, for non-FATES SP, CTSM puts crops on their own landunit (though I can never remember this for sure).

(2) If this PR is supposed to fully address #760, can you please make some comments about how much testing and/or code reading you've done to verify the various things in #760 that were raised along the lines of "we might need to do X / we should look into whether X is needed".

glemieux:
@rosiealice per our conversion at the fates software meeting a couple mondays ago, I tested this branch out using a single pft with an f45 grid run.  I'm still seeing the same failure mode from #782, so I think we need a separate fix for this issue.  Results can be found on cheyenne here:
`/glade/u/home/glemieux/scratch/ctsm-cases/spmode-pr817-newpfile-1pft.fates-sci.1.53.0_api.21.0.0-ctsm5.1.dev071-Cbcb9735e8-F3df5fa19.intel`

glemieux:
Closing per discussion with @rosiealice at the fates SE meeting today.

PR #821, opened 2021-12-15
Exact restart fix for satellite phenology mode

glemieux:
### Description:
This PR addresses issue https://github.com/ESCOMP/CTSM/issues/1485 in which fates satellite phenology (SP) mode does not pass an exact start (ERS) regression test.  

The issue has two subset issues that are fixed here:

1. It was determined that the ERS `COMPARE_base_rest` failure is due to the difference in the `gpp_tstep` during restart.  This difference is due both `gpp_tstep` needing to be carried over during restart and to a subset of cohort parameters not being updated to match the pre-restart state in `UpdateCohortBioPhysRates`.  This routine uses leaf biomass as a check to determine how to set `vcmax25top`, `jmax25top`, `tpu25top`, and `kp25top`, which eventually are used as inputs to update `gpp_tstep`.  This routine is run whenever cohorts are created or fused as well as during the restart process, although in SP mode this routine is only called during initial cohort creation and during restart.  During SP mode initialization all cohorts start with some leaf biomass and thus the above cohort parameters are defined as non-zero. That said its possible that the cohort is then set to a zero leaf biomass due to a satellite phenology update which is then written for the restart.  Given that there would be zero leaf biomass, the cohort parameters would be set to zero, thus causing a mismatch in the downstream `gpp_tstep` update.  
2. The ERS was failing to run after restart due to updates introduced in #735.  A simple SP mode check has been introduced to avoid this update.

### Collaborators:
@ckoven @rgknox 

PR should be b4b

### Checklist:

### Test Results:

CTSM (or) E3SM (specify which) test hash-tag: `ctsm5.1.dev066`

CTSM (or) E3SM (specify which) baseline hash-tag: `ctsm5.1.dev066`

FATES baseline hash-tag: `sci.1.52.0_api.20.0.0`

Test Output:
`/glade/u/home/glemieux/scratch/ctsm-tests/tests_issue_1485-dev066-sci1520-fatessuite1`

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -1935,12 +1935,9 @@ contains
        total_patch_area = 0._r8
        total_canopy_area = 0._r8
        bc_out(s)%canopy_fraction_pa(:) = 0._r8
-       
-       if (hlm_use_sp.eq.ifalse) then
-          bc_out(s)%dleaf_pa(:) = 0._r8
-          bc_out(s)%z0m_pa(:) = 0._r8
-          bc_out(s)%displa_pa(:) = 0._r8
-       endif
+       bc_out(s)%dleaf_pa(:) = 0._r8
+       bc_out(s)%z0m_pa(:) = 0._r8
+       bc_out(s)%displa_pa(:) = 0._r8
        
        currentPatch => sites(s)%oldest_patch
        c = fcolumn(s)
@@ -1970,7 +1967,7 @@ contains
              ! Avoid this if running in satellite phenology mode
              ! ----------------------------------------------------------------------------
 
-             if (currentPatch%total_canopy_area > nearzero .and. hlm_use_sp.eq.ifalse) then
+             if (currentPatch%total_canopy_area > nearzero) then
                 currentCohort => currentPatch%shortest
                 do while(associated(currentCohort))
                    if (currentCohort%canopy_layer .eq. 1) then
@@ -1988,7 +1985,8 @@ contains
                 currentCohort => currentPatch%shortest
                 do while(associated(currentCohort))
 
-                   ! mkae sure that allometries are correct
+                   if (hlm_use_sp.eq.ifalse) then
+                   ! make sure that allometries are correct
                    call carea_allom(currentCohort%dbh,currentCohort%n,sites(s)%spread,&
                         currentCohort%pft,currentCohort%c_area)
 
@@ -2000,6 +1998,7 @@ contains
                         currentCohort%c_area, currentCohort%n, currentCohort%canopy_layer, &
                         currentPatch%canopy_layer_tlai, currentCohort%treelai , &
                         currentCohort%vcmax25top,4)
+                   endif
 
                    total_patch_leaf_stem_area = total_patch_leaf_stem_area + &
                         (currentCohort%treelai + currentCohort%treesai) * currentCohort%c_area
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -1935,9 +1935,13 @@ contains
        total_patch_area = 0._r8
        total_canopy_area = 0._r8
        bc_out(s)%canopy_fraction_pa(:) = 0._r8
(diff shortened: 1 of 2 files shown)


glemieux:
All expected tests pass b4b, with the exception of the SMS satellite phenology test.  This is likely due to the fact that the baseline for the SMS test results include the use of the aerodynamic scaling update, which is now avoided for SP mode.

@rosiealice @ckoven I'm wondering if this makes sense to avoid for SP mode or not?  If I understand correctly, it looks like this gets passed to the hlm which is then brought back in `CanopyFluxes` and used to eventual update gpp.  Perhaps this actually should be re-ordered for SP mode instead of avoided?

glemieux:
After reviewing the code, I updated the it to narrow the restriction for on the sp mode check to only avoid the calls to `carea_allom`, `treelai`, and 'treesai' here: https://github.com/glemieux/fates/blob/1188a50a29fedfd2feffe32024d76a354759be2d/biogeochem/EDCanopyStructureMod.F90#L1981-L1994.  This still results in all expected regression tests passing b4b, but now sp mode includes the weighting calculation for the `bc_out` updates to `z0m`, `displa` and `dleaf`.  Additionally, I removed some dead code that I found via https://github.com/NGEET/fates/pull/821/commits/1188a50a29fedfd2feffe32024d76a354759be2d.  Test results can be found here: 
`/glade/u/home/glemieux/scratch/ctsm-tests/tests_issue_1485-dev066-sci152-1188a50a`

glemieux:
Per the [background](https://github.com/NGEET/fates/issues/823#issue-1100724763) discussion section issue #823, I've determined that the [additional calls to `carea_allom`, `tree_lai`, and `tree_sai`](https://github.com/NGEET/fates/blob/b27149270b459fc2518dc2a5bacfc48ccd256528/biogeochem/EDCanopyStructureMod.F90#L1986-L1997) are redundant based on the order of operations.  That said, the redundant calls are alleviating issue #823 (somewhat) since the understory values are being recalculated with the correct `canopy_layer_tlai` prior to calculating the aerodynamics canopy scaling values.

I think we have two options here:

1. Implement a (quick) fix for #823 in this PR and remove the redundant code.  
2. Leave the redundant code in with this PR for now and remove it later with a more comprehensive refactor for #823.

I think I'm leaning towards 1).  Thoughts?

glemieux:
>     2. Leave the redundant code in with this PR for now and remove it later with a more comprehensive refactor for [Order of operations for updating `canopy_layer_tlai` can result in `treelai` value discrepencies for understory trees #823](https://github.com/NGEET/fates/issues/823).

Per discussion with @ckoven, we've decided to move ahead with option 2) to integrate this PR as is so that we can address the other issue separately with some additional testing.

rgknox:
"That said its possible that the cohort is then set to a zero leaf biomass due to a satellite phenology update which is then written for the restart. Given that there would be zero leaf biomass, the cohort parameters would be set to zero, thus causing a mismatch in the downstream gpp_tstep update."
@glemieux I guess I don't understand this.  Shouldn't the leaf biomass be the same at the end of a base simulation, and the beginning of the restarted simualtion?

glemieux:
> Shouldn't the leaf biomass be the same at the end of a base simulation, and the beginning of the restarted simualtion?

Sorry for the confusion.  The leaf biomass is correct across the restart, but at some sites (mostly high latitude sites) it ends up being zero (per satellite phenology data) at the time of restart write.  As a result the `vcmax25top` is set to zero during the restart read via the call to `UpdateCohortBioPhysRates` since there is no leaf biomass at the time.  Later on after the restart, the sp data will set leaf biomass (via `assign_cohort_SP_properties`) to be non-zero for this particular site, but we never get to call `UpdateCohortBioPhysRates` to set the `vcmax25top` to a non-zero value.  As a result, we decided to basically just set the value if in sp mode (so I guess I don't technically need the check on the `hlm_is_restart`), since we are driving everything with sp data.

rgknox:
ok, thanks for explaining this in more detail. So as I understand, it is not that we have an inconsistency in biomass from a restart write through a restart read, it has more to do with the frequency of the biophysrate update.

rgknox (approved):

Issue #820, opened 2021-12-13
Pass Harvested C from FATES to ELM and Activate Dynamics of Wood Product Pools


sshu88:
Current ELM-FATES simulations can read land cover types and harvest rates information from surface data and pass these information in FATES. No information about harvested C flux into wood product pools is passed back to ELM. The following changes are made to pass harvested C wood products from FATES to ELM and activate the bookkeeping model in ELM when using FATES. 

First, the following new site-level variables are defined: fates%bc_out(s)%hrv_deadstemc_to_prod10c and fates%bc_out(s)%hrv_deadstemc_to_prod100c. Both variables are calculated in EDLoggingMortalityMod.F90 following the same method in ELM::dynHarvestMod::CNHarvest (dyn_subgrid/dynHarvestMod.F90), i.e., use parameter ELM::pftvarcon::pprodharv10 to determine the fraction of wood products (currentSite%mass_balance(el)%wood_product) being put into the 10-years product pool. 
I added a new subroutine UpdateHarvestC to handle the calculation because no parallel subroutines available in FATES to perform the calculation. Since logging in FATES does not distinguish logging by PFT types (only by woody or non-woody), a mean value of all woody types is used for simplicity and further parameterization of this parameter is required. 
Second, values from FATES (such as variable hrv_deadstemc_to_prod10c) are transferred to ELM variables, col_cf%hrv_deadstemc_to_prod10c and col_cf%hrv_deadstemc_to_prod100c, in elmfates_interfaceMod.F90 through a newly created subroutine ELM::elmfates_interfaceMod::wrap_WoodProducts. This subroutine is called through ELM::EcosystemDynMod::EcosystemDynNoLeaching2 before calling ELM::WoodProductsMod::WoodProducts. 
Finally, subroutine ELM::WoodProductsMod::WoodProducts is activated under the “use_fates” tag to update wood product pools in ELM and calculate land use emission.

ckoven:
Thanks @sshu88 for this update. I'm curious about the pros and cons of handling the prognostic wood product pools via a flux back to the host model versus keeping track of them at the site level within FATES (and possibly providing them as diagnostic variables to the HLM).  Do you see there being a strong scientific argument either way? What information about the product pools does the ELM bookkeeping model want and what does it do with that information?

jenniferholm:
Thanks @ckoven!  
[...]

PR #825, opened 2022-01-19
Pass Harvested C from FATES to ELM

sshu88:
### Description:

Revisions related to Issue #820 

This PR is a companion to https://github.com/E3SM-Project/E3SM/pull/5043.

### Collaborators:

@jenniferholm @rgknox @ckoven 

No change is expected.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


ckoven:
Thanks @sshu88! Is there a corresponding ELM PR or branch to see how the bc_out%hrv_deadstemc_to_prod10c and bc_out%hrv_deadstemc_to_prod100c info gets used?

sshu88:
Thank you for reminding me @ckoven. I have not created the corresponding PR in ELM repo and send a pull request yet since some part of the code passing harvest rate to FATES are added from @aldivi branch instead of the master branch. Any suggestions I shall create the PR now or I shall synchronize ELM code with @aldivi first? @jenniferholm

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -791,4 +799,72 @@ subroutine logging_litter_fluxes(currentSite, currentPatch, newPatch, patch_site
...
+          unit_trans_factor = 1000._r8 * days_per_sec / 30._r8
+      else
+          unit_trans_factor = 1000._r8 * days_per_sec / 28._r8
+      end if
+    else
+       ! Logging event one time every year
+       unit_trans_factor = 1000._r8 * days_per_sec * years_per_day
+    end if
+
+    do el = 1,num_elements
```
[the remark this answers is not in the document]
sshu88: Okay. I thought about passing isotopic C, N, P thus added this loop. Will remove the element loop at least for now.

glemieux:
@sshu88 is there an E3SM PR associated with this FATES PR forthcoming?

sshu88:
> I have not created the E3SM PR yet. I plan to check out and manually incorporate Alan's revision first and then create one E3SM PR for both this work and the C-based harvest in FATES currently I'm working on.

jenniferholm:
Following up at the corresponding ELM PR, and what is the best plan of attack?   

Originally Alan's branch that enables harvesting from ELM to FATES (https://github.com/E3SM-Project/E3SM/pull/4024) was very old. So, Shijie made his own E3SM branch (@sshu88 - what is your ELM branch name, so we can look at it?) that uses a newer version of E3SM, and manually added in Alan's changes.  But now, Alan's #4024 is pulled into E3SM master.  Should Shijie keep using his own E3SM branch where he manually added Alan's code changes from #4024, or Shijie be using the newest E3SM master, that has #4024 incorporated?

There is also another layer of work..... Shijie is in the middle of adding in C-based harvesting into FATES. But a PR has not been made for this C-based harvest yet.  **Should he make 1 ELM PR that checks out E3SM master, has Alan's updates, and also has both changes for this PR here, and also the C-based harvesting?** Does that make sense? 

Advice from @glemieux or @rgknox would be welcomed, thanks!

sshu88:
> Following up at the corresponding ELM PR, and what is the best plan of attack?
> 
> Originally Alan's branch that enables harvesting from ELM to FATES ([E3SM-Project/E3SM#4024](https://github.com/E3SM-Project/E3SM/pull/4024)) was very old. So, Shijie made his own E3SM branch (@sshu88 - what is your ELM branch name, so we can look at it?) that uses a newer version of E3SM, and manually added in Alan's changes. But now, Alan's #4024 is pulled into E3SM master. Should Shijie keep using his own E3SM branch where he manually added Alan's code changes from #4024, or Shijie be using the newest E3SM master, that has #4024 incorporated?
> 
> There is also another layer of work..... Shijie is in the middle of adding in C-based harvesting into FATES. But a PR has not been made for this C-based harvest yet. **Should he make 1 ELM PR that checks out E3SM master, has Alan's updates, and also has both changes for this PR here, and also the C-based harvesting?** Does that make sense?
> 
> Advice from @glemieux or @rgknox would be welcomed, thanks!

I only have local commits to my personal branch, which has not been uploaded to my github account yet.

sshu88:
I just pushed my ELM commit:
https://github.com/sshu88/E3SM/tree/fatesluc

glemieux:
> Should Shijie keep using his own E3SM branch where he manually added Alan's code changes from #4024, or Shijie be using the newest E3SM master, that has #4024 incorporated?

@jenniferholm and @sshu3, @rgknox and I conferred on this and we agreed that should bring the `fatesluc` E3SM branch up to date with master now that Alan's PR has been integrated into master.  Since there is only 1 commit on that branch that covers all your changes Shijie you could either start a new branch from e3sm master and `git cherry-pick` your commit onto the new branch, or your can try and just `git pull` in the latest master to your branch.  Let us know if you run into any trouble with conflicts.

> There is also another layer of work..... Shijie is in the middle of adding in C-based harvesting into FATES. But a PR has not been made for this C-based harvest yet. **Should he make 1 ELM PR that checks out E3SM master, has Alan's updates, and also has both changes for this PR here, and also the C-based harvesting?** Does that make sense?

Ryan and I agree that putting the fates side C-based harvesting updates into this PR and coordinating it with a single PR based on the `fatesluc` branch sounds like a fine plan.  The only reasons I would say to keep them separate is if the C-based harvesting is sufficiently complex enough as to worth being its own PR (which is subjective).

rgknox:
@sshu88 , @ckoven and I had a couple of questions

we noticed that site_mass%wood_product and site_in%harvest_carbon_flux are essentially doing the same thing.  This is old an code issue that you did not introduce, but one of these should be removed.  Is this something you could help us clean up?   site_mass%wood_product is used for mass conservation checks, and may be the better variable to retain, but if you have an argument to keep the other, that is fine too.

also, we were wondering about the input variable pprodharv10_forest_mean.  We wanted to double check with you, why is this an input boundary condition?  Is this dynamic, and requires information generated on the clm/elm side of the code, or is it a constant that we can potentially put in the fates parameter file?

sshu88:
> Both harvest_carbon_flux (kg/m2/day) and wood_product (kg/site/day) are already defined in previous version of the code and they both are harvested C flux. I guess one is the output variable for history file and another one is for C balance check. I can keep wood_product and remove harvest_carbon_flux.
pprodharv10_forest_mean is the mean **harvest mortality proportion of deadstem to 10-yr product pool (pprodharv10)** of all woody PFT types. This parameter is a constant and calculated in the elm fates interface based on the parameter **pprodharv10**, which is a ELM parameter in pftvarcon.f90.

ckoven:
HI @sshu3 Thanks! Yeah, I think with this code, we don't need `harvest_carbon_flux` and the related history variables anymore, so that would be great if you deleted them and only used the `wood_product` variable.

For the `pprodharv10`, I think it would be preferable to add this as a new parameter to the FATES parameter file rather than passing it via the interface from ELM and CLM -- we've tried to keep all the FATES parameters there rather than passing them around, as that keeps it much easier to trace the provenance of parameter values.

sshu88:
> HI @sshu88 Thanks! Yeah, I think with this code, we don't need `harvest_carbon_flux` and the related history variable anymore, so that would be great if you deleted them and only used the `wood_product` variable.
> 
> For the `pprodharv10`, I think it would be preferable to add this as a new parameter to the FATES parameter file rather than passing it via the interface from ELM and CLM -- we've tried to keep all the FATES parameters there rather than passing them around, as that keeps it much easier to trace the provenance of parameter values.

I will remove `harvest_carbon_flux` and redefine pprodharv10_forest_mean as a FATES parameter. Thank you for your suggestions @ckoven @rgknox

sshu88:
I have modified the code. The corresponding ELM commit ([fatesluc_merged](https://github.com/sshu88/E3SM/tree/fatesluc_merged)) is also updated.

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -791,4 +811,74 @@ subroutine logging_litter_fluxes(currentSite, currentPatch, newPatch, patch_site
...
+      ! Flush the older value before update
+      if(logging_time) then
+         bc_out%hrv_deadstemc_to_prod10c = 0._r8
+         bc_out%hrv_deadstemc_to_prod100c = 0._r8
+      end if
+  
+      ! First test tropic forest (PFT=1)
+      ! Calculate the unit transfer factor (from kgC m-2 day-1 to gC m-2 s-1)
+      ! for icode == 2, icode < 0, and icode > 10000 is one time harvest, thus
+      ! shall distribute into everyday
```
rgknox, line 847: @glemieux just brought up this point with me:  @sshu88 are we sure that we need different conversion factors for different event codes?  The wood_product is a rate, it isn't time integrated over the month.  It seems that the conversion factor should be 1000 * days_per_sec for all cases.  Can you clarify please?

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -791,4 +811,74 @@ subroutine logging_litter_fluxes(currentSite, currentPatch, newPatch, patch_site
...
+         ! Logging event once a month
+         ! Shijie: Shall think about a better if expreession?
+         if ((hlm_current_month == 1) .or. (hlm_current_month == 3) .or. &
+             (hlm_current_month == 5) .or. (hlm_current_month == 7) .or. &
+             (hlm_current_month == 8) .or. (hlm_current_month == 10) .or. &
+             (hlm_current_month == 12)) then
+            unit_trans_factor = 1000._r8 * days_per_sec / 31._r8
+        else if((hlm_current_month == 4) .or. (hlm_current_month == 6) .or. &
+            (hlm_current_month == 9) .or. (hlm_current_month == 11)) then
+            unit_trans_factor = 1000._r8 * days_per_sec / 30._r8
```
glemieux, line 865:
```suggestion
            unit_trans_factor = g_per_kg * days_per_sec / 30._r8
```
Recommend replacing with existing conversion constant.

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -61,7 +62,7 @@ module EDLoggingMortalityMod
    use FatesAllometryMod , only : set_root_fraction
    use FatesConstantsMod , only : primaryforest, secondaryforest, secondary_age_threshold
    use FatesConstantsMod , only : fates_tiny
-   use FatesConstantsMod , only : months_per_year
+   use FatesConstantsMod , only : months_per_year, days_per_sec, years_per_day
```
glemieux, line 65:
```suggestion
   use FatesConstantsMod , only : months_per_year, days_per_sec, years_per_day, g_per_kg
```

glemieux (changes requested):
A few minor updates to facilitate code reuse

biogeochem/EDLoggingMortalityMod.F90
rgknox: sorry, I still don't think I understand.  It seems to me that the wood_product rate will always have the same units, regardless of how many times logging occurs per month.  Appologies if I'm missing something obvious here.

sshu88: Sorry for the confusion. I discussed this issue with @jenniferholm and it seems like my code comments misled users. 

`wood_product` is in the unit of kgC m-2 day-1 and represent the harvested C flux, but not yet the input fluxes into the bookkeeping model in ELM (`bc_out%hrv_deadstemc_to_prod10c` and `bc_out%hrv_deadstemc_to_prod100c`). ELM LULCC perform logging once a year and distribute the harvested C into s-1 rate, i.e., a constant harvest rate for each model timestep. I added this piece of code to **distribute the harvested C flux from one day to a constant s-1 rate for each model timestep until a new logging event happen** for different type of logging event code (once a year, per month or everyday, etc.), thus to keep consistent to the behavior of ELM. 

Here attaches a figure showing the difference of the size of total product C pools between 1) the case directly putting all logged C into product pool ("spike") and 2) the case distributing the logged C to each day across the year then send C to product pool ("uniform case"):
comparison_uniform_vs_spike

Since this is a ELM specific feature, and also can be achieved by changing the logging event code in FATES parameter file, I will follow your suggestion to revise the code.

@jenniferholm, let me know if I have missed something.

Thank you for checking out this issue.

rgknox: Thanks @sshu88 now I understand

jenniferholm: Hi Shijie - thanks for doing these two comparison tests. 1) putting the logged C into product pool within one day, and 2) averaging the C product pool over the whole year, which I believe you said is what ELM prefers to use, right?   Could you please clarify for me what the decision is for the FATES code?  
Are you going to have the harvested C in FATES be HLM agnostic, and put all the harvest C into the product pool in a single time step (spike case), then on the ELM side do the unit transfer to constant s-1 rate?  Sorry if you are repeating yourself, I just wanted to double check.

sshu88: @jenniferholm Thank you for your comment. 

I decided to add these codes to avoid spurious response from ELM bookkeeping model if we provide "spiky" C input to ELM bookkeeping model, thus the simulation results shall look like what "uniform" case showed no matter what logging event code we provide.  But now I have tested both cases and the ELM bookkeeping model can handle "spiky" inputs, although the results varied due to a different timing of inputs. I think we can remove these codes and use the FATES logging event code to control how "spiky" the C input to product pool. Any further suggestions?

aldivi:
I agree that the amount and timing of inputs to the host product pools should be dictated by the harvest timing in fates, without smoothing in the host.

rgknox (approved):

rgknox:
all tests OK: /glade/scratch/rgknox/tests_0627-095210ch

Issue #578, opened 2019-09-26
Update PFT optical properties per Majasalmi and Bright 


jkshuman:
@olyson identified a recent manuscript evaluating LSMs that recommends updating PFT optical properties based on various spectral databases. Many of the FATES param values are drawn from historical CLM PFT values, so will likely need updating. Specifically, there were notable differences between CLM and measured values in the NIR, particularly for conifer trees. 

Majasalmi, T. and R. M. Bright, 2019: Evaluation of leaf-level optical properties employed in land surface models. Geoscientific Model Development, 12, 3923-3938.

@olyson is testing these updates in CLM per https://github.com/ESCOMP/ctsm/issues/807
When this is complete we will update the default FATES_param file for relevant values and PFTs based on @olyson recommendations.

jkshuman:
Modify FATES optical parameters per CLM PFT optical params per file ctsm51_params.c210923.nc
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/jkshuman/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/jkshuman/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



FATES |   |   |   |   |   |   |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
parameter | BETropT | NEExTropT | NCDExTropT | BEExTropT | BHDTropT | BCDExTropT | BEExTropS | BHDExS | BCDExS | C3arcG | C3nonarcG | C4
CLM param | BETroT | NETemT | NDBorT | BETemT | BDTroT | BDTemT | BES | BDTemS | BDBorS | C3ArcG | C3nonArcG | C4
rholnir | 0.46 | 0.41 | 0.39 | 0.46 | 0.41 | 0.41 | 0.46 | 0.41 | 0.41 | 0.28 | 0.28 | 0.28
[...]

PR #831, opened 2022-01-28
Update FATES PFT optical to Majasalmi Bright 2019

jkshuman:
FATES PFT optical params are updated to those published
in Majasalmi & Bright 2019 GeoSciModelDev and are
consistent with CLM PFT optical params

Fixes: #578

User interface changes?:No

Code review: none

Test suite: not tested
Test baseline:
Test namelist changes:
Test answer changes: will change answers
Test summary:

### Description:
This updates the FATES PFT optical parameters to those published in Majasalmi & Bright 2019 GeoSciModelDev, so FATES PFT are are now consistent with CLM PFT optical parameters. Mapping of CLM PFT to FATES PFT included in Issue #578 

This updates the default cdl file only.  @glemieux indicated there is a separate process to update the .nc file for the HLM. I will share this cdl file with folks for testing, or you can grab the cdl and generate a .nc file for use.

This PR is associated with https://github.com/ESCOMP/CTSM/pull/1681

### Collaborators:
@olyson @glemieux 

yes, answer changing. Given FATES-SP testing by @rosiealice expected to change answers.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


rosiealice (approved):
This all looks pretty straightforward. I can't vouch for whether the numbers are all aligned exactly, (I literally can't be trusted with those sorts of things) but they look to all be in the right categories.  Thanks for doing this Jackie :)

adrifoster:
I ran an ILAMB comparison using a FATES-SP run for this change:

output here: https://webext.cgd.ucar.edu/I2000/FATES_SP_2deg/lnd/_build_fates_sp_opt/index.html

2-degree resolution (f19_g16)
compset: 2000_DATM%GSWP3v1_CLM50%FATES_SICE_SOCN_SROF_SGLC_SWAV
years: 1965-2014
user_nl_clm changes:
```
 fates_parteh_mode = 1
 fates_spitfire_mode = 0
 use_fates = .true.
 use_fates_cohort_age_tracking = .false.
 use_fates_ed_prescribed_phys = .false.
 use_fates_ed_st3 = .false.
 use_fates_fixed_biogeog = .true.
 use_fates_inventory_init = .false.
 use_fates_logging = .false.
 use_fates_nocomp = .true.
 use_fates_planthydro = .false.
 use_fates_sp = .true.
```
cases are located on Cheyenne: 
master: /glade/work/afoster/FATES_SP/cases/FATES_SP_mster, /glade/scratch/afoster/FATES_SP_mster
update: /glade/work/afoster/FATES_SP/cases/FATES_SP_pft-optical, /glade/scratch/afoster/FATES_SP_pft-optical


1. CTSM5.1_FATES_SP_master: master code base (default parameter file)
2. CTSM5.1_FATES_SP_pftopt: https://github.com/jkshuman/fates/tree/pft-optical code (should be the same as master), but with the parameter file generated from the updated .cdl file

Seems like it's slightly *worse* with these updates, but not a ton different in general.

adrifoster:
Also, perhaps this is done automatically but if/when we do bring in these changes we should (?) update the default .nc parameter file on Cheyenne and Izumi (and other machines?)

glemieux:
> Also, perhaps this is done automatically but if/when we do bring in these changes we should (?) update the default .nc parameter file on Cheyenne and Izumi (and other machines?)

@adrifoster there is a workflow to update this during API update, but it looks like we might not have it written down anywhere public.  It's not automatic currently.  In short Ryan or I run `ncgen` on Cheyenne and copy it into the necessary input directory.  We then run a script called `rimport` from the input directory to make sure it gets copied over to the svn repo so others can automatically download up a case run.   I'll add that to the list to put in the developer's guide for the future.

parameter_files/fates_params_default.cdl
```diff
@@ -1286,10 +1286,10 @@ data:
  fates_smpso = -66000, -66000, -66000, -66000, -66000, -66000, -66000,
     -66000, -66000, -66000, -66000, -66000 ;
 
- fates_taulnir = 0.25, 0.1, 0.1, 0.25, 0.25, 0.25, 0.1, 0.25, 0.25, 0.34,
-    0.34, 0.34 ;
+ fates_taulnir = 0.33, 0.32, 0.42, 0.33, 0.43, 0.43, 0.33, 0.43, 0.43, 0.4,
+    0.4, 0.4 ;
 
- fates_taulvis = 0.05, 0.05, 0.05, 0.05, 0.05, 0.05, 0.05, 0.05, 0.05, 0.05,
+ fates_taulvis = 0.06, 0.04, 0.04, 0.06, 0.06, 0.06, 0.06, 0.06, 0.06, 0.05,
```
jkshuman, line 1292:
```suggestion
 fates_taulvis = 0.06, 0.04, 0.06, 0.06, 0.06, 0.06, 0.06, 0.06, 0.06, 0.05,
```

jkshuman:
Thank you to @adrifoster for testing with ILAMB. From the Majalsami and Bright paper https://gmd.copernicus.org/articles/12/3923/2019/ they went though various spectral databases to synthesize and harmonize key optical properties used by many land surface models. CLM5 was used as an example, and updates were recommended (Table S1) and implemented by @olyson and then here. In summary the updates were generally for the NIR. 

Given that the ILMAB comparison for the FATES implementation identified potential degradation of performance (though small), it is worth noting that the changes recommended for the CLM PFTs in table S1 are slightly different from the main text table 3 summary average. We could consider testing the values from table 3. It is also worth considering the additional recommendations of the manuscript: defining shrubs based on forest PFTs, importance of variation in leaf angle distribution (LAD) (especially for grasses). 

Also, @serbinsh could weigh in on these changes, and perhaps provide expert guidance as Serbin (2014) data was used for the changes compiled in this manuscript, and @serbinsh has new spectral data.

ckoven:
@jkshuman it also seems like, given that @rosiealice is seeing unexpected responses to changing the bark albedo parameters in https://github.com/NGEET/fates/issues/837, we should probably wait until that is sorted out before proceeding here, especially since this PR makes the already high albedo bias even worse?

rosiealice:
You could also argue that I'm using homogenized optical properties, and there is an (admittedly low) chance my weird response is causes by that? 

My next plan for that is to check I didn't make an error in the parameter files, fwiw. But today I have COVID 🙄

glemieux (approved):
I double checked the update against the source clm parameter file and the recommended changes in the original paper and found one minor translation error that Jackie quickly corrected.

adrifoster:
Sorry this took me so long - here is an updated ILAMB comparison for the typo fix. Basically the same:

https://webext.cgd.ucar.edu/I2000/FATES_SP_2deg/lnd/_build_FATES_SP_pftopt/index.html

serbinsh:
All, I totally missed this conversation!  In fact the Majalsami and Bright paper is an analysis I was thinking about doing for awhile, never go around to it, and am happy to see it done!  I need to explore the recommended new VIS/NIR albedo values by PFT but they seem reasonable on first blush.

I am pretty sure I also did some FATES runs a few years ago and presented on the impact of changing the leaf-level and other single-scattering albedos.  However, I do recall thinking the FATES defaults could be updated/improved.  

At the same time, I do want to provide a word of caution that they may seem like small changes but they can have significant impacts on related downstream processes. For example: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0216512.  That is an ED2 study but I think you can see how we may have very similar issues when changing FATES defaults for single-scattering albedo, LAD etc.  Not suggesting we do not do this, but that its going to require further pain and iteration, at least thats what I assume, given the highly-tuned nature of default parameters and expected FATES behavior 

@jkshuman I can give more guidance if you wanted to chat about it?  I also have more data for tropical species where we could very easily use to create additional constraints on Tropical single-scattering albedo.

jkshuman:
@serbinsh thanks for your input on this. @serbinsh recommendation is in line with what we had discussed:  merge this PR, but then open a separate issue and separate PR for further testing on the impact of these changes. 

@serbinsh and I had discussion on FATES slack about this PR and associated issue. @serbinsh indicated that the "mean suggested values will be fine, but I would be curious what would happen if we ran ensembles across the range of values to see what sort of FATES behavior we get" @serbinsh also offered to help figure out approaches on how to do this by doing some data mining to understand the potential parameter variability across PFTs. @serbinsh noted that "optical properties are important but also clumping (which we already included) and LAD assumptions, which may be more important for growth or competition especially for young or very old stands (lots of trees, or fewer larger trees/cohorts)"

serbinsh:
Thanks @jkshuman, yes exploring how we can inform/constrain/or test the impacts of "optical properties" or otherwise those properties will influence RT and thus other associated processes is something I am interested in helping with.  its something I was thinking about awhile back but before we had a much more stable FATES code.  For example, for reference her is some UQ results I presented to the modeling group in 2018 or so, which I wouldn't consider to be anything more than a guide given its based on a much earlier version of the model

Example_FATES_UQ_w_optical_properties

This is a simulation I had done where I allowed various params to vary either based on data distributions or uninformed, including temperature response params but also some of the RT params.  You can see some of the most influential were LAD and transmittance properties.  The response variable was GPP.

At the time I didnt continue this given a lot of changes to FATES but I think we are in an even better place to work on this now; more datasets we can use to inform how we parameterize and express possible param variation 

I think this (and the earlier ED2 work) does show that in VDMs these can have influence on carbon uptake and surface energy balance so it would be fun and interesting to develop some approaches to dig into this.  I like the idea @walkeranthonyp has for exploring in a reduce complexity framework as well.  We have also talked about just trying to explore optical trait space based on our datasets of leaf-level optical properties.  

Happy to discuss more as I think this has implications for some other issues raised recently for albedo response; something else I noticed at some sites that I think I reported on before, like flat albedo responses or very high albedo at temperate broadleaf forests.

glemieux:
Test results against the baseline using the fates suite on Cheyenne results in passing tests with DIFFs as expected.

`/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr831-check`

glemieux:
Final regression testing post hotfix tag shows all tests run with DIFFs as expected:
Cheyenne: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr831-pftopt-2`
Izumi: `/scratch/cluster/glemieux/ctsm-tests/tests_pr831-pftopt`

PR #838, opened 2022-02-17
Satellite phenology mode fixes for the gnu compiler

### LINKED ISSUE #832: Satellite phenology mode run fails with gnu compiler
Additional testing of satellite phenology mode with E3SM using the gnu compiler resulted in the follow issue: https://github.com/E3SM-Project/E3SM/issues/4755.  SP mode hadn't been tested with the gnu compiler as both our clm and elm regression tests for this mode utilize the intel compiler.

Replicating the issue on Cori the results in the following error immediately during initialization:
```
378:  in sub-canopy layer in leafc_from_treelai
378:  this is not set up to work for lower canopy layers.
378:  problem in leafc_from_treelai         379           2
378:  ENDRUN:ERROR in /global/u1/g/glemieux/E3SM/components/elm/src/external_models/fates/biogeochem/FatesAllometryMod.F90 at line 811                                                                                                                                                                                                                                                                                                                                                       
```

Tracing the code, I determined that the issue has to do with the `cohort%canopy_layer` being un-initialized and not specific to elm-fates.  This is pretty clear from the error message above as the `379` is the `canopy_layer` trying to be indexed here, which doesn't make sense for fates generally.  I confirmed this with a test run on Cheyenne as well.  I've implemented a simple fix to assign the temporary starting `canopy_layer` value to be one, but this has exposed a couple other issues with sp mode initialization as well that need to be addressed.

Note: sp mode will run with the gnu compiler if the case or test is set up in debug mode.

glemieux:
### Description:
This fixes #832 by explicitly setting the cohort canopy layer to one when initializing cohorts during satellite phenology (SP) mode.  In the course of correcting this issue it was also discovered that the SP mode was failing the exact restart test (ERS) due to `FATES_FDI` not being initialized (even though we force spitfire to be off during the SP runs).  This history variable output is the only fire output that is a direct assignment, which is why I think it might not have been showing up for non-fire gnu compiler ERS runs.  Setting the site-level `fdi` equal to zero in the `zero_site` subroutine corrects this issue.

I also realized that we don't specifically care about the `hlm_is_restart` in `UpdateCohortBioPhysRates`.  We can just check if we are in SP mode.

### Collaborators:
@peterdschwartz

This should be b4b against the baseline except for the SatPhen testmod.

### Checklist:

### Test Results:
The ERS gnu compiler test is not standard on Cheyenne or Cori, but I tested it for completeness sake:
`/glade/u/home/glemieux/scratch/ctsm-tests/tests_spmode-gnufail-gnuzero`

CTSM (or) E3SM (specify which) test hash-tag: https://github.com/ESCOMP/CTSM/commit/bcb9735e8128b2b83f11e72aff55c1d816ea2511

CTSM (or) E3SM (specify which) baseline hash-tag: https://github.com/ESCOMP/CTSM/commit/bcb9735e8128b2b83f11e72aff55c1d816ea2511

FATES baseline hash-tag:  477a8d57815c6246a5ba8466255e0e9c56066da5

Test Output: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr838-gnu_spmode_fix`


biogeochem/EDCohortDynamicsMod.F90
```diff
@@ -1995,6 +1995,8 @@ subroutine UpdateCohortBioPhysRates(currentCohort)
 
        ipft = currentCohort%pft
 
+       write(fates_log(),*) 'UpdateCohortBPRates: frac_leaf_aclass: ', sum(frac_leaf_aclass(1:nleafage))>nearzero
```
rgknox, line 1998: remove this line please when done with internal testing/debugging

glemieux: Good catch.  Thanks.

glemieux:
All expected tests PASS b4b against the baseline.  Test folder location:
`/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr838-gnu_spmode_fix`

glemieux:
I'm running a last check on Cori using a version of this PR branch that is compatible with the current ELM-FATES API: https://github.com/glemieux/fates/commits/spmode-gnufix.

rgknox (approved):

glemieux:
> I'm running a last check on Cori using a version of this PR branch that is compatible with the current ELM-FATES API: https://github.com/glemieux/fates/commits/spmode-gnufix.

For some reason this failed on Cori.  This is still on hold until I figure out what happened.

glemieux:
Ok, I think I see what the issue is here.  There is another initialization issue: we're calling `assign_cohort_SP_properties` prior to setting `vcmax25top` with `UpdateCohortBioPhysRates`.  The quick fix for this is to add a call to the latter during within the former; I've got a test for that queued on Cori.  That said, I'm thinking about spending time to see if we can refactor this in some way.  Doing the short term fix would work, but it seems inconsistent since there will be a downstream call to `create_cohort` that will `NaN` a whole bunch of cohort variables including `vcmax25top`.

glemieux:
After reviewing options for refactoring the code and talking with @ckoven, I've moved ahead with implementing the simple immediate fix of assigning the `vcmax25top` values during sp mode initialization (via https://github.com/NGEET/fates/pull/838/commits/5d40c3e4f86b2dc9b304b14277397f508455835a).  Implementing this update now resulting in a passing gnu compiler sp mode fates case.  The output for that test case is here on Cori:
`/global/homes/g/glemieux/scratch/e3sm-cases/satphenmatch-clfix_FDIfix-vcmaxfix3.fates-sci.1.50.0_api.17.0.0-v2.0.0-C3a1b5ed84b-Fb78153de.gnu`

rgknox:
Tests OK:
/glade/scratch/rgknox/tests_0302-090114ch/

glemieux:
With the change from 5d40c3e the `ReducedComplexSatPhen` testmods will not be b4b against the baseline now.  I confirmed that this is because prior to this change the `vcmax25top` value during initialization was zero everywhere (for the intel compiler).  This can be seen by comparing the output from the following folders:
- With 5d40c3e: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_spmode-pr838-revertdiag_wfix-more`
- With out 5d40c3e: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_spmode-pr838-revertdiag_wofix-more`

PR #848, opened 2022-03-18
Add logo to front page

ckoven:
We never added the logo that we had decided on a while ago to our GitHub repository, so here it is.

No other code changed, so no testing.


glemieux:
I should piggy-back on this and add the user's guide to the Readme.

glemieux:
@ckoven is there a version of this picture with a transparent background?

ckoven:
> @ckoven is there a version of this picture with a transparent background?

I don't have one but you're right we should do that.  I'll either track one down or make one.

glemieux:
I found a really large png logo from an official ngeet email.  I've uploaded the three different sized versions to a new `images` folder in the `.github` directory.

PR #849, opened 2022-04-06
add the continous taper function

xuchongang:
Based on our sensitivity analysis, the taper function play a key role regulating plant hydrodynamics. In the original code, the taper exponent can only take 3 discrete values: 1/3, 1/6, and 1. This pull request added a continuous function to interpolate the exponents through an empirical polynomial. 
### Description:
The main change is in the xylemtaper function in FatesPlantHydraulicsMod.F90.

### Collaborators:
@bchristo @rgknox 

The hydrodynamics will change as the exponents are based on the empirical polynomial, but should be close to the old code. 

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeophys/FatesPlantHydraulicsMod.F90
+++ b/biogeophys/FatesPlantHydraulicsMod.F90
@@ -4629,8 +4629,7 @@ function xylemtaper(pexp, dz) result(chi_tapnotap)
     real(r8) , intent(in) :: dz     ! hydraulic distance from petiole to node of interest[m]
     !
     ! !LOCAL VARIABLES:
-    real(r8) :: qexp                ! total conductance exponent (as in Fig. 2b of Savage et al. (2010)
-    real(r8) :: qexp_notap          ! same as qexp, but without xylem taper; i.e., pexp = 0
+    real(r8) :: qexp                ! total conductance exponent (as in Fig. 2b of Savage et al. (2010) minus a0 term
     real(r8) :: lN=0.005_r8         ! petiole length[m]
     real(r8) :: mu=1.0E-03_r8       ! viscosity[kg m-1 s-1]
     real(r8) :: rintN=0.00001_r8    ! internal conduit diameter in petiole[m]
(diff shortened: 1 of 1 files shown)


rgknox:
In an email conversation, @xuchongang , @bchristo @glemieux and I discussed some of the details of this routine, particularly how the scheme uses a fixed petiole conductance to arrive at tapers.  After the discussion we believe that this was used to derive the parameters and is not incompatible with hydro's assumptions about petiole conductance.

rgknox (approved):

rgknox:
@glemieux you mentioned in another thread that the cohort level error errh2o needs to be zero'd in certain contexts. Is this a related issue?

rgknox:
or maybe this is a symptom of #878 ?

glemieux:
> @glemieux you mentioned in another thread that the cohort level error errh2o needs to be zero'd in certain contexts. Is this a related issue?

That's a good thought, but it doesn't appear to be; I tested out that fix in this branch without success.  

> or maybe this is a symptom of https://github.com/NGEET/fates/issues/878 ?

Good memory, I'll look into this tomorrow.

glemieux:
> > @glemieux you mentioned in another thread that the cohort level error errh2o needs to be zero'd in certain contexts. Is this a related issue?
> 
> That's a good thought, but it doesn't appear to be; I tested out the fix from commit [9d9c192](https://github.com/NGEET/fates/commit/9d9c192f184431c5f329ac9ed195ee9de10cc952) in this branch without success.
> 
> > or maybe this is a symptom of #878 ?
> 
> Good memory, I'll look into this tomorrow.

Implementing the suggested fix from #878 doesn't improve the restart issue.

glemieux:
> I'm a little baffled as to why this change would cause such a restart failure since the change is so narrowly focused on `xylemtaper`.

The issue appears to be due to numerical precision in the calculation of `ktap` and `knotap` values.  The differences between the first and restart runs are on the order of `E-23`. This then propagated through the system resulting in differences in the difference in the total water mass after iteration in `ImTaylorSolve1D`.    

Upon closer inspection, I realized that `chi_tapnotap`  could be condensed to avoid intermediate calculations.  See https://github.com/NGEET/fates/pull/849/commits/78435acc986f9a576b5315ae0057477fe7e85853.  With this change the exact restart passes.

glemieux:
Re-running regression tests against the `fates-sci.1.58.0_api.24.0.0-ctsm5.1.dev100` baseline returns b4b results aside from the expected diffs in the `hydro` testbeds (and two other expected failures).

rgknox:
nice fix @glemieux , looks good from here

PR #851, opened 2022-04-07
improve hydro stability

xuchongang:
This pull request improves the numerical stability in the situation of extreme dry conditions
### Description:
Hydro uses the transpiration flux from last time step. For numerical stability, this pull request make sure that transpiration flux is less than 80% of total water availability. If it is larger than that, it will adjust the transpiration flux and record the excessive demand to the plant water storage pool, which will be dumped later in HLM.This should not affect the science of hydrodynamics as this basically delays the water limitation by a few steps with a benefit of numerical stability. 

### Collaborators:
@bchristo @rgknox 

This will not affect most of simulation and should only affect simulations with very dry soil.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


rgknox:
@xuchongang , I think we might need to subtrace the trans_err term to the plant stored water output boundary condition, here:

https://github.com/NGEET/fates/blob/aa290b3b8ad95dc86214cb3ac0ff7772128514cf/biogeophys/FatesPlantHydraulicsMod.F90#L1810-L1812

That way when E3SM and CTSM do their calculations on mass conservation, it will include this water we needed to create.

xuchongang:
> @xuchongang , I think we might need to subtract the trans_err term to the plant stored water output boundary condition, here:
> 
> https://github.com/NGEET/fates/blob/aa290b3b8ad95dc86214cb3ac0ff7772128514cf/biogeophys/FatesPlantHydraulicsMod.F90#L1810-L1812
> 
> That way when E3SM and CTSM do their calculations on mass conservation, it will include this water we needed to create.
> 
> Thoughts?

@rgknox , you are correct. I forget to copy that over as I am testing on the ELM, which is not the most recent compared to master. Will do!

rgknox (approved):
@xuchongang and I worked through a unit fix, which was applied.  Looks good.

PR #857, opened 2022-04-19
Active crown fire (Scott & Reinhardt 2001)

slevis-lmwg:
### Description:
This PR follows in the footsteps of
https://github.com/NGEET/fates/issues/573
https://github.com/NGEET/fates/pull/572
https://github.com/jkshuman/fates/pull/5
https://github.com/jkshuman/fates/tree/active_crown_Scott_2001
in getting the FATES fire module closer to having an active crown fire parameterization.

### Collaborators:
@jkshuman @lmkueppers @pollybuotte @ckoven @glemieux

At least when `fates_fire_active_crown_fire = 1` and active crown fires do occur.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


fire/SFMainMod.F90
```diff
@@ -540,15 +725,29 @@ subroutine rate_of_spread ( currentSite )
...
+          ! and residence time (min), Eq 3 Scott & Reinhardt 2001
+          time_r = 12.595 / currentPatch%fuel_sav 
+          heat_per_area = ir * time_r          
+
+          ! calculate torching index based on wind speed and crown fuels 
+          ! ROS for crown torch initation (m/min), Eq 18 Scott & Reinhardt 2001
+          ! TODO slevis: Are we missing a -phi_s after the -1?
+          ROS_torch = 1._r8 / (54.683_r8 * wind_reduce) * &
+    ((60._r8 * passive_crown_FI * currentPatch%fuel_bulkd * eps * q_ig / &
+      (heat_per_area * ir * xi) - 1._r8) / (c * beta_ratio**-e))**(1._r8 / b)
```
slevisconsulting, line 748: Moving post from https://github.com/slevisconsulting/slevis_fates_work/pull/1

@slevisconsulting wrote
`are we missing a -phi_s term after the -1`
@jkshuman replied
`that is for the slope. I meant to add it in as a placeholder with a value of zero as we do not have slope currently. So you can add a parameter "slope_factor" and set it to zero.`

slevisconsulting: Thanks, @jkshuman

fire/SFMainMod.F90
```diff
@@ -448,7 +448,7 @@ subroutine  characteristics_of_crown ( currentSite, canopy_fuel_load, passive_cr
 
                 !sort crown fuel into bins from bottom to top of crown
                 !accumulate across cohorts to find density within canopy 1m sections
-                do ih = int(height_cbb), int(currentCohort%hite)
+                do ih = max(1, nint(height_cbb)), max(1, nint(currentCohort%hite))
                    biom_matrix(ih) = biom_matrix(ih) + crown_fuel_per_m
```
slevisconsulting, line 452: This was needed to avoid attempting
`biom_matrix(0) = biom_matrix(0) + crown_fuel_per_m`

fire/SFMainMod.F90
```diff
@@ -737,15 +746,20 @@ subroutine rate_of_spread ( currentSite, ROS_torch, passive_crown_FI, heat_per_a
...
-      (heat_per_area * ir * xi) - 1._r8) / (c * beta_ratio**-e))**(1._r8 / b)
+          ! temp1 requires max(0... to avoid raising a negative value to a
+          ! fractional temp2 power.
+          ! TODO slevis: omitting "- phi_s" after the -1 bc phi_s = 0 for now.
+          !              @jkshuman recommended naming it slope_factor.
+          temp1 = max(0._r8, (60._r8 * passive_crown_FI * &
+  currentPatch%fuel_bulkd * eps * q_ig / (heat_per_area * ir * xi) - 1._r8) / &
+  (c * beta_ratio**-e))
+          temp2 = 1._r8 / b
+          ROS_torch = temp1**temp2 / (54.683_r8 * wind_reduce)
```
slevisconsulting, line 762: I added temp1 and temp2 to make Eq 18 easier to read. Eq 18 took me a while to debug. What eluded me until this morning was that `temp2` could be less than 1 at the same time as `temp1 < 0`, which causes a floating point exception. I added the max(0... to temp1 and the model now runs!

I completed 12 months using this test with `fates_fire_active_crown_fire = 0` in the params file:
`SMS_Lm12_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruRsGs.izumi_intel.clm-FatesFireLightningPopDens`
Next I'm rerunning with `fates_fire_active_crown_fire = 1` in the params file. I will update this post when I know.

slevisconsulting: The test with `fates_fire_active_crown_fire = 1` also completed 12 months. As soon I will get a chance I will check whether running with vs without active crown fire gave different answers and will update this post.

slevisconsulting: For now this test gives bit-for-bit the same answers with and without active crown fire.

fire/SFMainMod.F90
```diff
@@ -322,19 +337,180 @@ subroutine  charecteristics_of_fuel ( currentSite )
...
+
+    do while(associated(currentPatch))
+       !zero Patch level variables
+       height_base_canopy                   = 0.0_r8
+       canopy_fuel_load                     = 0.0_r8
+       passive_crown_FI                     = 0.0_r8
+       max_height = 0._r8
+       biom_matrix(:) = 0._r8
+
+!      if (currentPatch%active_crown_fire == 1) then
```
slevisconsulting, line 409: @jkshuman I see your comment here. I'm not putting the if-statement here because it will prevent `passive_crown_FI` from becoming greater than 0 when we run without active crown fire.

I am adding the if-statement later, after the comment
`! check threshold intensity and rate of spread`
That's the only place where I think this if-statement is required.

main/EDParamsMod.F90
```diff
@@ -62,7 +62,7 @@ module EDParamsMod
    real(r8),protected, public :: ED_val_canopy_closure_thresh ! site-level canopy closure point where trees take on forest (narrow) versus savannah (wide) crown allometry
    integer,protected, public  :: stomatal_model  !switch for choosing between stomatal conductance models, 1 for Ball-Berry, 2 for Medlyn
 
-   logical,protected, public :: active_crown_fire        ! flag, 1=active crown fire 0=no active crown fire
+   logical,protected, public :: active_crown_fire_switch  ! flag, 1=active crown fire 0=no active crown fire
```
slevisconsulting, line 65: Added suffix `_switch` to prevent compiler error that complained about same name as `subroutine active_crown_fire`.

slevisconsulting:
I haven't worked on this PR this week due to other priorities. I expect to get back to it next week.

fire/SFMainMod.F90
```diff
@@ -835,6 +1037,268 @@ subroutine  area_burnt_intensity ( currentSite, bc_in )
...
   end subroutine area_burnt_intensity
 
+  !*****************************************************************
+  subroutine  active_crown_fire ( currentSite)
+  !*****************************************************************
+
+    !evaluates if there will be an active crown fire based on canopy fuel and rate of spread
+    !returns final rate of spread and fire intensity in patch with added fuel from active crown fire.
+    !currentCohort%fraction_crown_burned is the proportion of crown affected by fire
+    ! TODO slevis: Jackie, you had ROS_torch coming in here but not used
```
slevisconsulting, line 1047: This far, the test
`SMS_Lm12_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruRsGs.izumi_intel.clm-FatesFireLightningPopDens`
returns identical answers when I set
`fates_fire_active_crown_fire = 1`  in the params file, as when I set
`fates_fire_active_crown_fire = 0`

@jkshuman before I begin investigating (site too wet? no lightning? maybe it's time to switch to the site that @pollybuotte is running), I wanted to ask you about two variables:
- You had `ROS_torch` entering subroutine active_crown_fire but **not used**.
- You also had `canopy_bulk_denisty` calculated but **not used**.

Had you intended to use these variables? If so, where? If the answer is in S&R (2001), feel free to refer me there.

By the way, I stopped passing ROS_torch and four other variables as subroutine arguments, because my sense was that they wouldn't work as scalars and needed to be currentPatch% variables.

jkshuman: @slevisconsulting canopy_bulk_density can be removed. it is no longer used.

fire/SFMainMod.F90
```diff
@@ -1218,16 +1218,19 @@ subroutine  active_crown_fire ( currentSite)
                 ! calculate canopy fraction burnt, Eq 28 Scott & Reinhardt App A 
                 ! using wind at which ROS_active = ROS_active_min
                 ! solve for phi_wind associated with ROS_active_min
-                phi_wind_ROS_SA = (ROS_active_min * fuel_bd * eps * q_ig - 3.34 * ir * xi) &
-                                     /(3.34 * ir * xi)
+                if (ir > 0._r8 .and. xi > 0._r8) then
+                   phi_wind_ROS_SA = (ROS_active_min * fuel_bd * eps * q_ig - 3.34 * ir * xi) / (3.34 * ir * xi)
+                else
+                   phi_wind_ROS_SA = 0._r8
+                end if
```
slevisconsulting, line 1224: @jkshuman running this by you:
I was getting `phi_wind_ROS_SA = infinity` due to `ir = 0`, which was due to `moist_damp = 0`. I added this if statement to confirm that `ir > 0 .and. xi > 0`.
**Question 1:** Did I make the right choice for the `else`?

This change caused `canopy_frac_burnt < 0` a few lines down due to `ROS_SA = 0`. So I added `max(0...` to the calculation of `canopy_frac_burnt`.
**Question 2:** Does that seem reasonable?

I have now completed a 5 yr simulation at the CZ2 site. I will share some figures here on github to initiate discussion.

slevisconsulting: Just saw your message about other updates @jkshuman 
I will try those out before posting figures. Thanks.

slevisconsulting: @jkshuman 
in the context of your latest modifications, I'm keeping the if statement and the max(0... that I talked about above, because they still seem relevant when moist_damp = 0 and ROS_SA = 0.

slevisconsulting:
I have confirmed my previous statement. More precisely:
`moist_damp` for the crown doesn't exceed 0 and, as a result, `ROS_active` doesn't exceed 0 in the 5-yr run.
The latter means no active crown fire in the 5-yr run.
The former means also no passive crown fire in the 5-yr run.

Sample values from year 2 (see plots above) during the fire season:

**Surface fuel**
| moist_damp | mw_weight | fuel_eff_moist | fuel_mef |
0.55 0.48 0.25 0.52

**Crown fuel**
| moist_damp | mw_weight | fuel_eff_moist | fuel_mef |
0 1.92 0.48 0.25

where `mw_weight = fuel_eff_moist / fuel_mef`
and `moist_damp > 0` for `mw_weight < 1`

@jkshuman I'm sharing all this in case something stands out as strange. To me things look reasonable for now.

fire/SFMainMod.F90
slevisconsulting: This post can be ignored for now because ROS_torch is not used.

fire/SFMainMod.F90
jkshuman: @slevisconsulting normally I would say yes, sounds good, but there must be something more going on in here. you are in this section because the patchFI >= passive_crown_FI indicating there is a fire. And then we are calculating the pieces to get the reaction intensity "ir" for the active crown fire using a given fuel loading for the FM 10 model.  It would seem that I have a mistake in this section prior to calculating "ir". I am looking it over now. I belive it is a conversion error. Let me make an adjustment.

jkshuman:
@slevisconsulting  I pushed a set of conversions for the FM10 fuels to your branch. Can you test again to see if you still get ir = 0 due to moist damp = 0?

fire/SFMainMod.F90
```diff
@@ -835,6 +1034,275 @@ subroutine  area_burnt_intensity ( currentSite, bc_in )
...
+
+             ! ir = reaction intenisty in kJ/m2/min
+             ! sum_fuel as kgBiomass/m2 for ir calculation
+             ir = reaction_v_opt*(net_fuel)*SF_val_fuel_energy*moist_damp*SF_val_miner_damp  
+ 
+             ! actual ROS (m/min) for FM 10 fuels for open windspeed, Eq 8 Scott & Reinhardt 2001
+             ROS_active = 3.34_r8 * ((ir*xi*(1.0_r8+phi_wind)) / (fuel_bd * eps * q_ig))
+
+             ! critical min rate of spread (m/min) for active crowning
+             ROS_active_min = (critical_mass_flow_rate / fuel_bd) * 60.0_r8
```
pollybuotte, line 1213: @jkshuman is this using the correct equation?  To me it looks like it's supposed to be calculating the critical minimum rate of spread for active crowning, which is Eq 14 in Scott & Reinhardt.  But this line corresponds to  Eq 12, which is for calculating the rate of spread for initiation.

pollybuotte: Sorry @jkshuman I see I was getting variable names mixed up and that ROS_active_min is the Scott and Reinhardt R'initiation.  You can ignore my previous comment.

slevisconsulting:
> Can you test again to see if you still get ir = 0 due to moist damp = 0?

Good news @jkshuman 
I now see `canopy_frac_burnt > 0` though always < 1 (in the two yrs of simulation). This means that we do not cross our active crown fire threshold in these two yrs, which seems fine, do you agree?

jkshuman:
@slevisconsulting that's great news! so now the race to active crown fire begins. 
With these where canopy_frac_burnt > 0 but < 1 there is some canopy burnt, and it alters the FI. Check Patch%FI before canopy burning, FI_final which updates Patch%FI after adjusting for potential canopy fuel consumed. A similar update should happen to patch%ROS_front, and that should be checked as well to confirm it actually changes with canopy consumption.

fire/SFMainMod.F90
```diff
@@ -835,6 +1034,284 @@ subroutine  area_burnt_intensity ( currentSite, bc_in )
...
+             else  
+                currentPatch%frac_burnt = 0.0_r8
+             endif ! lb
+
+             !final fireline intensity (kJ/m/sec or kW/m), Eq 22 Scott & Reinhardt 2001
+             FI_final = (currentPatch%heat_per_area + &
+  currentPatch%canopy_fuel_load * canopy_ignite_energy * canopy_frac_burnt) * &
+  currentPatch%ROS_front / 60._r8
+             ! update patch FI to adjust according to potential canopy fuel consumed (passive and active)
+             currentPatch%FI = FI_final
```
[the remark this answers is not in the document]
jkshuman: @slevisconsulting sounds good.

slevisconsulting:
@jkshuman anything additional that you recommend I check in my quick 2-yr tests at CZ2? Otherwise I may submit a multi-yr simulation to see if conditions cross the active crown fire threshold at some point, i.e.:
`ROS_active >= ROS_active_min`

slevisconsulting:
Two 20-yr simulations ran on **izumi**, one with `fates_fire_active_crown_fire = 1` and the other with `fates_fire_active_crown_fire = 0`. See history output here:
- /scratch/cluster/slevis/archive/CZ2_acf_on/lnd/hist
- /scratch/cluster/slevis/archive/CZ2_acf_off/lnd/hist

Case directories here (always on izumi):
- /home/slevis/cases_FATES/CZ2_acf_on
- /home/slevis/cases_FATES/CZ2_acf_off

The first simulation (CZ2_acf_on) satisfies the active crown fire requirement at times, so the outcomes from the two simulations differ. I am happy to perform additional testing or to hand over to @jkshuman and @pollybuotte .

jkshuman:
Thanks for your work on this @slevisconsulting Will keep you posted if there are other things that come up.

glemieux:
Superseded by https://github.com/NGEET/fates/pull/1362 per @adrifoster

PR #800, opened 2021-10-30
Carbon allocation: bug fixes and simpler (optional) allocation approach

mpaiao:
This pull request addresses the bug discussed in #784. I changed the order of the carbon allocation fraction, to ensure that allocation to fine roots and leaves were truly proportional to demand.

### Description:

1. Bug fix in DailyPRTAllometricCarbon (parteh/PRTAllometricCarbonMod.F90). When allocating to different tissues, the code was subtracting allocation of tissues before calculating the amount for the next tissue, potentially under-allocating carbon to fine roots.
2. Renamed variable laimemory with leafmemory. We were never tracking the LAI, but leaf carbon, and thus the name is dangerously misleading.
 3. Implemented an option carbon allocation routine (DailyPRTAllometricCarbonSimpler in parteh/PRTAllometricCarbonMod.F90).  In this routine, I applied much simpler rules for maintenance allocation: it simply checks how much each pool is in deficit, and allocates carbon (storage + carbon_balance) to the pools according to the debt. If there is any carbon left, then the plant finishes filling the storage pool and then allocates to growth.  This option is set by `fates_parteh_mode=0`, and will require changing the valid options in the host land models (see example of changes for CTSM in [here](https://github.com/mpaiao/CTSM/commit/2ceec76d538af7347af3b6ad014d66c1635e4255)).

This addresses issue #784

### Collaborators:

@rgknox @jkshuman 

### Expectation of Answer Changes:

This fixes the carbon allocation priorities, so it should cause higher use of carbon balance and storage than the current master.

### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the in-code documentation .AND. (the [technical note](https://github.com/NGEET/fates-docs) .OR. the wiki) accordingly.
- [ ] I have read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/master/CONTRIBUTING.md) document.
- [ ] FATES PASS/FAIL regression tests were run
- [ ] If answers were expected to change, evaluation was performed and provided

### Test Results:

CTSM (or) E3SM (specify which) test hash-tag:

CTSM (or) E3SM (specify which) baseline hash-tag:

FATES baseline hash-tag:

Test Output:

Issue #784, opened 2021-09-29
Potentially incorrect allocation to fine roots in PRTAllometricCarbonMod.F90?


mpaiao:
I was looking at this code and it seems to me that allocation to fine roots would be underestimated:

https://github.com/NGEET/fates/blob/70a3e133244877f325652f2cf40e637d785b3aa7/parteh/PRTAllometricCarbonMod.F90#L543-L555

Carbon balance is subtracted after finding leaf_c_flux, and then scaled by fnrt_c_demand/total_c_demand, so the relative demand of fine root demand is applied in a lower carbon balance value. Shouldn't the code be something like the one below?
```fortran
       leaf_c_flux    = min(leaf_c_demand, &
                        carbon_balance*(leaf_c_demand/total_c_demand))
       fnrt_c_flux    = min(fnrt_c_demand, &
                            carbon_balance*(fnrt_c_demand/total_c_demand))
       leaf_c(iexp_leaf) = leaf_c(iexp_leaf) + leaf_c_flux
       fnrt_c         = fnrt_c + fnrt_c_flux
       carbon_balance = carbon_balance - leaf_c_flux - fnrt_c_flux
```

mpaiao:
And I think the same rationale applies to the block a few lines above.

https://github.com/NGEET/fates/blob/70a3e133244877f325652f2cf40e637d785b3aa7/parteh/PRTAllometricCarbonMod.F90#L485-L506

rgknox:
I agree @mpaiao

rgknox:
yeah, we should not be updating the carbon balance until both pools have calculated their fluxes

jkshuman:
I also agree with you @mpaiao good catch.

rgknox:
I like simple, seems fine to have the alternative.
In the N & P enabled allocation scheme, we allow the user to define the order in which replace tissues.  We could potentially merge our ideas and create a c-only version:
https://github.com/NGEET/fates/blob/master/parteh/PRTAllometricCNPMod.F90#L615

glemieux:
Closing per https://github.com/NGEET/fates/pull/863

PR #863, opened 2022-05-04
Carbon allocation bug fix

mpaiao:
Revised the carbon allocation routine to ensure allocation to fine roots and leaves are truly proportional to demand.

### Description:

1. Bug fix in DailyPRTAllometricCarbon (parteh/PRTAllometricCarbonMod.F90). When allocating to different tissues, the code was subtracting allocation of tissues before calculating the amount for the next tissue, potentially under-allocating carbon to fine roots.
2. Minor code updates to simplify allocation calculations. It now uses a single
allocation factor based on availability and total need, which is applied to all tissues.
3. Replaced a few if statements with select case, which simplifies adding other hypotheses
in the future (and it's safer for selecting cases).

This pull request addresses the bug discussed in #784 and supersedes pull request #800.
Additional minor changes in former PR #800 will be added as subsequent pull requests.

### Collaborators:

@rgknox @glemieux @jkshuman 

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


rgknox (approved):

glemieux:
Regression testing against baseline `fates-sci.1.57.2_api.23.0.0-ctsm5.1.dev098` is complete.  There are no unexpected failures.  The satellite phenology and static stand structure test mods all pass B4B which is expected.  DIFFs appear to be as expected with initial and daily small differences in carbon balance and related variables.  Interestingly the PRT2 testmod passes B4B, but I'm not sure why this would be so.  @rgknox do you have thoughts here?

Results can be found on Cheyenne at: 
`/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr863-sci_1_57_2`

rgknox:
That makes sense, the PRT2 has a completely different allocation module, only the carbon-only code was changed.  I believe the PRT2 module does not have the bug that @mpaiao is fixing, so that should be fine

Issue #490, opened 2019-02-25
Make disturbance rates happen in parallel


ckoven:
A thing that's been on our to-do list for a while is to improve the disturbance logic so that it doesn't have to decide which is the dominant type of disturbance in a given timestep.  E.g. see here: https://github.com/NGEET/fates/issues/266#issuecomment-326411777 and subsequent comments and also point 5 in https://github.com/NGEET/fates/issues/450#issue-392327274 for discussion on proposed fix.  Marking this as an issue here to hopefully come back to this and try to fix it soon.  It seems to me like the simplest solution might be to handle the overlap and possible disturbed area > 100% problems in the subroutine disturbance_rates and then loop over disturbance types instead of picking the largest in subroutine spawn_patches?

rosiealice:
Just thinking about this a bit, and adding my voice to the idea from the perspective of fire, where we can in principle have low level fires every day that wouldn't result in any mortaliy or disturbance if they were smaller than the physiological mortality, abnd vice-versa... Then if we add  low level  logging in too, we might end up substantially under estimating disturbance and turnover rates, i think the idea of dealing with overlap in disturbance rates should work fine. 

Do we assume that disturbances are non-overlapping until they are > 100% in total, or do we impose some kind of conditional probability thing? (assuming, for example, that logger might avoid forests that are currently on fire...)

rgknox:
If we assume that the disturbances are uniformly distributed in space, then the fraction of ground Fj experienced jointly by both of two disturbances with area fractions F1 and F2 would simply be their multiplier, right?   Fj = F1 * F2,  Ftotal = Fj + (F1-Fj) + (F2-Fj) And for three different entities.. I feel like we could figure this out be drawing out a cube with overlapping fractions in the 3 dimensions, but there would be three regions where 2 types overlap, and 1 region where 3 types overlap.

rosiealice:
I guess we could write some algorithm for that type of logic. Would perhaps
not be trivially extensible to n dimensional disturbance types, but there's
probably code that does that on the internet somewhere though.

ckoven:
[...]

PR #873, opened 2022-05-24
Don't consider grass mortality as treefall disturbance

ckoven:
Currently FATES tracks a fraction of all canopy mortality rates as patch-generating disturbance, irrespective of what type of plant is dying.  I..e is grasses are in the canopy because of savanna or grassland conditions, then when the grasses die, if the parameter `fates_mortality_disturbance_fraction` is nonzero, then their crown area will be assigned to a new patch. This doesn't seem right, particularly in that it competes against fire as a disturbance process (because #490), and fire should be the more relevant type of disturbance in grass-dominated ecosystems.

So this is a small code change to make the logic so that when grasses die due to things like drought or carbon starvation or background mortality, then that isn't considered as patch-generating disturbance.  It does still consider fire and fire mortality as patch-generating disturbance.

This will change answers anywhere where grasses are in the canopy.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

I've run this and it does what I expected, which is to much better align total disturbance rates with burn rates in grass-dominated systems.


rosiealice:
Seems reasonable to me to make this change

rgknox (approved):

jkshuman:
@ckoven I cherrypicked this commit to test in an older version of code used for my SA fire work. There were no differences to note for a 100 year run with active fire and three PFTs (two tropical trees, and grass). I evaluated veg dynamics, mortality and fire patterns and behavior across SA for the final ten years of simulation.

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -275,18 +276,54 @@ subroutine Mortality_Derivative( currentSite, currentCohort, bc_in, frac_site_pr
...
+       ! and part in disturbance routine.
+
+       if ( .not. ExemptTreefallDist(currentCohort)) then
+          currentCohort%dndt= -(1.0_r8-fates_mortality_disturbance_fraction) &
+               * (cmort+hmort+bmort+frmort+smort+asmort) * &
+               currentCohort%n
+       else
+          currentCohort%dndt= - (cmort+hmort+bmort+frmort+smort+asmort) * &
+               currentCohort%n
+       endif
```
glemieux, line 293: Noticed a super minor white space "issue" and thought this might be good to reduce duplication:
```suggestion
       currentCohort%dndt= -(cmort+hmort+bmort+frmort+smort+asmort) * currentCohort%n
       if ( .not. ExemptTreefallDist(currentCohort)) then
          currentCohort%dndt = (1.0_r8-fates_mortality_disturbance_fraction) * currentCohort%dndt
       endif
```

glemieux (approved):
Minor suggestion

glemieux:
Regression tests show differences across all test mods with the exception of satellite phenology tests and static stand structure which are B4B.  The DIFFs include expected variations in the disturbance history output variables. 

File location on Cheyenne: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr873`

jkshuman:
Testing with these latest updates show a difference to earlier results that were missing the recent commit details. These are for 100 years across South America using 3 PFTs (two trees and a C4grass). The earlier results are for a 0.5 degree run and this recent test is 1 degree, but they share the same parameter file and code base outside of these specific grass related commits. (I also have results for the step-wise change between these commits). The large patterns are the same, but the extent of grass decreases, as do associated fire intensities. 

These broad scale tests do not consider the implications to the loss of heterogeneity in grass patches that do not have fire. In discussion today there was discussion about the possibility of a "persistent grass" patch for a patch of grass in an area that does not have fire when the only way to generate a new patch is through fire. Grasses and other non-tree vegetation are an important part of the successional trajectory, and I am not sure how this change to patch-level dynamics will play out in a system without fire. Simulation of these non-tree portions of succession have little to no representation in most forest successional models. ED2 recently added lianas, but they are a vine that exist mainly on trees. Just trying to think about implications in systems outside of the fiery ones that I think about.

There is a recent paper out regarding more explicit modeling of grasses that is worth consideration as improvements to handling of grasses in simulation moves forward: [https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0236546#sec014](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0236546#sec014)

 old method:  has slightly more grass (PFT 3)

download-13
 new method - only fire creates a new patch for grasses: slightly less grass (PFT 3)

download-14

old method: has higher fire intensities in grass region

download-2

new method: has lower fire intensities extending into forest
download

ckoven:
Thanks @jkshuman for doing that analysis.  Do you know if there is any appreciable difference to the patch age distribution in the grass-dominated gridcells in your experiment?

jkshuman:
I attach a few figures here. Is this what you were  asking about? From this it seems the distribution of grasses is the primary change. (For area of grass distribution see earlier figures as these are not restricted in area.)

old way: 
patch area by age
download-1

new method only fire creates new patches when grass in canopy:
patch area by age
download

old way:
canopy cover by patch age mean final 10 years
download-3

new method only fire creates new patches when grass in canopy:
canopy cover by patch age mean final 10 years
download-2

old way:
fire intensity by patch age weighted by patch area by patch age
download-7

new method of only fire creates new patch when grass in canopy:
fire intensity by patch age weighted by patch area by patch age

download-6

ckoven:
Great, thanks @jkshuman! Indeed those changes all look pretty subtle.

jkshuman:
@ckoven thanks for the further discussion on this change. I say we incorporate it and keep an eye on grass systems that do not have fire for any odd behavior.

ckoven:
Thanks @jkshuman!

PR #875, opened 2022-05-25
parallel disturbance

### LINKED ISSUE #490: Make disturbance rates happen in parallel

ckoven:
We've had an issue for ages that on a given day, FATES only resolves the larger of disturbance types between tree fall, fire, and logging.  This code changes that so that all three are resolved every day.   The solution is to just loop over disturbance types rather than evaluate which is largest.

fixes #490.

This should change answers.  In particular, the sum of all disturbance rates should now equal the `FATES_POTENTIAL_DISTURBANCE` rate.  

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -515,6 +440,8 @@ subroutine spawn_patches( currentSite, bc_in)
     ! If nocomp is not enabled, then this is not much of a loop, it only passes through once.
     nocomp_pft_loop: do i_nocomp_pft = min_nocomp_pft,max_nocomp_pft
 
+    disturbance_type_loop: do i_disturbance_type = 1,N_DIST_TYPES
+
     ! calculate area of disturbed land, in this timestep, by summing contributions from each existing patch.
```
rgknox, line 445: lets get an indent in here, like the loop label

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1118,8 +1041,18 @@ subroutine spawn_patches( currentSite, bc_in)
...
              currentPatch%area = currentPatch%area - patch_site_areadis
 
+             ! for all disturbance rates that haven't been resolved yet, increase their amount so that 
+             ! they are the same amount of gridcell-scale disturbance relative to the original patch size
+             if (i_disturbance_type .ne. N_DIST_TYPES) then
+                do i_dist2 = i_disturbance_type+1,N_DIST_TYPES
+                   currentPatch%disturbance_rates(i_dist2) = currentPatch%disturbance_rates(i_dist2) &
+                        * oldarea / currentPatch%area
+                end do
+             end if
```
rgknox, line 1054: Following through the logic here with a generic example, and I think it checks out:

lets assume that the patch starts out with an area of 1., and all three types apply 0.333 disturbance rates.

After the first disturbance, we have an area of 0.666, and the remaining two rates are bumped up to 0.333*(1/0.666) = 0.5.

After the second disturbance, we have an area of 0.33 = 0.66 * 0.5, and the remaining disturbance rate is bumped up to 0.5*(0.66/0.33) = 1.0  

The last disturbance will cover the entire remaining area fraction.

rgknox (approved):
I think the changes look great.  Only request is that we apply the indent to the big new loop.  Also, this change is very central to how FATES operates. I think it would be good if we run a couple of multi-decade sanity check simulations on the f10 grid and compare to base (with fire turned on) just to see how things compare and if the results are sensible.  Alternatively, or additionally, we could consider a regional simulation with a higher resolution grid, which may capture a few events with 100% disturbance in high fire areas.  But I'm not sure if that is necessary.

biogeochem/EDPatchDynamicsMod.F90
ckoven: OK I did a big auto-indent for this subroutine.

glemieux:
I started up a 50 year f10 run on Cheyenne and ran into a patch fusion error around the 12 year mark.  I haven't looked into it deeply yet, but wanted to report it asap:
```
34: FATES fuse_patches(): currentPatch is not associated?
34: ERROR in EDPatchDynamicsMod.F90 at line 2335
...
34: ERROR: Unknown error submitted to shr_abort_abort.
34:Image              PC                Routine            Line        Source
34:cesm.exe           00000000013867D6  Unknown               Unknown  Unknown
34:cesm.exe           0000000000F07BE0  shr_abort_mod_mp_         114  shr_abort_mod.F90
34:cesm.exe           00000000008E0380  fatesglobals_mp_f          94  FatesGlobals.F90
34:cesm.exe           0000000000899E9E  edpatchdynamicsmo        2335  EDPatchDynamicsMod.F90
34:cesm.exe           0000000000887B87  edmainmod_mp_ed_e         276  EDMainMod.F90
34:cesm.exe           00000000005F3883  clmfatesinterface        1025  clmfates_interfaceMod.F90
34:cesm.exe           00000000005CC59D  clm_driver_mp_clm        1136  clm_driver.F90
34:cesm.exe           0000000000577A72  lnd_comp_nuopc_mp         886  lnd_comp_nuopc.F90
```

The case directory is here: 
```
/glade/u/home/glemieux/scratch/ctsm-cases/pr875-longterm.fates-sci.1.57.4_api.24.0.0-8-gbd994a1c-ctsm5.1.dev099-C25a7cd3f2-Fbd994a1c.intel
```

rgknox:
The reason it is changing the pointer of currentPatch is because it will cease to exist during fuse_2_patches, see these lines: https://github.com/NGEET/fates/blob/bd994a1c548d9c13e1bcc9a40657de2f670515c4/biogeochem/EDPatchDynamicsMod.F90#L2666-L2679

I think we just need to add some more complete logic where we set tmpptr prior to the fusion call.

if(associated(currentPatch%older))then
       tmpptr => currentPatch%older
else
       tmptr => currentPatch%younger
end if

If the second condition (ie younger) yields a null pointer, then you only have 1 patch. If you only have 1 patch, than you don't need to be fusing, so that shouldn't happen.  Anyone else see this the same way?

glemieux:
Per the fates software discussion today, it was agreed that I should implement Ryan's suggested fix and retest on this PR branch even though the issue is not directly caused by these changes.

glemieux:
Prior to running the original long-term case, I also ran regression tests against the baseline and all expected tests pass, with three non-b4b results:
- ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdDefLandUse
- ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdDefReducedComplexNoComp
- ERS_Ld60.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdDefLogging

I believe these make sense as two of the three add in explicit disturbance modes to compete with fire and treefall and thus would be more likely to exhibit differences.  Differences just due to fire and treefall disturbance are not likely to show up except in `NoComp` mode, where the isolation of pfts tends to exhibit certain dynamic updates earlier (compared to fully dynamic competition).

File location on Cheyenne: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr875-dev099`

rgknox:
Another solution to this issue, which might be a little better, would be to just remove the endun:

https://github.com/NGEET/fates/blob/bd994a1c548d9c13e1bcc9a40657de2f670515c4/biogeochem/EDPatchDynamicsMod.F90#L2333-L2336

If the currentPatch loop is not associated, it would be because the we just fused the oldest patch and are now pointing to null(), which is an expected outcome.  Following this we check for association, so the code should behave.

I like this better than the previous idea because we would not be attempting fusion on the second oldest patch again. By attempting fusion on the second oldest, it gets two opportunities to fuse, which is more than the other patches and is not equitable. If more fusion is required, we should restart the loops. Which will be achieved by just removing the endrun listed above. 

Another option would be to have the result of the fusion live in the currentPatch, instead of the tpp patch.  That way, if the tpp patch was at fused at its oldest position, the pointer would be moved to the older (null) position, which will just break out of the loop.  This would seem to be slightly more efficient, but would require some double checking and care that I'm not overlooking something here ...

rgknox:
https://github.com/rgknox/fates/commit/b341df53c494b1538509a8dd8aff891ef6425b6a

glemieux:
Regression tests re-run with the same results [as before](https://github.com/NGEET/fates/pull/875#issuecomment-1180966695).

Folder location on cheyenne: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr875-remove_endrun`

PR #887, opened 2022-07-29
Remove memory and lai variables

### LINKED ISSUE #541: remove cohort variables "lai" and "laimemory"
The cohort variables lai and laimemory have become obsolete.  LAI memory used to hold the LAI on deciduous trees as their leaves fell, however with the new allometry routines, we don't really need to remember this (I think?).  My understanding is that we can re-flush leaves based on the target leaf mass for the size of the plant, and that laimemory does not provide any special information.  @rosiealice, can you confirm this?

The cohort variable "lai" is barely used, and can be derived from cohort variable treelai, crown area and the patch area. It looks like its just taking up memory.

rgknox:
### Description:

The memory (laimemory, sapwmemory and structmemory) and lai terms are unecessary.  We don't need the memory terms because we can use the allometry functions to calculate targets.  We don't need the cohort level lai and sai variables because we can derive them from treelai, treesai, crown area and canopy area.  

Fixes #541 

I also removed the smooth leaf distribution code from the canopy vertical profile routine because that is old and unmaintained and is triggering lots of compiler warnings.

I expect small but subtle differences in results due to the laimemory stuff.  The flushing target was previously laimemory, which was the target when the leaves dropped.  Now the target is just re-caclulated when the leaves are ready to flush.  Its possible that a trimming event would had happened in between, or perhaps due to canopy promotion/demotion/fusion there was a small change in dbh to maintain mass conservation, which would also trigger a difference in the target.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -1493,7 +1493,6 @@ contains
     real(r8) :: fraction_exposed         ! how much of this layer is not covered by snow?
     real(r8) :: layer_top_hite           ! notional top height of this canopy layer (m)
     real(r8) :: layer_bottom_hite        ! notional bottom height of this canopy layer (m)
-    integer  :: smooth_leaf_distribution ! is the leaf distribution this option (1) or not (0)
     real(r8) :: frac_canopy(N_HITE_BINS) ! amount of canopy in each height class
     real(r8) :: minh(N_HITE_BINS)        ! minimum height in height class (m)
     real(r8) :: maxh(N_HITE_BINS)        ! maximum height in height class (m)
@@ -1504,7 +1503,6 @@ contains
 
     !----------------------------------------------------------------------
 
-    smooth_leaf_distribution = 0
 
     ! Here we are trying to generate a profile of leaf area, indexed by 'z' and by pft
     ! We assume that each point in the canopy recieved the light attenuated by the average
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -1449,8 +1449,6 @@ contains
     !
     ! currentCohort%treelai    ! LAI per unit crown area  (m2/m2)
     ! currentCohort%treesai    ! SAI per unit crown area  (m2/m2)
-    ! currentCohort%lai        ! LAI per unit canopy area (m2/m2)
-    ! currentCohort%sai        ! SAI per unit canopy area (m2/m2)
     ! currentCohort%NV         ! The number of discrete vegetation
     !                          ! layers needed to describe this crown
     !
@@ -1477,7 +1475,7 @@ contains
     ! !USES:
 
     use EDtypesMod           , only : area, dinc_vai, dlower_vai, hitemax, n_hite_bins
-  
+
     !
     ! !ARGUMENTS
     type(ed_site_type)     , intent(inout) :: currentSite
@@ -1497,7 +1495,6 @@ contains
     real(r8) :: layer_bottom_hite        ! notional bottom height of this canopy layer (m)
     integer  :: smooth_leaf_distribution ! is the leaf distribution this option (1) or not (0)
     real(r8) :: frac_canopy(N_HITE_BINS) ! amount of canopy in each height class
-    real(r8) :: patch_lai                ! LAI summed over the patch in m2/m2 of canopy area
     real(r8) :: minh(N_HITE_BINS)        ! minimum height in height class (m)
     real(r8) :: maxh(N_HITE_BINS)        ! maximum height in height class (m)
     real(r8) :: dh                       ! vertical detph of height class (m)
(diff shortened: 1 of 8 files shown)


rosiealice:
This is great @rgknox. Such a simple solution! I'm struggling to remember why it didn't work like this in the first place. How does it work with crown damage? I guess we just change the target?

rgknox:
right, crown damage level is remembered, and that instructs the allometry target

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1541,307 +1537,223 @@ subroutine leaf_area_profile( currentSite )
...
+             cl = currentCohort%canopy_layer
 
-                currentCohort => currentCohort%taller
+             ! ----------------------------------------------------------------
+             ! How much of each tree is stem area index? Assuming that there is
+             ! This may indeed be zero if there is a sensecent grass
+             ! ----------------------------------------------------------------
 
-             enddo !currentCohort
+             if( (currentCohort%treelai+currentCohort%treesai) > 0._r8)then
```
mpaiao, line 1558: To be consistent with other places in the code, should we use variable "nearzero" instead of 0._r8?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -1291,17 +1266,11 @@ subroutine phenology_leafonoff(currentSite)
...
-                   currentCohort%leafmemory   = leaf_c
-
                    call PRTDeciduousTurnover(currentCohort%prt,ipft, &
                         leaf_organ, leaf_drop_fraction)
 
                    if(prt_params%woody(ipft).ne.itrue)then
 
-                      currentCohort%sapwmemory   = sapw_c * stem_drop_fraction
-                      currentCohort%structmemory   = struct_c * stem_drop_fraction
-
```
mpaiao, line 1304: Another good reason to drop the memory variables: leafmemory was the prior leaf biomass, whereas sapwmemory/structmemory were the amount of woody biomass lost. I actually had changed the definition of sapwmemory/structmemory in #801 to the same as leafmemory because I found the original definitions very confusing...

main/EDPftvarcon.F90
```diff
@@ -1690,6 +1690,15 @@ subroutine FatesCheckParams(is_master)
...
 
+        if( (prt_params%woody(ipft) == itrue) .and. &
+            (EDPftvarcon_inst%phen_stem_drop_fraction(ipft) > nearzero ) ) then
+           write(fates_log(),*) ' Non-zero stem-drop fractions are not allowed for woody plants'
+           write(fates_log(),*) ' PFT#: ',ipft
+           write(fates_log(),*) ' part_params%woody:',prt_params%woody(ipft)
+           write(fates_log(),*) ' phen_stem_drop_fraction: ', EDPFtvarcon_inst%phen_stem_drop_fraction(ipft)
+           write(fates_log(),*) ' Aborting'
+           call endrun(msg=errMsg(sourcefile, __LINE__))
+        end if
```
mpaiao, line 1701: Good idea to prevent the model to run if this combination is not acceptable.

mpaiao (changes requested):
@rgknox I checked the changes and they look all good to me, I just made one really minor suggestion in one "if" block to make the code a bit less likely to run into floating point exceptions.

biogeochem/EDPhysiologyMod.F90
rgknox: i saw that too, not sure if this will come up on tests, but glad its correct now

biogeochem/EDCanopyStructureMod.F90
rgknox: this has been changed to use nearzero

mpaiao (approved):

glemieux:
Testing on Cheyenne is resulting in DIFFs across all tests expect the clm50 non-fates test (expected) and the `1x1_brazil` satellite phenology tests.  Looking at the other satellite phenology test mods with DIFFs, I'm seeing RMS values in the second time step for respiratory variables, `FATES_LBLAYER_COND`, the `GPP`, `NPP`, and `NEP` variables, and `FATES_STOMATAL_COND`.  I'm not sure if this is expected for these changes.  @rgknox thoughts?

Test folder location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr887`

rgknox:
yes, there should be diffs,  some of it is related to what @mpaiao noted, in how we previously defined memory for non-leaf pools, and we were replacing these pools based on the total, and not the remainder (where the latter is correct)

glemieux:
Per the discussion at the fates software meeting last week, @rgknox is going to assess splitting this PR into two separate issues: one to remove the `laimemory` terms, and the other to remove the cohort level `lai` and `sai` terms.  The goal in doing so would be to isolate what changes are causing non-b4b results in the SatPhen tests (recommended by @mpaiao).

rgknox:
Updated tests with the changes to fleaf.

/glade/scratch/rgknox/tests_0913-100825ch

Issue #884, opened 2022-07-26
Grass PFT'


ZacharyRobbins:
Currently grass PFT's have the same assignment as all cohorts with regards to allocation of coarse woody debris (and thus fuels). This ultimately results in grass producing 1000 hour fuels, 100 hour fuels and other nonsense. Ultimately the best fix for this would be to make CWD assignment a PFT parameter so each PFT can assign to different pools based on the fuel it would produce. However in the short term my fix has been to say if PFT is not woody then all the normal CWD pools go to the dead leaves pool.  This works in the case of grass, however, would not work with grass, chaparral and a tree PFT. Below I illustrate the fuel output for a grass PFT, as you can see it creates trunk fuels (based on fuel classification) which don't map well with how spitfire works. 
![image](https://user-images.githubusercontent.com/25291127/181098119-da875e4a-ca13-4633-b809-9e5d8ad4bab1.png)
@ckoven

ckoven:
Thanks @ZacharyRobbins for identifying this!

rgknox:
@ZacharyRobbins could you point to the location in the code where you see grasses contributing to CWD please?

ckoven:
@rgknox I think the issue is here: https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L2248-L2272. The sapwood gets routed to CWD pools when it turns over, whether it is from woody or nonwoody plants.  Since grasses can still have "sapwood" (i.e. live stem biomass) in their allometry, that biomass get sent to the CWD pools.  And since FATES uses the same partitioning to fuel classes for all PFTs / plant sizes, that is generating small but nonzero amounts of coarse fuel.

jkshuman:
Agreed. Thank you for identifying this @ZacharyRobbins and a fix for the problem!

glemieux:
Closed per #891

PR #891, opened 2022-08-16
Grass PFT only creates leaves not cwd

ZacharyRobbins:
In response to issue #884

Modified EDCohortDynamicsMod.F90 (@ln970): if the pft is grass then all of its organs are treated as leaves for calculation of coarse woody debris(cwd).
EDPatchDynamicsMod.F90 (@ln 1641):  When patches spawn, only tree pft donate dead tress or cwd, and mass to the new patch, if pft is grass it assumed that grass was consumed in fire. This may not resolve fluxes to the atmosphere accurately.
EDPhysiologyMod.F90 (@ln 2199):  When other litter fluxes occur, if pft is grass, then it only turnsover  or contributes leaves not coarse woody debris.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeochem/CMIP data.txt
+++ b/biogeochem/CMIP data.txt
@@ -1,0 +1,24 @@
+. 	
+CMIP6.CMIP.NCAR.CESM2.historical.r11i1p1f1.day.rlds.gn'
+
+
+
+CF Standard Name | Unit | AMIP
+
+---------------- | ---- | ----
+
+specific_humidity | 1 | hus
+
+precipitation_flux | kg m-2 s-1 | pr
+
+surface_air_pressure | Pa | ps 
+
+surface_downwelling_longwave_flux_in_air | W/m2 | rlds
+
+surface_downwelling_shortwave_flux_in_air | W/m2 | rsds
+
+air_temperature | K | ta
+
+eastward_wind | ms-1 | ua
+
+northward_wind | ms-1 | va
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -1864,6 +1864,7 @@ contains
              sapw_m          = currentCohort%prt%GetState(sapw_organ,element_id)
              struct_m        = currentCohort%prt%GetState(struct_organ,element_id)
           else
+             ! for non-woody plants all stem fluxes go into the same leaf litter pool
              leaf_m          = currentCohort%prt%GetState(leaf_organ,element_id) + &
                      currentCohort%prt%GetState(sapw_organ,element_id) + &
                      currentCohort%prt%GetState(struct_organ,element_id)
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -1661,8 +1661,7 @@ contains
              repro_m  = currentCohort%prt%GetState(repro_organ, element_id)
           
              if (prt_params%woody(currentCohort%pft) == itrue) then
-                ! Assumption: for woody plants, we lump fluxes from deadwood and sapwood together in CWD pool.
-                ! for non-woody plants, we put all stem fluxes into the same leaf litter pool.
+                ! Assumption: for woody plants fluxes from deadwood and sapwood go together in CWD pool
                 leaf_m          = currentCohort%prt%GetState(leaf_organ,element_id)
                 sapw_m          = currentCohort%prt%GetState(sapw_organ,element_id)
                 struct_m        = currentCohort%prt%GetState(struct_organ,element_id)
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -859,7 +859,7 @@ contains
                             do el = 1,num_elements
 
                                leaf_m = nc%prt%GetState(leaf_organ, element_list(el))
-                               ! for grasses burn all aboveground tissues, for woody plants burn only leaves
+                               ! for woody plants burn only leaves
                                if(int(prt_params%woody(currentCohort%pft)) == itrue)then
 
                                   leaf_m = nc%prt%GetState(leaf_organ, element_list(el))
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -865,7 +865,7 @@ contains
                                   leaf_m = nc%prt%GetState(leaf_organ, element_list(el))
 
                                else
-
+                               ! for grasses burn all aboveground tissues
                                   leaf_m = nc%prt%GetState(leaf_organ, element_list(el)) + &
                                        nc%prt%GetState(sapw_organ, element_list(el)) + &
                                        nc%prt%GetState(struct_organ, element_list(el))
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -1859,8 +1859,7 @@ contains
           repro_m  = currentCohort%prt%GetState(repro_organ, element_id)
 
           if (prt_params%woody(currentCohort%pft) == itrue) then
-             ! Assumption: for woody plants, we lump fluxes from deadwood and sapwood together in CWD pool.
-             ! for non-woody plants, we put all stem fluxes into the same leaf litter pool.
+             ! Assumption: for woody plants fluxes from deadwood and sapwood go together in CWD pool
              leaf_m          = currentCohort%prt%GetState(leaf_organ,element_id)
              sapw_m          = currentCohort%prt%GetState(sapw_organ,element_id)
              struct_m        = currentCohort%prt%GetState(struct_organ,element_id)
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -1637,13 +1637,17 @@ contains
              ! across ground surface. 
              ! -----------------------------------------------------------------------
              
-             sapw_m   = currentCohort%prt%GetState(sapw_organ, element_id)
-             struct_m = currentCohort%prt%GetState(struct_organ, element_id)
-             leaf_m   = currentCohort%prt%GetState(leaf_organ, element_id)
-             fnrt_m   = currentCohort%prt%GetState(fnrt_organ, element_id)
-             store_m  = currentCohort%prt%GetState(store_organ, element_id)
-             repro_m  = currentCohort%prt%GetState(repro_organ, element_id)
-             
+			if ( prt_params%woody(pft) == itrue) then !trees only, May require additional accounting for the atm_fluxes when PFT is grass. 
+													  ! Goal was to not add to pools in the event of grass fire. Grass would typically not introduce a flux of dead leaves	
+													  ! or coarse woody debris. This may be too coarse a solution in the long term. 
+                leaf_m   = currentCohort%prt%GetState(leaf_organ, element_list(el))
+                store_m  = currentCohort%prt%GetState(store_organ, element_list(el))
+                sapw_m   = currentCohort%prt%GetState(sapw_organ, element_list(el))
+                fnrt_m   = currentCohort%prt%GetState(fnrt_organ, element_list(el))
+                struct_m = currentCohort%prt%GetState(struct_organ, element_list(el))
+                repro_m  = currentCohort%prt%GetState(repro_organ, element_list(el))
+
+
              ! Absolute number of dead trees being transfered in with the donated area
              num_dead_trees = (currentCohort%fire_mort*currentCohort%n * &
                                patch_site_areadis/currentPatch%area)
@@ -1654,7 +1658,7 @@ contains
 
              ! Contribution of dead trees to leaf burn-flux
              burned_mass  = num_dead_trees * (leaf_m+repro_m) * currentCohort%fraction_crown_burned
-             
+
              do dcmpy=1,ndcmpy
                  dcmpy_frac = GetDecompyFrac(pft,leaf_organ,dcmpy)
                  new_litt%leaf_fines(dcmpy) = new_litt%leaf_fines(dcmpy) + &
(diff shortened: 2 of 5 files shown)


biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1637,13 +1637,17 @@ subroutine fire_litter_fluxes(currentSite, currentPatch, &
...
              ! -----------------------------------------------------------------------
              
-             sapw_m   = currentCohort%prt%GetState(sapw_organ, element_id)
-             struct_m = currentCohort%prt%GetState(struct_organ, element_id)
-             leaf_m   = currentCohort%prt%GetState(leaf_organ, element_id)
-             fnrt_m   = currentCohort%prt%GetState(fnrt_organ, element_id)
-             store_m  = currentCohort%prt%GetState(store_organ, element_id)
-             repro_m  = currentCohort%prt%GetState(repro_organ, element_id)
-             
+			if ( prt_params%woody(pft) == itrue) then !trees only, May require additional accounting for the atm_fluxes when PFT is grass.
```
jkshuman, line 1640: @ZacharyRobbins there is something going on with spacing here. Can you fix this so that the if statement is moved back to the left?

jkshuman (changes requested):
There is an issue with spacing in one section. Also, why are we sending fine roots to leaves?

biogeochem/EDPatchDynamicsMod.F90
ZacharyRobbins: I believe I have resolved the spacing issue, and return fine roots to their original location.

ckoven:
@ZacharyRobbins Thanks so much for submitting this PR.  I am trying to work through the code and have a couple questions.  I'm particularly struck by the big `if` block in `fire_litter_fluxes`, which skips most of the logic for plants that aren't woody (since the termination is [here](https://github.com/NGEET/fates/pull/891/files#diff-28c679efbe804905b8b4fac3ccdbb04f6bb3bcd2794fe1a4556d5541e008a750R1727)).  My first thought was that that shouldn't be necessary, since the `currentCohort%fraction_crown_burned` argument [here](https://github.com/NGEET/fates/blob/master/biogeochem/EDPatchDynamicsMod.F90#L1655-L1656) should mean that grasses are burning up when fire occurs anyway and so shouldn't be contributing their biomass to litter when fire happens.  But in staring at the (current) code, three things strike me: one is that we only let 80% of the grasses burn anyway ([here](https://github.com/NGEET/fates/blob/master/fire/SFMainMod.F90#L613-L614)), the second is that we only consider grass leaf carbon as live fuel, and not the sum of live fuel and stem carbon ([here](https://github.com/NGEET/fates/blob/master/fire/SFMainMod.F90#L191-L203)), and the third is that we then only burn off the leaf carbon and not the live stem carbon ([here](https://github.com/NGEET/fates/blob/master/biogeochem/EDPatchDynamicsMod.F90#L862-L875)).  

I feel like we probably need to make a choice, at least as regards to the last two points: FATES either treats all aboveground grass tissues as live fuel, and then burns then accordingly, or FATES only treats grass leaves as live fuel, and then the other aboveground tissues become litter.  Does that seem right, and if so is there a strong consensus which is the better choice?

jkshuman:
@ckoven thanks for digging into this. The 80% limit on grass burning is a legacy piece that should be tested. Perhaps @rosiealice has insight on this, but I recall it had to do with instability? The grass pieces of the code have not changed a lot, but other aspects of the fire code have changed. For burning of grasses, I vote to include all above ground grass tissues as live fuel. If the consensus goes the other way, then I agree those tissues should go to litter.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1637,96 +1637,94 @@ subroutine fire_litter_fluxes(currentSite, currentPatch, &
...
-             struct_m = currentCohort%prt%GetState(struct_organ, element_id)
-             leaf_m   = currentCohort%prt%GetState(leaf_organ, element_id)
-             fnrt_m   = currentCohort%prt%GetState(fnrt_organ, element_id)
-             store_m  = currentCohort%prt%GetState(store_organ, element_id)
-             repro_m  = currentCohort%prt%GetState(repro_organ, element_id)
-             
-             ! Absolute number of dead trees being transfered in with the donated area
-             num_dead_trees = (currentCohort%fire_mort*currentCohort%n * &
+			if ( prt_params%woody(pft) == itrue) then !trees only, May require additional accounting for the atm_fluxes when PFT is grass. 
+													  ! Goal was to not add to pools in the event of grass fire. Grass would typically not introduce a flux of dead leaves
```
rgknox, line 1641: I would like to see an else attached to this if clause where we use a logging message to indicate non-woody disturbance is not supported here, and include an endrun() and a verbose message. This prevents vulnerable code in case someone modifies disturbance generation to allow for non-woody, they will then realize it needs to be applied here as well.

ZacharyRobbins:
> I think your point is correct, I think that my be a relic of testing what needed to be changed to get the grass out of the CWD pool.  To clarify your point, you mean all live ag live grass tissue is treated as live fuel? or all above ground tissue (even dead tissue) as live fuel. I think it makes the most sense that all above ground grass is treated as crown consumed fuel while alive, and then all treated as leaves (or technically <1 hr fine fuels). 

To clarify, some people refer to these fine fuels as live fuels (as it readily burns).  In this instance I believe we are referring to live fuel as the pool called live-grass which is its own litter class (separate from dead leaves, which I think we are referring to as litter). Given that this class already exists I would say all living grass biomass should be placed in this pool, and parameters should reflect whole grass (stem, leaves) fuel characteristics. Dead grass should be treated as the pool dead leaves (as it is mostly like a <1hr fine fuel). 

I can remove the if statement in fire litter flux, does that still require a logging message?
Thanks.

jkshuman:
@ZacharyRobbins I agree with you on this that the live fuel for grass should include the sum of aboveground parts (only stem and leaf or also seeds?). 

In the current version live grass is the only live fuel considered, with crown fire other live fuels would be considered. Once grass dies it becomes part of the dead leaves pool consisting of dead tree leaves and dead grass. So, the grass is either alive or dead. These leaf pools are separate so that the live grass can have different fuel moisture compared to dead leaves. The fuel moisture of live grass changes in response to the Nesterov Index, as all fuels do, but follows the trajectory of 1 hr fuels so it retains more moisture than the dead leaves. Fuel moisture needs focused attention, and testing grasses with hydro would also be a big improvement.

For grasses in reality they go through a process of curing as they transition from live fuels to dead fuels, and this would be an exciting enhancement to add. With curing through the year there is a gradual transition associated with die-off that directly impacts flammability with higher curing being associated with increased flammability. Considering a difference in die-off or fuel moisture for the leaves vs stem of grasses might capture these flammability changes, but that should be its own development issue IMO.

jkshuman:
@ZacharyRobbins thank you for moving this forward. 
Can I impose on you to fix a documentation comment typo in SFMainMod with this PR?
Update "Bottom" to Wotton here:
  ! Information Report GLC-X-10 by Bottom et al., 2009 because there is a typo in CFFBPS Ont.Inf.Rep. ST-X-3, 1992)
The full first author name is B.M. Wotton, so I think the letters got transposed.

biogeochem/EDPatchDynamicsMod.F90
ZacharyRobbins: Is this outstanding still?

ZacharyRobbins:
I believe we have enacted all suggested changes. Please re-review.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1823,13 +1854,24 @@ subroutine mortality_litter_fluxes(currentSite, currentPatch, &
...
-          sapw_m   = currentCohort%prt%GetState(sapw_organ, element_id)
-          struct_m = currentCohort%prt%GetState(struct_organ, element_id)
-          leaf_m   = currentCohort%prt%GetState(leaf_organ, element_id)
           fnrt_m   = currentCohort%prt%GetState(fnrt_organ, element_id)
           store_m  = currentCohort%prt%GetState(store_organ, element_id)
           repro_m  = currentCohort%prt%GetState(repro_organ, element_id)
 
+          if (prt_params%woody(currentCohort%pft) == itrue) then
+             ! Assumption: for woody plants, we lump fluxes from deadwood and sapwood together in CWD pool.
+             ! for non-woody plants, we put all stem fluxes into the same leaf litter pool.
```
jkshuman, line 1863:
```suggestion
             ! Assumption: for woody plants fluxes from deadwood and sapwood go together in CWD pool
```

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1823,13 +1854,24 @@ subroutine mortality_litter_fluxes(currentSite, currentPatch, &
...
+             leaf_m          = currentCohort%prt%GetState(leaf_organ,element_id)
+             sapw_m          = currentCohort%prt%GetState(sapw_organ,element_id)
+             struct_m        = currentCohort%prt%GetState(struct_organ,element_id)
+          else
+             leaf_m          = currentCohort%prt%GetState(leaf_organ,element_id) + &
```
jkshuman, line 1868:
```suggestion
             ! for non-woody plants all stem fluxes go into the same leaf litter pool
             leaf_m          = currentCohort%prt%GetState(leaf_organ,element_id) + &
```

biogeochem/EDPhysiologyMod.F90
```diff
@@ -2193,24 +2193,41 @@ subroutine CWDInput( currentSite, currentPatch, litt, bc_in)
...
-       leaf_m          = currentCohort%prt%GetState(leaf_organ,element_id)
        store_m         = currentCohort%prt%GetState(store_organ,element_id)
        fnrt_m          = currentCohort%prt%GetState(fnrt_organ,element_id)
-       sapw_m          = currentCohort%prt%GetState(sapw_organ,element_id)
-       struct_m        = currentCohort%prt%GetState(struct_organ,element_id)
        repro_m         = currentCohort%prt%GetState(repro_organ,element_id)
 
+       if (prt_params%woody(currentCohort%pft) == itrue) then
+          ! Assumption: for woody plants, we lump fluxes from deadwood and sapwood together in CWD pool.
+          ! for non-woody plants, we put all stem fluxes into the same leaf litter pool.
```
jkshuman, line 2209:
```suggestion
          ! Assumption: for woody plants fluxes from deadwood and sapwood go together in CWD pool
```

biogeochem/EDPhysiologyMod.F90
```diff
@@ -2193,24 +2193,41 @@ subroutine CWDInput( currentSite, currentPatch, litt, bc_in)
...
+          leaf_m_turnover   = currentCohort%prt%GetTurnover(leaf_organ,element_id)
+          sapw_m_turnover   = currentCohort%prt%GetTurnover(sapw_organ,element_id)
+          struct_m_turnover = currentCohort%prt%GetTurnover(struct_organ,element_id)
+
+          leaf_m          = currentCohort%prt%GetState(leaf_organ,element_id)
+          sapw_m          = currentCohort%prt%GetState(sapw_organ,element_id)
+          struct_m        = currentCohort%prt%GetState(struct_organ,element_id)
+       else
+          leaf_m_turnover   = currentCohort%prt%GetTurnover(leaf_organ,element_id) + &
```
jkshuman, line 2218:
```suggestion
          ! for non-woody plants all stem fluxes go into the same leaf litter pool
          leaf_m_turnover   = currentCohort%prt%GetTurnover(leaf_organ,element_id) + &
```

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -1631,19 +1650,31 @@ subroutine fire_litter_fluxes(currentSite, currentPatch, &
...
-             leaf_m   = currentCohort%prt%GetState(leaf_organ, element_id)
+
              fnrt_m   = currentCohort%prt%GetState(fnrt_organ, element_id)
              store_m  = currentCohort%prt%GetState(store_organ, element_id)
              repro_m  = currentCohort%prt%GetState(repro_organ, element_id)
-             
+          
+             if (prt_params%woody(currentCohort%pft) == itrue) then
+                ! Assumption: for woody plants, we lump fluxes from deadwood and sapwood together in CWD pool.
+                ! for non-woody plants, we put all stem fluxes into the same leaf litter pool.
```
jkshuman, line 1665:
```suggestion
                ! Assumption: for woody plants fluxes from deadwood and sapwood go together in CWD pool
```

fire/SFMainMod.F90
```diff
@@ -195,7 +195,9 @@ subroutine  charecteristics_of_fuel ( currentSite )
           if( prt_params%woody(currentCohort%pft) == ifalse)then
```
jkshuman, line 195:
```suggestion
           ! for grasses sum all aboveground tissues
           if( prt_params%woody(currentCohort%pft) == ifalse)then 
```

fire/SFMainMod.F90
```diff
@@ -762,7 +764,7 @@ subroutine  area_burnt_intensity ( currentSite, bc_in )
                  lb = (1.0_r8 + (8.729_r8 * &
                       ((1.0_r8 -(exp(-0.03_r8 * m_per_min__to__km_per_hour * currentPatch%effect_wspeed)))**2.155_r8)))
              else ! EQ 80 grass fuels (CFFBPS Ont.Inf.Rep. ST-X-3, 1992, but with a correction from an errata published within 
-                  ! Information Report GLC-X-10 by Bottom et al., 2009 because there is a typo in CFFBPS Ont.Inf.Rep. ST-X-3, 1992)
+                  ! Information Report GLC-X-10 by Wotten et al., 2009 because there is a typo in CFFBPS Ont.Inf.Rep. ST-X-3, 1992)
```
jkshuman, line 767:
```suggestion
                  ! Information Report GLC-X-10 by Wotton et al., 2009 for a typo in CFFBPS Ont.Inf.Rep. ST-X-3, 1992)
```

jkshuman (changes requested):
Looks good, but I suggest small changes to wording and organizing of comments. I also fixed a typo in the name for Wotton et al in SFMain. At a minimum that typo should be fixed.

jkshuman (approved):
I checked in with Greg on this, and since they are just edits to comments, I went ahead and committed them. The PR is approved from me.

biogeochem/CMIP data.txt
```diff
@@ -0,0 +1,24 @@
+.
```
glemieux, line 1: @ZacharyRobbins @ckoven @jkshuman should this file be retained?

glemieux:
I pre-merged master into the PR branch locally on Cheyenne and ran the fates regression test suite against baseline `fates-sci.1.60.0_api.24.1.0-ctsm5.1.dev112`.  All tests ran successfully with DIFFs aside from the expected failures.  The DIFFS appear to be consistent with the changes as they start off with differences in the litter in/out and extremely small carbon balance errors and fuel variables and eventually show up in FATES_HET_RESP and FATES_NEP.

Test folder: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr891`

biogeochem/CMIP data.txt
ckoven: it should be deleted.

PR #892, opened 2022-08-22
Add two new ecosystem-level height metrics

ckoven:
This PR adds two metrics for average canopy height.  The first is a crown-area-weighted canopy height for all plants in the canopy. This approach averages in zero height for the unoccupied areas when the canopy is not closed. This is roughly what would be returned as the mean height from lidar returns, and is also close to what FATES is currently doing for aerodynamic roughness height scaling.  The second new variable is a basal-area-weighted average height of all trees; this is equivalent to what is called "Lorey's height" in the forestry literature. This variable doesn't take into account canopy strata or canopy closure.

### Collaborators:
email discussion thread with @mpaiao @jkshuman @JessicaNeedham @adamhb @pollybuotte @lmkueppers @XiulinGao

This also changes two numbers in adjacent lines of code that are very close to pi to just make them pi, so it will be answer changing for the existing basal area calculation as well as the 2 new variables.

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -195,8 +195,8 @@ module FatesHistoryInterfaceMod
 
   integer :: ih_fates_fraction_si
 
-  integer :: ih_ba_weighted_height
-  integer :: ih_ca_weighted_height
+  integer :: ih_ba_weighted_height_si
+  integer :: ih_ca_weighted_height_si
 
   integer :: ih_cwd_elcwd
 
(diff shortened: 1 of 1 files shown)


glemieux (approved):

glemieux:
Fates test suite run against the `fates-sci.1.59.4_api.24.1.0-ctsm5.1.dev106` baseline shows all expected pass, with expected DIFFs in testmods that report `FATES_BASALAREA_SZPF`.   This is due to the update to the pi constant as noted in the PR description. 
Tests results here: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr892`

main/FatesHistoryInterfaceMod.F90
```diff
@@ -2954,6 +2966,13 @@ subroutine update_history_dyn(this,nc,nsites,sites)
          cpatch => cpatch%younger
       end do patchloop !patch loop
 
+      ! divide basal-area-weighted height by basal area to get mean
+      if ( sum(hio_ba_si_scpf(io_si,:)) .gt. tiny ) then
```
rgknox, line 2970: I've never seen tiny used as a constant, ie not used as a function call. My only concern here is that not all compilers support it and perhaps not part of the fortran standard. I could be wrong but just don't know.  tiny()  or nearzero are safe though.

rgknox: correction, I see it defined here: https://github.com/NGEET/fates/blob/2c2eb561584833c9e221eea13727c9ef4d11b809/main/FatesHistoryInterfaceMod.F90#L1802
Can we get rid of tiny and use nearzero here?

glemieux:
Retesting per [82dbd83](https://github.com/NGEET/fates/pull/892/commits/82bdb8327631a34ca015ae2c94c04b32403e2c9c) update to change all uses of `tiny` to `nearzero`.

rgknox (approved):

glemieux:
Re-test shows the same results as previously found.  

`/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr892_nearzero`

PR #931, opened 2022-11-08
Add Atkin et al 2017 leaf respiration model

### LINKED ISSUE #729: Update FATES respiration model

ckoven:
This code adds the option for a new leaf maintenance respiration model based on Atkin et al 2017.

### Description:
We've been using the Ryan 1991 https://doi.org/10.2307/1941808 maintenance respiration model in FATES so far.  Newer schemes have been developed that are based on a lot more data.  Discussion in #729 pointed to the Atkin et al 2017 https://doi.org/10.1007/978-3-319-68703-2_6 as one that is based on a very large amount of observations. The model also suggests that respiration acclimates with growing temperature, with important implications (e.g. as explored in Huntingford et al https://doi.org/10.1038/s41467-017-01774-z)

fixes #729.

Some things to think about:

(1) Vertical N scaling and how it scales with respiration: In the Ryan 1991 model leaf respiration scales directly with leaf N, both at the leaf level and therefore also at the canopy level.  Whereas in the Atkin et al 2017 model, respiration is not exactly proportional to leaf nitrogen -- there is a linear leaf N term but also an offset (due to the base PFT-dependent R_0 and growth-temperature-acclimation R_3 terms). So  I think we need to understand how this governs the vertical profiles of respiration (and thus also the vertical profiles of leaf costs vs benefits that governs LAI).

(2) This code allows respiration, but not Vcmax, to acclimate to temperature.  @alistairrogers [makes the point](https://github.com/NGEET/fates/issues/729#issuecomment-816779479) that we probably want to shift to acclimating vcmax and respiration together. So possibly we may want to keep the old respiration model the default until we also have vcmax acclimation merged and also ready to use?

(3) This code only applies to leaves of C3 plants.  Respiration from C4 plant leaves and all non-leaf organs is unchanged here.

(4) We may want to further encapsulate the C4 plant logic into its own subroutine for neatness. Also there are still some preexisting inline magic numbers in the old respiration code that could be better cleaned up and/or pulled out as parameters (Noted I think [here](https://github.com/NGEET/fates/issues/729#issuecomment-1060982197)).

(5) This code pulls out the R_0 terms as PFT parameters (which will need to be moved to a dedicated variable from their current spot in the `EDPftvarcon_inst%dev_arbitrary_pft` variable) but not the scalar parameters.  We may want to pull all of them out into the parameter file?

(6) Overall I think a key question is whether we would want to make this the default leaf respiration model prior to / as part of the larger calibration exercise or not.  This will tend to increase leaf respiration, which the model with current defaults is biased low on, but possibly require other more complex things to be worked out.

### Collaborators:
Discussion with @JessicaNeedham @jenniferholm @rgknox and also discussion in #729.

This will change answers if activated. If we don't activate it, it should be bit for bit.  Activating the new respiration model is via the parameter file flag `fates_maintresp_model `.

### Checklist:

### Test Results:
not yet run through test suite.

FATES baseline hash-tag:

Test Output:

### DIFF
--- a/biogeophys/FatesPlantRespPhotosynthMod.F90
+++ b/biogeophys/FatesPlantRespPhotosynthMod.F90
@@ -2071,7 +2071,7 @@ subroutine LeafLayerMaintenanceRespiration_Atkin_etal_2017(lnc_top, &
    ! and Heskel et al., 2016 https://doi.org/10.1073/pnas.1520282113
    real(r8), parameter :: b = 0.1012_r8       ! (degrees C**-1)
    real(r8), parameter :: c = -0.0005_r8      ! (degrees C**-2)
-   real(r8), parameter :: Tref = tfrz+25._r8  ! (degrees K)
+   real(r8), parameter :: TrefC = 25._r8      ! (degrees C)
    real(r8), parameter :: r_1 = 0.2061_r8     ! (umol CO2/m**2/s / (gN/(m2 leaf))) 
    real(r8), parameter :: r_2 = -0.0402_r8    ! (umol CO2/m**2/s/degree C)
 
@@ -2086,13 +2086,14 @@ subroutine LeafLayerMaintenanceRespiration_Atkin_etal_2017(lnc_top, &
    if ( nint(EDPftvarcon_inst%c3psn(ft)) == 1)then
 
       ! r_0 currently put into the EDPftvarcon_inst%dev_arbitrary_pft
+      ! all figs in Atkin et al 2017 stop at zero Celsius so we will assume acclimation is fixed below that
       r_0 = EDPftvarcon_inst%dev_arbitrary_pft(ft)
-      r_t_ref = r_0 + r_1 * lnc + r_2 * tgrowth
+      r_t_ref = r_0 + r_1 * lnc + r_2 * max(0._r8, (tgrowth - tfrz) )
 
-      lmr = r_t_ref * exp(b * (veg_tempk - Tref) + c * (veg_tempk**2 - Tref**2))
+      lmr = r_t_ref * exp(b * (veg_tempk - tfrz - TrefC) + c * ((veg_tempk-tfrz)**2 - TrefC**2))
 
    else
-      ! revert to Q10 model for C4 plants, parameter values as described above
+      ! revert to Q10 model for C4 plants, parameter values as described above in Ryan 1991 method
 
       lmr25top = 2.525e-6_r8 * (1.5_r8 ** ((25._r8 - 20._r8)/10._r8))
       lmr25top = lmr25top * lnc / (umolC_to_kgC * g_per_kg)
--- a/biogeophys/FatesPlantRespPhotosynthMod.F90
+++ b/biogeophys/FatesPlantRespPhotosynthMod.F90
@@ -40,6 +40,10 @@ module FATESPlantRespPhotosynthMod
   use PRTGenericMod,     only : max_nleafage
   use EDTypesMod,        only : do_fates_salinity
   use EDParamsMod,       only : q10_mr
+  use EDParamsMod,       only : maintresp_model
+  use FatesConstantsMod, only : lmrmodel_ryan_1991
+  use FatesConstantsMod, only : lmrmodel_atkin_etal_2015
+  use FatesConstantsMod, only : lmrmodel_heskel_etal_2018
   use PRTGenericMod,     only : prt_carbon_allom_hyp
   use PRTGenericMod,     only : prt_cnp_flex_allom_hyp
   use PRTGenericMod,     only : all_carbon_elements
@@ -520,17 +524,40 @@ contains
 
                                end select
 
-                               ! MLO - Shouldn't these numbers be parameters too?
-                               lmr25top = 2.525e-6_r8 * (1.5_r8 ** ((25._r8 - 20._r8)/10._r8))
-                               lmr25top = lmr25top * lnc_top / (umolC_to_kgC * g_per_kg)
-
-
                                ! Part VII: Calculate dark respiration (leaf maintenance) for this layer
-                               call LeafLayerMaintenanceRespiration( lmr25top,                 &  ! in
-                                    nscaler,                  &  ! in
-                                    ft,                       &  ! in
-                                    bc_in(s)%t_veg_pa(ifp),   &  ! in
-                                    lmr_z(iv,ft,cl))             ! out
+                               select case (maintresp_model)
+
+                               case (lmrmodel_ryan_1991)
+
+                                  ! MLO - Shouldn't these numbers be parameters too?
+                                  lmr25top = 2.525e-6_r8 * (1.5_r8 ** ((25._r8 - 20._r8)/10._r8))
+                                  lmr25top = lmr25top * lnc_top / (umolC_to_kgC * g_per_kg)
+
+                                  call LeafLayerMaintenanceRespiration_Ryan_1991( lmr25top,     &  ! in
+                                       nscaler,                  &  ! in
+                                       ft,                       &  ! in
+                                       bc_in(s)%t_veg_pa(ifp),   &  ! in
+                                       lmr_z(iv,ft,cl))             ! out
+
+                               case (lmrmodel_atkin_etal_2015)
+
+                                  call LeafLayerMaintenanceRespiration_Atkin_etal_2015(lnc_top, &  ! in
+                                       nscaler,                  &  ! in
+                                       ft,                       &  ! in
+                                       bc_in(s)%t_veg_pa(ifp),   &  ! in
+                                       currentPatch%tveg_lpa,    &  ! in
+                                       lmr_z(iv,ft,cl))             ! out
+
+                               case (lmrmodel_heskel_etal_2016)
+
+                                  call LeafLayerMaintenanceRespiration_Heskel_etal_2016(lnc_top, &  ! in
+                                       nscaler,                  &  ! in
+                                       ft,                       &  ! in
+                                       bc_in(s)%t_veg_pa(ifp),   &  ! in
+                                       currentPatch%tveg_lpa,    &  ! in
+                                       lmr_z(iv,ft,cl))             ! out
+
+                               end select
 
                                ! Part VII: Calculate (1) maximum rate of carboxylation (vcmax),
                                ! (2) maximum electron transport rate, (3) triose phosphate
@@ -1966,7 +1993,7 @@ end subroutine GetCanopyGasParameters
 
 ! ====================================================================================
 
-subroutine LeafLayerMaintenanceRespiration(lmr25top_ft, &
+subroutine LeafLayerMaintenanceRespiration_Ryan_1991(lmr25top_ft, &
    nscaler,   &
    ft,        &
    veg_tempk,     &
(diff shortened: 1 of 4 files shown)


ckoven:
Thanks @alistairrogers.  Just wanted to note that the equations in Atkin et al 2017 will (I think) give the opposite sign to what you describe for the vertical profile in the ratio of Rd to Vcmax.  Since the form of [the equation](https://github.com/NGEET/fates/pull/931/files#diff-f62a39ba3dc682263ce52ac5f1bec68ac12bc849e5987c1a793938bb2a1339e8R2091) is (R0 + R1*N + R2*Tgrowth), given the reported coefficients for R0 and R2, as long as Tgrowth is less than 37 Celsius, the sum of the first and third terms will be positive and so for low N/area as found deeper in the canopy, there will be relatively more respiration than N/area.  That will hold for leaf respiration scaling within a given plant; between plants of different PFTs there could be other differences depending on their relative traits, and between plants of the same PFT in different canopy strata there could be differences based on the FATES logic that governs carbon starvation. But it isn't obvious that this scheme is consistent with the vertical profiles that Julien measures.

alistairrogers:
Thanks doing laying that out @ckoven. I think the use of Atkin to resolve canopy gradient of Rd will lead to problems. Their approach should be sound for top of canopy Rd but I think will produce the opposite of what we would expect within a canopy. Respiration basically has two components the (1) supply - availability of substrate (photosynthate) to fuel respiratory processes which is lower closer to the ground (lower photosynthetic rates) and (2) the cost of maintaining existing function which is essentially driven by how much protein, primarily associated with photosynthesis, that you have. As we know leaf N and protein decrease as you get closer to the ground a potential third component (which would explain Julien's observations) would be respiration associated with repair from excess light (obviously higher at the top of the canopy). I find it hard to think of a scenario whereby photosynthetic capacity would be negatively correlated with Rd within a canopy. This is perhaps why the model assumption of Rd = Vcmax * 0.015 is so robust. Perhaps FATES could use Atkin to set top-of-canopy Rd but then scale within a canopy using either existing fixed ratios or new ratios based on Julien's work.

parameter_files/fates_params_default.cdl
```diff
@@ -1020,7 +1020,8 @@ data:
 
  fates_damage_recovery_scalar = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ;
 
- fates_dev_arbitrary_pft = _, _, _, _, _, _, _, _, _, _, _, _ ;
+ fates_dev_arbitrary_pft = 1.7560, 1.4995, 1.4995, 1.7560, 1.7560, 1.7560,
```
rgknox, line 1023: @ckoven what would we call this parameter?  I see it matches r_0: "! base respiration rate, PFT-dependent (umol CO2/m**2/s)"

Also, is there a similar parameter in the Ryan model that could also leverage the parameter?

I could add this parameter to the nutrient coupling PR (which has other parameter changes, and assuming it gets through queue more quickly).

parameter_files/fates_params_default.cdl
```diff
@@ -1490,7 +1491,7 @@ data:
 
  fates_leaf_theta_cj_c4 = 0.999 ;
 
- fates_maintresp_model = 1 ;
+ fates_maintresp_model = 2 ;
```
rgknox, line 1494: Also, what is our staging for making the change in the default file?  My sense from discussion with others is that the Atkin model is the path forward.  However, all of the other parameters are calibrated around the Ryan model, and this will undoubtedly have an impact on output.

Was the plan to switch over to Atkins first, and accept the differences good or bad to output, and then continue with other calibrations? Or was the plan to switch over to Atkins during a more comprehensive calibration exercise perhaps with the satphen simuluation or other biogeographically constrained testbed?

In either situations, we should probably run this through the ilamb mill to get a sense of the changes.

parameter_files/fates_params_default.cdl
ckoven: Good question, and good idea.  Maybe `fates_maintresp_atkinetal2017model_baserate`?  There isn't a direct analogue in the Ryan 1991 model.

parameter_files/fates_params_default.cdl
ckoven: I think for staging purposes, we should bring in the code first with this switch reverted to the current value, and if we decide to make it the default model, that would likely need to be accompanied by other default parameter changes as well.

ckoven:
@alistairrogers That is an interesting idea, to use the published equation for relating leaf N/area to respiration rates only to the top of canopy value and then assuming proportionality with N/area vertically through the canopy.  Worth exploring the dynamics of that at least, and seeing whether the evidence available is generally supportive of it. Per @rosiealice's comment, I'm not sure if the available data speaks to it either way, but agreed that we should investigate.

parameter_files/fates_params_default.cdl
rgknox: how about: fates_maintresp_atkin2017_baserate

ckoven: @rgknox that works for me, thanks!

ckoven:
I want to raise a question here.  As best I can tell, none of the Atkins papers really cover C4 plants, since they are not well represented in the GlobResp database that that parameterization is based on. I had thought that CLM5 defaulted entirely to the old model for C4 plants, but I just noticed [here in the tech note](https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Plant_Respiration/CLM50_Tech_Note_Plant_Respiration.html#id3) and [here in the code](https://github.com/ESCOMP/CTSM/blob/master/src/biogeophys/PhotosynthesisMod.F90#L3277-L3285), that it is actually using the same Atkin et al 2015 algorithm and traits for C4 grasses as for C3 grasses for the base respiration rate (which acclimates based on 10-day temperature), but then [a different short-term temperature scaling for C3 and C4 plants](https://github.com/ESCOMP/CTSM/blob/master/src/biogeophys/PhotosynthesisMod.F90#L3334-L3344) where the C4 plants follow the old model. JULES does not use the Atkin scheme for C4 plants, at least as reported [here in the Huntingford et al 2017 paper](https://www.nature.com/articles/s41467-017-01774-z/tables/2). Big-leaf ELM still [uses the pre-CLM5 scheme](https://github.com/E3SM-Project/E3SM/blob/master/components/elm/src/biogeophys/PhotosynthesisMod.F90#L687-L746) for all PFTs (which includes the different temperature scaling for C4 plants).  I don't know what the best solution here is, so figured I'd at least log this here and open for discussion if anyone knows the best thing to do.

dlawrenncar:
Maybe a discussion at a CLM meeting would be good.  Not sure best practice,
but should this issue be repeated or linked on the CTSM github?

jkshuman:
@danicalombardozzi any thoughts or recommendations on the C4 conversation here? Or recommendations for moving forward.

ckoven:
Noting a couple brief updates to this PR. I have cleaned up the C4 logic, to put that into its own subroutine.  A couple of decisions had to be made here for the integration of the Atkin 2017 model to FATES. For vertical scaling, I think that @alistairrogers's proposal of keeping the vertical profiles of respiration proportional to the vertical profiles of leaf N makes the most sense, so have gone that route here. Since none of the C4 options seem all that great, I am for now at least keeping the C4 reverted to the current/old Ryan 1991 model. Also I consolidated the basal respiration rate parameter for the Ryan 1991 model to use the value in the parameter file rather than the hard-coded number (and updated the value in the parameter file to be the old hard-coded number), and have set the switch in the parameter file that it still use the Ryan 1991 model by default.  A subsequent PR can then switch the default model to be the Atkin 2017 model. I believe that this PR should now be ready for review and integration.

glemieux:
Per @ckoven suggestion, I tested this PR on a [custom branch](https://github.com/glemieux/fates/tree/respiration_newmodels_merge_revert-10db8b9) that reverts commit 10db8b9 to confirm that all other changes are B4B.  After merging in master, testing against baseline `fates-sci.1.63.1_api.25.1.0-ctsm5.1.dev117` all expected tests passed B4B. 

folder location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr931_revert-commit-10db8b9`

parameter_files/fates_params_default.cdl
```diff
@@ -1023,7 +1023,8 @@ data:
 
  fates_damage_recovery_scalar = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ;
 
- fates_dev_arbitrary_pft = _, _, _, _, _, _, _, _, _, _, _, _ ;
+ fates_dev_arbitrary_pft = 1.7560, 1.4995, 1.4995, 1.7560, 1.7560, 1.7560, 
+    2.0749, 2.0749, 2.0749, 2.1956, 2.1956, _ ;
```
glemieux, line 1027: @ckoven to clarify, the last entry here is supposed to be `_`?

ckoven: Yes, it relates to the C4 question. Since we aren't using this scheme for C4 plants at this point, I thought it would be clearest to leave the parameter missing for them as well.

glemieux: Roger that.  Thanks!

glemieux:
Per @ckoven I tested this pr at https://github.com/NGEET/fates/pull/931/commits/22212b5dae2d90e923724cdb0d2bfdf6b1b4c0b5 to make sure that this would run in all the modes we test knowing there would be diffs.  All expected tests ran successfully.

Folder location on cheyenne: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr931-atkin-on`

glemieux:
@rgknox I updated the default parameter file to revert the changes to the `fates_dev_arbitrary_pft` variable and add the new respiration parameter.  I created two patch files and updated the tooling to do this as the new parameter has an mixed values which the tooling couldn't quite handle.  I also fixed the BatchPatchParameter.py tool to make sure the sorting was the last thing being called as we discussed.

I've got final regression tests against the current baseline going on cheyenne now.

glemieux:
All expected tests pass on cheyenne.  The tests are not b4b as expected due to the bug fix in https://github.com/NGEET/fates/commit/10db8b92373e766cfd0bcfe5b433fedb6de11bc4 that is now using the appropriate parameter variable instead of the hardcoded value.  It should be noted that the parameter file will be updated with the correction to this variable in the HLMs at a later date so the baselines will have non-b4b differences during that update as well.

Location on cheyenne: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr931-final`

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -2070,26 +2107,125 @@ subroutine LeafLayerMaintenanceRespiration(lmr25top_ft, &
...
+   real(r8), intent(in)  :: nscaler      ! Scale for leaf nitrogen profile
+   real(r8), intent(in)  :: veg_tempk    ! vegetation temperature  (degrees K)
+   real(r8), intent(in)  :: tgrowth      ! lagged vegetation temperature averaged over acclimation timescale (degrees K)
+   real(r8), intent(out) :: lmr          ! Leaf Maintenance Respiration  (umol CO2/m**2/s)
+
+   ! Locals
+   real(r8) :: lmr25   ! leaf layer: leaf maintenance respiration rate at 25C (umol CO2/m**2/s)
+   real(r8) :: r_0     ! base respiration rate, PFT-dependent (umol CO2/m**2/s)
+   real(r8) :: r_t_ref ! acclimated ref respiration rate (umol CO2/m**2/s)
+   real(r8) :: lnc     ! Leaf nitrogen content per unit area at this level [gN/m2]
```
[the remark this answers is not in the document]
glemieux: Appears so.  @ckoven?

glemieux:
Per the software meeting today, it was deceided that @ckoven would test out using the C3 Atkin parameter for the C4 grasses.  If this doesn't break the model then we will make the update to the parameter file and the parameter xml patch file.  It was also decided to update the parameter file to utilize a new agnostic naming scheme for the related variables per @rgknox's slides here: https://docs.google.com/presentation/d/1tnttlh9CGFG5fTRjyZf-ALXMKyVOF8OnFnQqmTyJmJk/edit

biogeophys/FatesPlantRespPhotosynthMod.F90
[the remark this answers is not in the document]
ckoven: this is fixed now.

ckoven:
I've updated this to reflect the new decision on what to do about C4 grasses, which is now also in the code for this PR. The decision is that it doesn't really make sense to use an entirely different leaf maintenance respiration model for C3 and C4 plants, particularly given the large difference in baseline maintenance respiration rates between the two models, which could undermine attempts to meaningfully compete C3 and C4 grass PFTs. So, in the absence of better options, we decided to use the same base respiration rates and temperature sensitivities from C3 grasses for the C4 grasses as well, until such time as better data is available to parameterize the C4s separately. Note that this is also what CLM5 does. Choosing the Ryan 1991 model reverts to the same behavior as before (where C4 grasses have the same basal respiration rates but a different temperature sensitivity than the C3 PFTs).

glemieux:
Regression tests on Cheyenne are complete.  All expected tests pass RUN and all result in DIFFs due to the fix [mentioned above](https://github.com/NGEET/fates/pull/931#issuecomment-1446745231).

Folder location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr931-c4update`

rgknox (approved):

Issue #994, opened 2023-03-02
Bare ground establishment problem and increased soil moisture 


JessicaNeedham:
This issue is related to discussion #985 . Two changes that improve the lack of bare ground establishment in some grid cells is to initialise the run with large trees rather than seedlings, and to increase initial soil moisture. 

I will make a PR with changes that allow for no comp mode to be initialised with large trees. 

Soil moisture initialisation is on the HLM side so will need ELM and CLM PRs. 

ELM code here: 
https://github.com/E3SM-Project/E3SM/blob/b30a6a0cb3e610f80a00cd0fcdbfffab7044df54/components/elm/src/data_types/ColumnDataType.F90#L1619

Not sure about the CLM side. 

I suggest changing use_fates_planthydro to  use_fates so that higher soil  moisture is default for all FATES runs.

JessicaNeedham:
PR  for ELM  is here: 
https://github.com/E3SM-Project/E3SM/pull/5508

PR #995, opened 2023-03-02
Initialize nocomp mode with large trees

JessicaNeedham:
### Description:
This PR allows nocomp mode to be initialised with large trees by allowing the initial_density parameter  to be interpreted as initial dbh when negative. It then calculates the crown area of a single tree using that DBH, and calculates number density needed to fill the canopy. 
Related to issue #994 and discussion #985 

### Collaborators:
@ckoven @jenniferholm @rgknox @glemieux 

This PR will change answers only in nocomp mode when the  parameter fates_recruit_init_density  is  set to a negative number.  

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:


main/EDInitMod.F90
```diff
@@ -839,151 +843,174 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
-             call bagw_allom(temp_cohort%dbh,pft,temp_cohort%crowndamage,c_agw)
+          else ! interpret as initial diameter and calculate density 
+             if(hlm_use_nocomp .eq. itrue)then
+                temp_cohort%dbh = abs(EDPftvarcon_inst%initd(pft))
 
-             ! Calculate coarse root biomass from allometry
-             call bbgw_allom(temp_cohort%dbh,pft,c_bgw)
+                ! calculate crown area of a single plant
+                dummy_n = 1.0_r8 ! make n=1 to get area of one tree
+                spread = 1.0_r8  ! fix this to 0 to remove dynamics of canopy clousre, assuming a closed canopy.
```
rgknox, line 857: There are these two named constants that you can optionally use for spread, if you like. Seems like we initialize with a spread of 0 for an inventory init, 1 for a bare ground init.
https://github.com/NGEET/fates/blob/sci.1.64.2_api.25.2.0/main/EDTypesMod.F90#L108-L109

main/EDInitMod.F90
```diff
@@ -839,151 +843,174 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
+          call prt_obj%CheckInitialConditions()
+
+          call create_cohort(site_in, patch_in, pft, temp_cohort%n, temp_cohort%hite, &
+               temp_cohort%coage, temp_cohort%dbh, prt_obj, cstatus, rstatus,        &
+               temp_cohort%canopy_trim, temp_cohort%c_area,1,temp_cohort%crowndamage, site_in%spread, bc_in)
+
+          write(fates_log(),*) 'c_area: ', temp_cohort%c_area
+          write(fates_log(),*) 'p_area: ', patch_in%area
+
+          deallocate(temp_cohort) ! get rid of temporary cohort
```
rgknox, line 997: this is a duplicate deallocate, will be harmless, but better to remove it because the following deallocation has more logging.

main/EDInitMod.F90
JessicaNeedham: Looks like I should be using init_spread_inventory since this isn't a near bare ground case. I've changed it now.

main/EDInitMod.F90
JessicaNeedham: Oh good catch, thanks. Deleted the duplicate.

main/EDInitMod.F90
```diff
@@ -839,151 +842,171 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
+
+          end do
+
+          call prt_obj%CheckInitialConditions()
+
+          call create_cohort(site_in, patch_in, pft, temp_cohort%n, temp_cohort%hite, &
+               temp_cohort%coage, temp_cohort%dbh, prt_obj, cstatus, rstatus,        &
+               temp_cohort%canopy_trim, temp_cohort%c_area,1,temp_cohort%crowndamage, site_in%spread, bc_in)
+
+          write(fates_log(),*) 'c_area: ', temp_cohort%c_area
```
rgknox, line 992: these look temporary, can remove

main/EDInitMod.F90
```diff
@@ -803,15 +807,21 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
-          if(EDPftvarcon_inst%initd(pft)>nearzero) then
 
-             allocate(temp_cohort) ! temporary cohort
+          allocate(temp_cohort) ! temporary cohort
+          temp_cohort%pft         = pft
+          temp_cohort%l2fr = prt_params%allom_l2fr(pft)
+          temp_cohort%canopy_trim = 1.0_r8
+          temp_cohort%crowndamage = 1  ! Assume no damage to begin with
+
+          if(EDPftvarcon_inst%initd(pft)>nearzero) then   ! interpret as initial density and calculate diameter
```
rgknox, line 823: @JessicaNeedham Should we gracefully fail (during parameter reading) a cold-start simulation where initd=0 (ie abs(initd)<nearzero)  ?  It seems like if someone sets the diameter to zero, or the density to zero, the run will fail anyway.  Although the user may be wondering what happened.

main/EDInitMod.F90
JessicaNeedham: Removed

main/EDInitMod.F90
JessicaNeedham: Good idea. I've added a check in EDPftvarcon

adrifoster (approved):
This all looks great to me. Thanks for adding the fail. I guess my only thing is that I would prefer having the dbh initialization be a separate parameter rather than used when `initd` is negative, but it's not a super strong opinion.

JessicaNeedham:
Thanks for the reviews @rgknox and @adrifoster !

@adrifoster I agree that initial dbh value might be better as its own parameter. It's kind of confusing to have parameters mean different things depending on context. Unless anyone has strong feelings otherwise I will go ahead and add it as a new parameter. 

Couple of questions on how best to do this: 

1. Should it still have the recruit prefix given it might be used to initialise with large trees? fates_recruit_init_dbh seems like the obvious name but I'm not sure if the recruit bit is confusing. 
2. Do we need another parameter to specify whether this is a density or dbh initialisation (e.g. no_comp_cold_start_mode)? An alternative would be to have the initial_dbh parameter as unset and a note in the long  name that if set to positive values (in no comp mode) they will be used and will override the  init_density values?

JessicaNeedham:
At the software meeting today we discussed whether to add another parameter v include better comments in the code to explain the dual function of the fates_recruit_init_density parameter. We decided on better comments so I have gone ahead and added those to EDInitMod.

glemieux:
Regression testing on Cheyenne underway

main/EDInitMod.F90
```diff
@@ -839,151 +848,168 @@ subroutine init_cohorts( site_in, patch_in, bc_in)
...
+             endif
+          endif
 
-             ! Calculate sapwood biomass
-             call bsap_allom(temp_cohort%dbh,pft,temp_cohort%crowndamage, &
-                  temp_cohort%canopy_trim,a_sapw,c_sapw)
+          ! Calculate the leaf biomass from allometry
+          ! (calculates a maximum first, then applies canopy trim)
+          call bleaf(temp_cohort%dbh,pft,temp_cohort%crowndamage, &
+               temp_cohort%canopy_trim,c_leaf)
```
glemieux, line 878: I'm seeing non-b4b results for sp mode only regression tests for the default parameter file (i.e. with `initd = 0.2`).  I think it might be due to this change here.  The call to `bleaf` was originally being avoided for sp mode runs as part of the check starting at line 842.  Should it be moved back inside?

JessicaNeedham: Oh good catch. It does also need to be called in the no comp dbh initialisation mode though so maybe we need two calls - in both dbh and density part of the if statement starting 829?

glemieux:
All expected tests are B4B against the `fates-sci.1.65.1_api.25.3.0-ctsm5.1.dev118` baseline.  

Folder on Cheyenne: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr995`

Issue #696, opened 2020-09-21
remove more calls to ELM/CLM globals


rgknox:
In this PR by @bishtgautam:

https://github.com/NGEET/fates/pull/695

There are still some calls from FATES to ELM/CLM routines and data structures. These calls need to be removed and replaced by interface structures.

glemieux:
This is still open.  See the linked closed PR above.  The only issue is with `clm_time_manager` in FatesPlantHydraulicsMod.

rgknox:
This is a call from FATES to a CLM/ELM routine. This can be removed by calling it from the CLM/ELM side of the interface, should be pretty straight forward, but we'll have to create a coupled API-change PR.
https://github.com/NGEET/fates/blob/sci.1.64.2_api.25.1.0/main/FatesInterfaceMod.F90#L744

rgknox:
This line was pretty straight forward, and I'm just removing it: https://github.com/NGEET/fates/blob/sci.1.64.2_api.25.1.0/biogeophys/FatesPlantHydraulicsMod.F90#L101
The variables aren't used

PR #998, opened 2023-03-09
issue triage - three small issues

### LINKED ISSUE #708: prescribed nuptake and puptake value checking
When running a carbon only run, we should not enforce expected range of prescribed N uptake and P uptake.

Here is an error that charlie generated:

Negative values for EDPftvarcon_inst% prescribed_nuptake(:)
 are not allowed. Reasonable ranges for this parameter are zero
 to something slightly larger than 1, so we set a cap at 10.
 Set to zero to turn off and use coupled nutrients.
  Aborting
 ENDRUN:
 ERROR in EDPftvarcon.F90 at line 1484

### LINKED ISSUE #715: SEED_GERM_ELEM history variable is basically zeros
I just noticed that the history variable SEED_GERM_ELEM, which is the total element flux (so mass of C,N, etc per area per time) of seed transforming into new recruits, is basically all zeros.  looking through the code, I think its because of how we're handling the litter%seed_germ here: https://github.com/NGEET/fates/blob/master/biogeochem/EDPhysiologyMod.F90#L1830-L1832 where we zero the flux prior to it arriving in the history diagnostics (here: https://github.com/NGEET/fates/blob/master/main/FatesHistoryInterfaceMod.F90#L3144-L3145 ).  not a super big deal, but f we want to track the mass of recruitment fluxes, we may need to rework slightly in order to send the right thing to history.

rgknox:
### Description:

This PR will:   
fix #715
fix #708 

partially address: #696

### Collaborators:

@glemieux @ckoven 

This will not change answers, except: FATES_SEED_GERM_EL  and FATES_SEED_BANK

### Test Results:
TBD

Test Output:

### DIFF
--- a/biogeophys/FatesPlantHydraulicsMod.F90
+++ b/biogeophys/FatesPlantHydraulicsMod.F90
@@ -98,8 +98,6 @@ module FatesPlantHydraulicsMod
   use PRTGenericMod,          only : num_elements
   use PRTGenericMod,          only : element_list
 
-  use clm_time_manager  , only : get_step_size, get_nstep
-
   use EDPftvarcon, only : EDPftvarcon_inst
   use PRTParametersMod, only : prt_params
 
@@ -1097,7 +1095,6 @@ contains
     real(r8) :: th_uncorr                    ! Uncorrected water content
     real(r8), parameter :: small_theta_num = 1.e-7_r8  ! avoids theta values equalling thr or ths         [m3 m-3]
 
-    integer :: nstep !number of time steps
     !-----------------------------------------------------------------------
 
     ccohort_hydr => ccohort%co_hydr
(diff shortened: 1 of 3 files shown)


glemieux (approved):

glemieux:
Regression testing on Cheyenne in progress.

glemieux:
All tests that do not include either `FATES_SEED_BANK` or `FATES_SEED_GERM_EL` PASS b4b against baseline `fates-sci.1.65.0_api.25.3.0-ctsm5.1.dev118` as expected.  All other expected tests run successfully with DIFFs only against those history outputs.

Folder on Cheyenne: `/glade/scratch/glemieux/ctsm-tests/tests_pr998`

Issue #697, opened 2020-09-22
Replace ALM to ELM as supported hlms


bishtgautam:
- In FatesHistoryInterfaceMod.F90, `hlms='CLM:ALM'` should be updated to `hlms='CLM:ELM'` (e.g. [FatesHistoryInterfaceMod.F90#L3881](https://github.com/NGEET/fates/blob/master/main/FatesHistoryInterfaceMod.F90#L3881)), **and**
- ELM code in [clmfates_interfaceMod.F90](https://github.com/E3SM-Project/E3SM/blob/master/components/elm/src/main/clmfates_interfaceMod.F90#L290) need to be updated accordingly

bishtgautam:
@rgknox If I find more issues related to the name change in ELM, would you prefer that I use a single issue that tracks all needed changes in FATES, or would you prefer I create separate issues?

rgknox:
lets corral them into one thread if possible

bishtgautam:
Sounds good. I will update this issue if additional changes are needed.

glemieux:
This will require a coupled update with e3sm.  Add this to the next API update for the upcoming parameter file update.

PR #999, opened 2023-03-10
compatability changes for clm bgc

rgknox:
### Description:

These changes are necessary to get FATES compatible with the full BGC call sequence in CLM. Since CLM always has the nitrogen cycle active, with limitation applied by withdolding supplementation, FATES must always present an N litter flux. These changes ensure that in all scenarios (c only, and cnp) we allocated and zero-out the N litter arrays.  In the C-only version of fates, we will present the zero fluxes, in the cnp version of fates, these will have real, positive, values.

Synchronized (API change) with CTSM [PR 1959](https://github.com/ESCOMP/CTSM/pull/1959)

### Collaborators:

@adrifoster @billsacks @wwieder @glemieux @ckoven @rosiealice 

We do not expect qualitative changes in FATES runs, wiht the exception of soil related variables, such as SOM, fragmented litter and heterotrophic respiration. These are expected to change in fates runs. These changes are likely to not propogate into FATES-side results, but slight deviations may emerge.  non-fates simulations, and fates-sp simulations are not expected to change.

### Checklist:

- [ ] FATES PASS/FAIL regression tests were run (TBD)
- [x] If answers were expected to change, evaluation was performed and provided  (see CTSM PR)

### Test Results:

TBD

FATES baseline hash-tag:

Test Output:


glemieux:
@rgknox and I discussed potentially addressing https://github.com/NGEET/fates/issues/697 here as well.

glemieux (approved):

PR #1023, opened 2023-04-24
Add termination mortality to FATES_MORTALITY_CFLUX_PF history variable

JessicaNeedham:
This PR adds termination mortality carbon flux to the history variable FATES_MORTALITY_CFLUX_PF. It also shortens the name of FATES_MORTALITY_HYDRAULIC_CFLUX_PF to FATES_MORTALITY_HYDRO_CLFUX_PF since the former was too long and wasn't being read properly by the HLM (runs compile but then give an error message about the variable not being found). 

The FATES_MORTALITY_CFLUX_PF variable will change. Everything else should be b4b. 

### Checklist:
- [ X] If answers were expected to change, evaluation was performed and provided

### Test Results:
CTSM (or) E3SM (specify which) baseline hash-tag: E3SM b645be3aa2

FATES baseline hash-tag: 1c7ec096

Test Output: 
Main branch compared with branch with termination carbon flux added to total carbon flux. This was a global run with default parameter files - grid cell shown is BCI (lat=9.15, lon=-79.85). 

termfix

### DIFF
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -760,14 +760,13 @@ contains
                                   currentSite%imort_carbonflux(currentCohort%pft) = &
                                        currentSite%imort_carbonflux(currentCohort%pft) + &
                                        (nc%n * ED_val_understorey_death / hlm_freq_day ) * &
-                                       total_c * g_per_kg * days_per_sec * years_per_day * ha_per_m2
+                                       total_c * days_per_sec * years_per_day * ha_per_m2
 
                                   currentSite%imort_abg_flux(currentCohort%size_class, currentCohort%pft) = &
                                        currentSite%imort_abg_flux(currentCohort%size_class, currentCohort%pft) + &
                                        (nc%n * ED_val_understorey_death / hlm_freq_day ) * &
                                        ( (sapw_c + struct_c + store_c) * prt_params%allom_agb_frac(currentCohort%pft) + &
-                                       leaf_c ) * &
-                                       g_per_kg * days_per_sec * years_per_day * ha_per_m2
+                                       leaf_c ) * days_per_sec * years_per_day * ha_per_m2
 
 
                                   ! Step 2:  Apply survivor ship function based on the understory death fraction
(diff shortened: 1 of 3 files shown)


ckoven:
@JessicaNeedham I think the units on the imort_carbonflux variable comment in the current EDTypesMod are wrong?
Here it says it is kgC/ha/day:
https://github.com/NGEET/fates/blob/1c7ec096c519b5447bc4d6820f36c5047bf48bc7/main/EDTypesMod.F90#L849
but here it looks as though it should be gC/m2/s:
https://github.com/NGEET/fates/blob/1c7ec096c519b5447bc4d6820f36c5047bf48bc7/biogeochem/EDPatchDynamicsMod.F90#L760-L763

I think the current usage of it is correct, but we should probably fix the descriptor in EDTypesMod?

Also noting that there are no units listed for `term_carbonflux_canopy` and `term_carbonflux_ustory` here:
https://github.com/NGEET/fates/blob/1c7ec096c519b5447bc4d6820f36c5047bf48bc7/main/EDTypesMod.F90#L845-L848

I think their units should be [kg C / ha / day] based on:
https://github.com/NGEET/fates/blob/1c7ec096c519b5447bc4d6820f36c5047bf48bc7/biogeochem/EDCohortDynamicsMod.F90#L897-L910

and so your use of the variables is correct here, but would be good to get the units listed next to the variable in EDTypesMod.

ckoven (approved):

JessicaNeedham:
@ckoven - thanks, I'll add units for `term_carbonflux_canopy` and `term_carbonflux_ustory`  in EDTypesMod  - agree they should be kg C ha-1 day-1. 

`imort_carbonflux` gets multiplied by g_per_kg in EDPatchDynamics but then divided by g_per_kg in FatesHistoryInterface. 

Can we just delete both of those and make the units in EDTypesMod kg m-2 s-1?

ckoven:
> @ckoven - thanks, I'll add units for `term_carbonflux_canopy` and `term_carbonflux_ustory` in EDTypesMod - agree they should be kg C ha-1 day-1.
> 
> `imort_carbonflux` gets multiplied by g_per_kg in EDPatchDynamics but then divided by g_per_kg in FatesHistoryInterface.
> 
> Can we just delete both of those and make the units in EDTypesMod kg m-2 s-1?

Yeah, good idea.

glemieux:
Regression testing on `cheyenne` is complete.  The tests are b4b against baseline `fates-sci.1.65.4_api.25.4.0-ctsm5.1.dev123`, with the exception of the two tests where `FATES_MORTALITY_CFLUX_PF` is present (`FatesColdAllVars` and `FatesFireLightningPopDens`), which is as expected given the changes.  There are `FIELDLIST` diffs due to the name change to `FATES_MORTALITY_HYDRAULIC_CFLUX_PF` as expected.

Folder location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1023`

PR #1024, opened 2023-05-05
Refactor of patch and cohort objects

adrifoster:
First refactor of patch and cohort objects to facilitate unit testing and decrease circular dependencies

### Description:
I moved the patch and cohort objects into their own modules, and moved some procedures in `FatesCohortDynamicsMod` and `FatesPatchDynamicsMod` into those modes as bound procedures associated with each type.

This also necessitated some modifications of other modules/types to get rid of circular dependencies.

Most notably, I got rid of the pointers from the `cohort` type to their respective `patch`. I did not think this was necessary, and created circular dependencies (the pointer from each `patch` to its `cohort` linked list remains). Where information about the patch was necessary, I converted the procedures to take in patch-specific variables/attributes rather than the entire patch structure itself.

To be coordinated with https://github.com/ESCOMP/CTSM/pull/2000 and https://github.com/E3SM-Project/E3SM/pull/5849

### Collaborators:
@rgknox ; @billsacks ; @glemieux ; @JessicaNeedham 

None - all tests should be B4B

### Checklist:

### Test Results:
see /glade/scratch/afoster/tests_fates_refactor_latest 

all tests are PASS or EXPECTED FAIL

CTSM (or) E3SM (specify which) test hash-tag: `ctsm5.1.dev123-14-g2e8d71075`

CTSM (or) E3SM (specify which) baseline hash-tag: ctsm5.1.dev123

FATES baseline hash-tag: `sci.1.65.4_api.25.4.0`

Test Output:
```
fates_refactor_latest_gnu: 3 tests
    FAIL ERS_Lm13.f10_f10_mg37.I2000Clm50Fates.cheyenne_gnu.clm-FatesCold COMPARE_base_rest (EXPECTED FAILURE)

 
fates_refactor_latest_int: 33 tests
    PEND ERP_P72x2_Ld30.f45_f45_mg37.I2000Clm50FatesRs.cheyenne_intel.clm-mimicsFatesCold SHAREDLIB_BUILD (EXPECTED FAILURE)
    PASS ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdHydro COMPARE_base_rest (UNEXPECTED: expected FAIL)
    FAIL ERS_Lm12.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesFireLightningPopDens COMPARE_base_rest (EXPECTED FAILURE)
    FAIL ERS_Lm13.f45_f45_mg37.I2000Clm50Fates.cheyenne_intel.clm-FatesColdNoComp COMPARE_base_rest (EXPECTED FAILURE)
```


adrifoster:
Apologies that this became somewhat massive. I will make much more incremental changes next time.

glemieux:
Per the fates software meeting today, @adrifoster is investigating where an observed ~20% performance slow down is coming from.

adrifoster:
I re-did testing with the allocatable arrays turned back into static arrays and the timing was back to being similar to baseline.

I am going to update a bit more to match style in the style guide and then re-test.

main/EDTypesMod.F90
```diff
@@ -713,8 +233,8 @@ module EDTypesMod
   type, public :: ed_site_type
      
      ! POINTERS  
-     type (ed_patch_type), pointer :: oldest_patch => null()   ! pointer to oldest patch at the site  
-     type (ed_patch_type), pointer :: youngest_patch => null() ! pointer to yngest patch at the site
```
rgknox, line 717: Even though this is the FATES model, it is built on top of the Ecosystem Demography scaling approach, which at its core is patches and cohorts.  I think it is a good think to maintain the "ed_patch_type" and "ed_cohort_type" here, so that people are aware of its heritage.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -2026,7 +2027,7 @@ subroutine recruitment( currentSite, currentPatch, bc_in )
     !----------------------------------------------------------------------
 
     allocate(temp_cohort) ! create temporary cohort
-    call zero_cohort(temp_cohort)
+    call temp_cohort%ZeroValues()
```
rgknox, line 2030: @adrifoster I've always felt like creating this temporary cohort was kind of an extra unnecessary step that just confuses me each time.  We just use it's data structure's primitives as arguments for the create_cohort routine, which is the routine that generates the real cohort used.  Can we drop the temp_cohort here, and just create a set of the key primitives that are passed to create_cohort() instead?

main/EDTypesMod.F90
adrifoster: Okay, I thought in a FATES dev meeting we had discussed and agreed to change to `fates_` etc. I can change back though. @glemieux @ckoven @rosiealice

biogeochem/EDPhysiologyMod.F90
adrifoster: Yes I agree it's not really serving us right now. Unless we wanted to add some kind of checking ability to the cohort type (via setters or getters, etc.), I think it's fine to just use intrinsic types for all of these values. I can add this change.

rgknox: I saw a similar approach in EDInitMod init_cohorts() too btw

main/FatesConstantsMod.F90
```diff
@@ -30,11 +30,34 @@ module FatesConstantsMod
   ! Integer equivalent of false (in case come compilers dont auto convert)
   integer, parameter, public :: ifalse = 0
 
+  integer , parameter, public :: N_DBH_BINS = 6  ! no. of dbh bins used when comparing patches
```
rgknox, line 33: This seems like a better place for these constants

main/FatesInventoryInitMod.F90
```diff
@@ -94,7 +97,7 @@ module FatesInventoryInitMod
                                                            ! defined in model memory and a physical
                                                            ! site listed in the file
 
-   logical, parameter :: do_inventory_out = .false.
+   logical, parameter :: do_inventory_out = .true.
```
rgknox, line 100: inventory out

biogeochem/EDCohortDynamicsMod.F90
```diff
@@ -150,318 +145,144 @@ Module EDCohortDynamicsMod
...
-       ! This allocates array spaces
-       call InitHydrCohort(currentSite,new_cohort)
-
-       ! zero out the water balance error
-       new_cohort%co_hydr%errh2o = 0._r8
-
-       ! This calculates node heights
-       call UpdatePlantHydrNodes(new_cohort,new_cohort%pft, &
-                                new_cohort%hite,currentSite%si_hydr)
+subroutine create_cohort(currentSite, patchptr, pft, nn, hite, coage, dbh,   &
```
rgknox, line 148: Are there instances where the old "create_cohort()" routine is still required?  I was surprised to see this still here.

adrifoster: the `create_cohort` is still used to set up a cohort on a patch (think of it as a patch-level method, though I haven't moved it to be a type-bound patch method yet). the `cohort%Create` does the initing and setting of values, and is called within `create_cohort`

biogeochem/FatesPatchMod.F90
```diff
@@ -0,0 +1,744 @@
...
+    
+    contains
+
+      procedure :: Init
+      procedure :: NanValues 
+      procedure :: ZeroValues
+      procedure :: InitRunningMeans
+      procedure :: InitLitter
+      procedure :: Create
+      procedure :: FreeMemory
```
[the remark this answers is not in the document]
adrifoster: sure, it just felt redundant to do patch%FreePatchMemory

main/FatesConstantsMod.F90
adrifoster: I would like to go through (in a different PR) and gather the rest of these types of entities into either the constants or parameters modules...

biogeochem/FatesPatchMod.F90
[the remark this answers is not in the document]
rgknox: i've come full circle, I don't think we should change, I like FreeMemory

biogeochem/EDPhysiologyMod.F90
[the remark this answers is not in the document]
adrifoster: @rgknox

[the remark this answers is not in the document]
rgknox: just noting that @adrifoster and I checked in on this yesterday, and noted that since the two if conditions are mutually exclusive, it can be done either way.

main/EDTypesMod.F90
rgknox: after team discussion, I'm dropping this suggestion.  There are better/other ways to recognize the underpinnings of FATES, using the namespace here would be confusing

rgknox (approved):

rgknox:
I've had a good look through these changes and have had some pretty thorough back and forth with @adrifoster offline as well.  These changes are totally awesome and really improve the code's organization

glemieux (changes requested):
A few clean up suggestions and a couple of questions.  Sorry for the delay in reviewing!

main/FatesInventoryInitMod.F90
```diff
@@ -540,11 +549,14 @@ subroutine initialize_sites_by_inventory(nsites,sites,bc_in)
 
          ! If this is flagged as true, the post-fusion inventory will be written to file
          ! in the run directory.
+         
          if(do_inventory_out)then
              call write_inventory_type1(sites(s))
          end if
 
       end do
+      !call canopy_summarization(nsites, sites, bc_in)
```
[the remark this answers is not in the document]
adrifoster: sorry not sure why that is there! probably debugging...thanks for catching it

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -1934,13 +1865,14 @@ subroutine ConstrainRecruitNumber(csite,ccohort, bc_in)
 
   ! Arguments
   type(ed_site_type), intent(inout), target     :: csite
-  type(ed_cohort_type) , intent(inout), target  :: ccohort
+  type(fates_cohort_type) , intent(inout), target  :: ccohort
+  type(fates_patch_type), intent(inout), target       :: cpatch
```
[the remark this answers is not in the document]
adrifoster: The `cpatch` is still used below for litter I think

glemieux:
@adrifoster found during regression testing late last week that the fates spmode tests where not B4B against the baseline.  The culprit appears to have been that in the call to `create_cohort` at the end of `init_cohorts` the refactor was passing ` EDPftvarcon_inst%hgt_min(pft)` as the input height argument for the new cohort during *all* run modes.  Currently on `main`, we set `temp_cohort%hite` directly to either `0.5_r8` or `EDPftvarcon_inst%hgt_min(pft)` depending on if we're in sp mode or note.  This `temp_cohort%hite` value is then passed into `create_cohort` via argument.

Cheyenne is currently down, so I can't run the fix in commit cf41e0a against our typical fates suite regression, but I did run this on lobata comparing the output of a one year long smoke test at BCI against `main` and the results were B4B, so I'm pretty confident this is the fix.

As an aside, this raises the question: for initialization only, is using the `hgt_min` for the given pft a decent enough usage for initializing cohorts during sp mode as well?  Thoughts @rosiealice?

glemieux:
It looks like the issue that I identified is only part of the story: global sp mode tests are still failing against the baseline.  The 1x1_brazil smoke test passes b4b as expected (given that my test for the fix was for a single site).  

Test results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1024-spfix`

glemieux:
It looks like the thing that was causing the non-b4b results against the baseline for the global sp mode runs was the additional call of `leafc_from_treelai`.  The refactor had a call to this function inside the new `calculate_SP_properties` and then again inside `assign_cohort_sp...` after the `calculate_SP_properties` call.  I've removed the additional call via a6c3a74.

This has me thinking that we were seeing as not-b4b prior to my commit is actually more correct.  Right now on main, we adjust `cohort%c_area` and `cohort%n` after the call to `leafc_from_treelai` (as necessary), which uses the former cohort values as input:
https://github.com/NGEET/fates/blob/caf4aae1f6c537ed6e5fb08efcb57a34eab76941/biogeochem/EDPhysiologyMod.F90#L1952-L1979

@rosiealice given the dependency of leaf carbon on cohort area and number, should we actually be waiting to call `leafc_from_treelai` after the adjustments to cohort area and number?

biogeophys/FatesPlantHydraulicsMod.F90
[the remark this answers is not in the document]
glemieux: Oops yep!  Thanks.

glemieux (approved):

glemieux:
Generating a new fates baseline using `ctsm5.1.dev132`.  I'll rerun the full `fates` test suite against this later tonight.  If that looks good, I'll kick off `aux_clm` tests as well.

rosiealice:
Right, I see what you mean. I agree it seems like there is no reason to call the leafc_dfrom_treelai before the error correction of the canopy area here... 
This was all a very strange piece of precision tolerance wrangling. I feel like it would all work much better if the tolerances had some inbuilt bit of canopy gap from which tiny errors could be added and subtracted without it causing havoc. @rgknox this might be easier with the 'ghost element' in the new two stream code is suspect?

rosiealice:
Also responding to the hgt_min question from earlier in this thread, does this mean that the intitializatio value will only operate for perhaps the very first time step (or not even that) before being overwritten with the values from the surface dataset? If so then that is fine...

glemieux:
Regression testing on cheyenne against `fates-sci.1.67.0_api.25.5.0-ctsm5.1.dev132` is complete and all expected tests are B4B.  Testing on izumi and `aux_clm` suite is in progress.

Folder location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1024-leaf_c_fix`

glemieux:
> Also responding to the hgt_min question from earlier in this thread, does this mean that the intitializatio value will only operate for perhaps the very first time step (or not even that) before being overwritten with the values from the surface dataset? If so then that is fine...

That's correct @rosiealice, it would only be during the `init_cohorts` subroutine prior to the first call of `satellite_phenology`.

PR #1032, opened 2023-05-22
FATES LUH2 data curation tool

glemieux:
### Description:
This pull request adds the initial version of the fates luh2 data curation pipeline.  The tool consists of four parts: 1) a module, luh2mod.py, that condenses a majority of the luh2 updates into a set of functions, 2) a python script, luh2.py, that calls the module functions and takes a set of command line arguments, 3) a bash script, luh2.sh, that sets up the expected arguments and calls the python script and 4) a conda environment `yml` file to provide the user with the necessary dependencies to run the tool.

This partially addresses https://github.com/NGEET/fates/issues/450.  The script was heavily adapted from work done by @ckoven.

Raw LUH2 data from https://luh.umd.edu/ is available on cheyenne in the input data directory under `/lnd/clm2/rawdata/LUH2`.

### Collaborators:
@ckoven @adrifoster 

### Expectation of Answer Changes:
No changes as this only impacts tooling

### Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have updated the in-code documentation .AND. (the [technical note](https://github.com/NGEET/fates-docs) .OR. the wiki) accordingly.
- [x] I have read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/master/CONTRIBUTING.md) document.
- [ ] FATES PASS/FAIL regression tests were run
- [ ] If answers were expected to change, evaluation was performed and provided

### Test Results:
Testing not necessary

CTSM (or) E3SM (specify which) test hash-tag:

CTSM (or) E3SM (specify which) baseline hash-tag:

FATES baseline hash-tag:

Test Output:

Issue #964, opened 2022-12-20
Should we add history variable dimensions for land-use-category and land-use-category x pft?


ckoven:
We currently have a few history variables that are specific to secondary forest, such as these:
https://github.com/NGEET/fates/blob/main/main/FatesHistoryInterfaceMod.F90#L5318-L5351

The new PR https://github.com/NGEET/fates/pull/888 adds considerably to that list of variables that apply to secondary lands only, see the new ones here:
https://github.com/NGEET/fates/pull/888/files#diff-03feca0f1df8d3791c0d8f7b6055a0368d5241716d19d6a97c3d227fd6c5ad23R4926-R6904

This seems like something that may start to become a bit unmanageable as we add new land use types in FATES.  So I wanted to suggest that we add a new history variable dimension for these land use categories, which we can then index as 1 = primary lands, 2 = secondary, etc.  That way we can consolidate some of these variables into that dimension and simplify things.  

If we do this, the question is as always where to stop. I can imagine that it would be very useful to add a land-use by PFT dimension too, that way we can keep track of how different PFTs exist and behave differently on different land use categories.  Do we also add land-use x age? etc.  I'm not sure.  I guess my initial instinct would be to just add land-use and land-use x pft and stop there.  That would leave some variables still as separate secondary-forest-only variables (like these two: https://github.com/NGEET/fates/blob/main/main/FatesHistoryInterfaceMod.F90#L5339-L5351), which I think is ok as I think the content of those is pretty specific to secondary lands and wouldn't apply to, e.g., pasture. Likewise some of the new ones from #888. But interested to hear if others have thoughts.

I propose to do this after #888 is merged to main.

sshu88:
@ckoven Thank you for the suggestion. It will be very helpful to have different land-use categories as an additional dimension. I'm thinking more about how we form a reasonable list of this "land-use-category" to be applicable for the longer term when we have more managed patches from different types of managements. How about using disturbance type (harvested, LUC, fire, irrigated, ...) instead of the simple primary vs. secondary tag list. In this case secondary forest will be those patches with disturbance dimension of harvested and LUC.

ckoven:
HI @sshu88 -- right, the idea would be that as we build out more land use categories, we would add to the length of that dimension. Hopefully we can do it in a fairly automatic way so that we don't have to increment anything manually in HLMs as we add a new category.

ckoven:
[...]

PR #1040, opened 2023-06-08
v1 Land Use Change

### LINKED ISSUE #450: Towards FATES w/ land use

ckoven:
This PR is a first version of including land use change other than forest harvest into FATES.  

### Description:
The patch-level land use labels are extended from their prior values of only primary or secondary lands to also include pasture, rangeland, and crops.  Land use change in FATES is treated as a disturbance rate (in area/time units) from existing patches with a given land use label to new patches with the new land use label. The code thus now has four distinct types of disturbances: treefall, fire, logging, and land-use change, any or all of which may occur on a given patch on a given day.  The structure of the overall disturbance routines had to be somewhat changed to accommodate this new logic.

The code uses as its driving data the LUH2 state and transition data available at https://luh.umd.edu.  The code in #1032 regrids this data into a new FATES land use driving dataset; i.e. it is decoupled from the existing code used to drive land use change in big-leaf CLM and ELM; this is because the logic is totally different, instead of translating land use to land cover offline, the land use data is directly ingested during runtime. The five FATES land use categories are aggregated during runtime from a greater number of LUH2 land use types (which include forest/nonforest, and multiple crops). The LUH2 state data is used to initialize the patch structure at the year of model initialization. The LUH2 transition matrix is used to drive land-use-change disturbance rates in FATES.  During land use change disturbance, vegetation is either cleared or not cleared on the newly-disturbed patch based on the specific transition, following rules described in Ma et al (2020) https://doi.org/10.5194/gmd-13-3203-2020.

There are corresponding ELM- and CLM-side changes required to read and pass the LUH2 data to FATES, as well as a few other small things.  This code is currently working with ELM, where a PR will be coming shortly. 
- E3SM pull request: https://github.com/E3SM-Project/E3SM/pull/5760

This code only really handles the basic infrastructure of land use change, subsequent developments that have not yet been written are still needed to handle lots of other things:
1. extending land use change to nocomp logic (where both land use and land cover need to be jointly specified).
2. treatment of a simplified crop PFT on croplands.
3. grazing and/or fire management on pasture, rangeland, and croplands.
4. other management approaches that govern land cover as a function of land use when not using nocomp configurations.
5. consolidating the drivers of logging from their current location in the surface dataset into the FATES-LUH2 dataset (where both area-based and mass-based driving data currently now also exists)

Two new output history dimensions are created: (1) a land-use "`lu`" dimension, and (2) a land-use x land-use "`lulu`" dimension.  The former can be used for looking at any quantity across all land use types (area, biomass, etc); the latter can be used to look at, e.g., a disturbance rate matrix from a given land-use type to a given land-use type.  Either as part of this PR (not yet done) or in a subsequent one, all of the existing history variables that are specific to secondary lands should be moved to the new `lu` dimension (see also https://github.com/NGEET/fates/issues/964, which also proposes a land use x PFT dimension, which would be useful).

fixes #450.
fixes https://github.com/ESCOMP/CTSM/issues/1077.

### Collaborators:
I worked extensively with @glemieux in writing and testing this. Discussions with many people, including @rosiealice @jkshuman @adrifoster @lawrencepj1 @rgknox @jenniferholm @aldivi @sshu88 @JessicaNeedham @dlawrenncar @mpaiao @ekluzek @billsacks 

If land use is enabled, answers will be very different.  If not enabled, roundoff-level changes would seem likely due to the reorganization of the loop structure in the disturbance code.

### Checklist:

### Test Results:
Currently not tested.

FATES baseline hash-tag:

Test Output:


main/EDTypesMod.F90
```diff
@@ -421,12 +420,10 @@ module EDTypesMod
...
-     real(r8) :: disturbance_rates_primary_to_secondary(N_DIST_TYPES)    ! actual disturbance rates from primary patches to secondary patches [m2/m2/day]
-     real(r8) :: disturbance_rates_secondary_to_secondary(N_DIST_TYPES)  ! actual disturbance rates from secondary patches to secondary patches [m2/m2/day]
-     real(r8) :: potential_disturbance_rates(N_DIST_TYPES)               ! "potential" disturb rates (i.e. prior to the "which is most" logic) [m2/m2/day]
-     real(r8) :: primary_land_patchfusion_error                          ! error term in total area of primary patches associated with patch fusion [m2/m2/day]
-     
+     real(r8) :: disturbance_rates(N_DIST_TYPES,n_landuse_cats, n_landuse_cats)  ! actual disturbance rates for each disturbance type  [m2/m2/day]
+     real(r8) :: primary_land_patchfusion_error             ! error term in total area of primary patches associated with patch fusion [m2/m2/day]
+     real(r8) :: landuse_transition_matrix(n_landuse_cats, n_landuse_cats) ! land use transition matrix as read in from HLM and aggregated to FATES land use types [m2/m2/year]
```
glemieux, line 425: @ckoven we currently don't use this site-level variable.  Currently, we are defining the same thing locally in the `disturance_rates` subroutine, which is fine since it is only ever needed in that space:
https://github.com/NGEET/fates/blob/150271047867f54ac83e1569db41fb78b33e1986/biogeochem/EDPatchDynamicsMod.F90#L215
That said, is there a potential future in which we'd want to record the matrix for each site?

ckoven:
Just noting here that this PR is ready for review. So I am pinging some potential reviewers who might be interested: @rosiealice @adrifoster @lawrencepj1 @rgknox @jenniferholm @aldivi @sshu88 @mpaiao. Please feel free to review any or all parts of these code changes. Thanks!

wwieder:
Thanks for working on this, Charlie!

lawrencepj1:
Thanks for the update and gentle push @ckoven . I am pretty flat out this week and most of next but I can have a look for consistency at the end of next week. I haven't been through the FATES code so there may be a delay in getting familiar with the code base as well. Congrats on getting the code to this stage. Peter

aldivi:
@ckoven Conceptually it looks good. Without spending lots of time re-acquainting with the code I can't really comment on the technical aspects of the code itself. I do have questions about the transition rulesets. There are nine coded for selection, but the first one (from Ma) is hardcoded and there does not appear to be a way to select a different one (other than changing the code). Did you just include these for testing or potential future use? Are the other rulesets from other sources or did you just try to cover the different possible permutations?

ckoven:
Currently chatting with @lawrencepj1 about this. Peter notes that we should probably add a tracker for the case where total disturbance rates exceed 100% of the patch area per day, here: https://github.com/NGEET/fates/pull/1040/files?diff=unified&w=1#diff-28c679efbe804905b8b4fac3ccdbb04f6bb3bcd2794fe1a4556d5541e008a750R401-R408

ckoven:
2 other notes from discussion with @lawrencepj1:

1. We should move the currently-hard-coded parameters that define what fraction of woody biomass gets turned into product pools and what fraction of leaf biomass gets burned during any given land-use transition with clearing to the parameter file. We may also want to make them PFT-dependent. See here: https://github.com/NGEET/fates/pull/1040/files#diff-28c679efbe804905b8b4fac3ccdbb04f6bb3bcd2794fe1a4556d5541e008a750R2153-R2154
2. There are two problems with the clearing ruleset logic; one is that there is a logic error for ruleset number one that needs to be fixed, and two is that we should switch the default behavior to the ruleset four (always clear transitions to pasture and crop, never to rangeland) as the simplest case to start with.

ckoven:
One other thing to do here for this PR:
- [ ] consolidate the new parameters that are max patches for each land use type. instead of a bunch of scalar parameters, this would be better as a vector of length equal to the number of land-use types.

lawrencepj1:
Walked through all the Land Use PR code with @ckoven Everything looks good to me. Glad to have an explanation and better insight into the way FATES is implementing Land Use. Peter

rgknox:
@ckoven I noticed that the logging mortality routines did not extend any logic to cover new land-cover types. Can you confirm this is not necessary? My guess that very little logging (if any) would come from anything but primary or secondary forests, but I suppose it is possible.  Also, will we or do we need to incorporate the tracking of wood products as a result of conversion (where we don't apply the logging module)?

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -69,16 +73,16 @@ module EDPatchDynamicsMod
...
   use FatesAllometryMod    , only : carea_allom
   use FatesAllometryMod    , only : set_root_fraction
   use FatesConstantsMod    , only : g_per_kg
   use FatesConstantsMod    , only : ha_per_m2
   use FatesConstantsMod    , only : days_per_sec
   use FatesConstantsMod    , only : years_per_day
   use FatesConstantsMod    , only : nearzero
-  use FatesConstantsMod    , only : primaryforest, secondaryforest
-  use FatesConstantsMod    , only : n_anthro_disturbance_categories
+  use FatesConstantsMod    , only : primaryland, secondaryland, pastureland, rangeland, cropland
```
rgknox, line 83: good idea to change this from forest to land

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -0,0 +1,307 @@
...
+    urban_fraction = 0._r8
+
+    use_luh_if: if ( hlm_use_luh .eq. itrue ) then
+
+       ! Check the LUH data incoming to see if any of the transitions are NaN
+       temp_vector = bc_in%hlm_luh_transitions
+       call CheckLUHData(temp_vector,modified_flag)
+       if (.not. modified_flag) then
+          ! identify urban fraction so that it can be factored into the land use state output
+          urban_fraction = bc_in%hlm_luh_states(findloc(bc_in%hlm_luh_state_names,'urban',dim=1))
```
rgknox, line 93: This is a nice feature, very self describing.

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -0,0 +1,307 @@
...
+       end if
+
+       !!TODO: may need some logic here to ask whether or not ot perform land use change on this timestep. current code occurs every day.
+       !!If not doing transition every day, need to update units.
+
+       transitions_loop: do i_luh2_transitions = 1, hlm_num_luh2_transitions
+
+          ! transition names are written in form xxxxx_to_yyyyy where x and y are donor and receiver state names
+          transition_name = bc_in%hlm_luh_transition_names(i_luh2_transitions)
+          donor_name = transition_name(1:5)
```
rgknox, line 103: This seems like vulnerable code, thinking about how to fix it though.

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -0,0 +1,307 @@
...
+       call CheckLUHData(temp_vector,modified_flag)
+       if (.not. modified_flag) then
+          ! identify urban fraction so that it can be factored into the land use state output
+          urban_fraction = bc_in%hlm_luh_states(findloc(bc_in%hlm_luh_state_names,'urban',dim=1))
+       end if
+
+       !!TODO: may need some logic here to ask whether or not ot perform land use change on this timestep. current code occurs every day.
+       !!If not doing transition every day, need to update units.
+
+       transitions_loop: do i_luh2_transitions = 1, hlm_num_luh2_transitions
```
rgknox, line 99: is hlm_num_luh2_transitions set by the user, or is it directly interpreted by getting the dimensions from the input transition file?  It seems like it would be more robust to diagnose this value when needed instead of passing it in.

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -0,0 +1,307 @@
...
+    type(bc_in_type) , intent(in) :: bc_in
+    real(r8), intent(inout) :: landuse_transition_matrix(n_landuse_cats, n_landuse_cats)  ! [m2/m2/day]
+
+    ! !LOCAL VARIABLES:
+    type(luh2_fates_lutype_map) :: lumap
+    integer :: i_donor, i_receiver, i_luh2_transitions, i_luh2_states
+    character(5) :: donor_name, receiver_name
+    character(14) :: transition_name
+    real(r8) :: urban_fraction
+    real(r8) :: temp_vector(hlm_num_luh2_transitions)
```
rgknox, line 79: why the use of "temp_vector"? Its not modified from the bc_in structure.

rgknox: strike that, the check routine does modify it

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -0,0 +1,307 @@
...
+    logical, intent(out)    :: modified_flag
+
+    ! Check to see if the incoming luh2 vector is NaN.
+    ! This suggests that there is a discepency where the HLM and LUH2 states
+    ! there is vegetated ground. E.g. LUH2 data is missing for glacier-margin regions such as Antarctica.
+    ! In this case, states should be Nan.  If so,
+    ! set the current state to be all primary forest, and all transitions to be zero.
+    ! If only a portion of the vector is NaN, there is something  amiss with
+    ! the data, so end the run.
+
```
rgknox, line 288: @ckoven I can't tell if there are situations here where we should be gracefully failing, instead of pushing on.  Can you confirm that the only possibility here is glacier intervention? Is this glacier retreat?

It seems like if we have a nan, something is wrong, no?  Are there any scenarios where a nan would be worth aborting? Like what if we have a precedent on previous steps with valid transitions?

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -0,0 +1,307 @@
...
+    modified_flag = .false.
+    if (all(isnan(luh_vector))) then
+       luh_vector(:) = 0._r8
+       ! Check if this is a state vector, otherwise leave transitions as zero
+       if (size(luh_vector) .eq. hlm_num_luh2_states) then
```
rgknox, line 293: Checked this through, considered what to do with states versus transitions, and I follow your logic and the indexing.

biogeochem/FatesLandUseChangeMod.F90
ckoven: The issue is just that there are places where luh2 doesn't cover, e.g., all of Antarctica even though not all of Antarctica is ice-covered. So we don't want it to fail there, we just want to not use the land use data, and instead just assume that there is no land use at that point (and thus everything is primary). I think the subsequent logic handles the case if something has gone wrong somehow in the luh2 data for a given site and it is NaN somewhere but not everywhere.

ckoven:
> @ckoven I noticed that the logging mortality routines did not extend any logic to cover new land-cover types. Can you confirm this is not necessary? My guess that very little logging (if any) would come from anything but primary or secondary forests, but I suppose it is possible. Also, will we or do we need to incorporate the tracking of wood products as a result of conversion (where we don't apply the logging module)?

It's a good point, and something worth digging into in a future version. Right now (i.e., in LUH2), all wood harvest is assumed to occur on either primary or secondary land. In reality, grazing and harvest can of course occur on the same land. Perhaps LUH3 will include wood harvest from other land-use types? But I think we should leave as-is for now.

Yes, this PR adds the wood products resulting from conversion into the same pool as other harvest products. In the next step of this, I'm planning to separate out the products into distinct harvest and land-use-change product pools and with their own parameters (e.g., as discussed in #1084).

main/FatesInterfaceMod.F90
```diff
@@ -547,6 +557,14 @@ subroutine allocate_bcin(bc_in, nlevsoil_in, nlevdecomp_in, num_lu_harvest_cats,
 
       allocate(bc_in%pft_areafrac(natpft_lb:natpft_ub))
 
+      ! LUH2 state and transition data
+      if (hlm_use_luh .gt. 0) then
```
rgknox, line 561: .eq. itrue

biogeochem/FatesPatchMod.F90
```diff
@@ -169,10 +170,12 @@ module FatesPatchMod
...
    ! DISTURBANCE 
-    real(r8) :: disturbance_rates(n_dist_types) ! disturbance rate [0-1/day] from 1) mortality 
-                                                !                                 2) fire 
-                                                !                                 3) logging mortatliy
-    real(r8) :: fract_ldist_not_harvested       ! fraction of logged area that is canopy trees that weren't harvested [0-1]
+    real(r8) :: disturbance_rates(n_dist_types)           ! disturbance rate [0-1/day] from 1) mortality
+                                                          !                                 2) fire
+                                                          !                                 3) logging mortatliy
+                                                          !                                 4) land use change
+    real(r8) :: landuse_transition_rates(n_landuse_cats)  ! land use tranision rate
```
rgknox, line 177: ideally these should be dynamically allocated based on the number of cats defined in the dataset, or perhaps the number of cats that we want to remember based on the combining or grouping of cats in the dataset.

parameter_files/fates_params_default.cdl
```diff
@@ -668,6 +672,9 @@ variables:
 	double fates_frag_cwd_frac(fates_NCWD) ;
 		fates_frag_cwd_frac:units = "fraction" ;
 		fates_frag_cwd_frac:long_name = "fraction of woody (bdead+bsw) biomass destined for CWD pool" ;
+	double fates_maxpatches_by_landuse(fates_landuseclass) ;
+		fates_maxpatches_by_landuse:units = "count" ;
+		fates_maxpatches_by_landuse:long_name = "maximum number of patches per site on each land use type" ;
```
rgknox, line 677: maybe we should add that primary is index 1 here?

parameter_files/fates_params_default.cdl
```diff
@@ -930,6 +931,13 @@ data:
   "sapwood",
   "structure" ;
 
+ fates_landuseclass_name =
+  "primaryland",
```
rgknox, line 935: I like having these labels. I suppose that a user who's only interested in primary lands will be able to reference this list to understand that they only need to set maxpatch for index 1.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -266,13 +270,22 @@ subroutine disturbance_rates( site_in, bc_in)
```
rgknox, line 270: We should have had comments explaining what this is and in what situations it was used. We missed this the first time around.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -295,6 +308,15 @@ subroutine disturbance_rates( site_in, bc_in)
        currentPatch%disturbance_rates(dtype_ifire) = 0.0_r8
 
        dist_rate_ldist_notharvested = 0.0_r8
+
+       ! Avoid this calculation to avoid NaN due to division by zero result if luh is not used
+       if (hlm_use_luh .eq. itrue) then
+          currentPatch%landuse_transition_rates(1:n_landuse_cats) = min(1._r8, &
```
rgknox, line 314: @ckoven do you have a guide, or documentation on how the units of tracking these area changes flows from the datasets into the patch structures?  When dealing with fractions of fractions of physical area, things get hard to follow.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -98,11 +102,9 @@ module EDPatchDynamicsMod
   use SFParamsMod,            only : SF_VAL_CWD_FRAC
   use EDParamsMod,            only : logging_event_code
   use EDParamsMod,            only : logging_export_frac
+  use EDParamsMod,            only : maxpatches_by_landuse
   use FatesRunningMeanMod,    only : ema_sdlng_mdd
   use FatesRunningMeanMod,    only : ema_sdlng_emerg_h2o, ema_sdlng_mort_par, ema_sdlng2sap_par
-  use EDParamsMod,            only : maxpatch_primary
-  use EDParamsMod,            only : maxpatch_secondary
-  use EDParamsMod,            only : maxpatch_total
   use FatesRunningMeanMod,    only : ema_24hr, fixed_24hr, ema_lpa, ema_longterm
```
rgknox, line 109: This routine in general, is getting out of hand regards to comments and transparency.  We have blocks of code that have no comments explaining what is about to be performed.  This may had not been a big deal in the past when it performed a small number of features, but its long enough now that we need more explanations.

biogeochem/EDPatchDynamicsMod.F90
```diff
@@ -411,28 +430,30 @@ subroutine spawn_patches( currentSite, bc_in)
...
     ! !ARGUMENTS:
     type (ed_site_type), intent(inout) :: currentSite
     type (bc_in_type), intent(in)      :: bc_in
     !
     ! !LOCAL VARIABLES:
-    type (fates_patch_type) , pointer :: new_patch
-    type (fates_patch_type) , pointer :: new_patch_primary
-    type (fates_patch_type) , pointer :: new_patch_secondary
+    type (fates_patch_type) , pointer :: newPatch
+    ! type (fates_patch_type) , pointer :: new_patch_primary
```
rgknox, line 446: remove comment

biogeochem/EDPatchDynamicsMod.F90
ckoven: @rgknox I added a bunch of comments to the spawn_patches subroutine to try to better orient the reader on what is going on there and why.

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -0,0 +1,307 @@
...
+  !----------------------------------------------------------------------------------------------------
+
+  function GetLUCategoryFromStateName(this, state_name) result(landuse_category)
+
+    class(luh2_fates_lutype_map) :: this
+    character(len=5), intent(in) :: state_name
+    integer :: landuse_category
+
+    landuse_category = this%landuse_categories(findloc(this%state_names,state_name,dim=1))
+
```
rgknox, line 127: If bogus names are passed in, findloc will return a zero:
https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-1/findloc.html
So it seems impossible that the model will chug along if there is not agreement between fates definitions and the host/dataset definitions of the category names.

biogeochem/FatesLandUseChangeMod.F90
rgknox: After looking at how '%GetIndex' works, the model will crash if the LU names do not match between the host/dataset registry of names and the fates registry of names. resolving.

biogeochem/EDPatchDynamicsMod.F90
rgknox: @ckoven added comments and generally improved readability in EDPatchDynamics, resolving

biogeochem/FatesLandUseChangeMod.F90
rgknox: I'm not going to push hard for this change, but I think ideally we would have a graceful check somewhere that would provide an informative message that the fates and host/dataset landclass names do not match up here.  At least the model will not continue to run using mismatched classes.

rgknox (approved):

main/EDTypesMod.F90
ckoven: In subsequent developments, I am now using this site-level `landuse_transition_matrix` variable for history output, so let's leave it in this PR even though it is not yet being used here.

biogeochem/FatesLandUseChangeMod.F90
glemieux: Thanks @rgknox, this is a good point.  Added.

glemieux:
I've implemented a landuse testmod and I'm seeing a failure in the CNBalanceCheckMod.F90 during the first `hist_htapes_wrapup` call:
```
2928  clm: leaving fates model           1          11
2929  --WARNING-- skipping CN balance check for first timesteps after startup or data
2930   assimilation
2931  --WARNING-- skipping CN balance check for first timesteps after startup or data
2932   assimilation
2933  --WARNING-- skipping CN balance check for first timesteps after startup or data
2934   assimilation
2935  cbalance warning at c =           4  5.266753678508790E-007
2936    911.953104391837
2937  cbalance warning at c =           5  2.277429033081813E-006
2938    911.974234598625
2939  cbalance warning at c =           6  1.683174109460076E-006
2940    911.967568518515
2941  cbalance warning at c =           7  1.537941879714219E-006
2942    911.957395544391
2943  column cbalance error    =   1.537941879714219E-006           7
2944  is fates column?         =  T
2945  Latdeg,Londeg=   30.0000000000000        260.000000000000
2946  begcb                    =    911.965299656670
2947  endcb                    =    911.957395544391
2948  delta store              =  -7.904112279447872E-003
2949  --- Inputs ---
2950  fates litter_flux        =   0.000000000000000E+000
2951  fates wood product flux  =   0.000000000000000E+000
2952  --- Outputs ---
2953  hr                       =   7.904112279464640E-003
2954  -1*som_c_leached         =  -3.964367563847151E-014
2955 iam = 0: local  column   index = 7
2956 iam = 0: global column   index = 5224
2957 iam = 0: global landunit index = 1953
2958 iam = 0: global gridcell index = 865
2959 iam = 0: gridcell longitude    =  260.0000000
2960 iam = 0: gridcell latitude     =   30.0000000
2961 iam = 0: column   type         = 1
2962 iam = 0: landunit type         = 1
2963  ENDRUN:
2964  ERROR: ERROR in CNBalanceCheckMod.F90 at line 391
```

glemieux:
Reviewing the carbon balance check for elm and clm, I found some inconsistency between the two HLMs.  Discussing this with @rgknox, I've removed the wood products fluxes from the carbon balance check for both host land models in the respective HLM-side PRs.  This results in the ctsm-fates landuse testmod passing.  I've got a longer-term case running to check beyond the typical regression test length, which is usually no more than a year or so.

glemieux:
Regression testing the fates suite on cheyenne is partially B4B with the following exceptions:
```
    FAIL ERS_Lm13.f10_f10_mg37.I2000Clm50Fates.cheyenne_gnu.clm-FatesCold BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL SMS_Lm13.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_gnu.clm-FatesCold BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdLandUse BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdNoComp BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdNoCompFixedBioGeo BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL ERS_Ld60.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesColdLogging BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL ERS_Lm12.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesFireLightningPopDens BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL ERS_Lm13.f45_f45_mg37.I2000Clm50Fates.cheyenne_intel.clm-FatesColdNoComp BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL ERS_P144x1_Lm25.f10_f10_mg37.I2000Clm51Fates.cheyenne_intel.clm-FatesColdNoComp BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL SMS_Lm13.1x1_brazil.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-FatesCold BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
    FAIL SMS_Lm6.f45_f45_mg37.I2000Clm50FatesCruRsGs.cheyenne_intel.clm-Fates BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev155: DIFF
```
I'm still waiting on one test that is stuck in the queue.  

These appear to be associated with testmods that activate disturbances and/or are long enough to elicit changes due to fire disturbance.  The differences are very small at round-off level.  This makes sense given that this PR changes the disturbance code as noted in the PR description.

glemieux:
Regression testing the `fates` suite on izumi has completed.  All expected tests pass with only `FIELDLIST` differences as expected due to the history output changes.  One minor note is that I had to submit the new `FatesColdLUH2` testmod twice to get the test to complete without an MPI timeout failure (similar to the one seen in ESCOMP/CTSM#1317).  

Retesting on derecho is underway.

glemieux:
Testing on derecho is complete and results are similar to the previous results on cheyenne with small initial DIFFs showing up for the longer testmods:
```
pr2079-fates-2_gnu: 9 tests
    FAIL ERS_Lm13.f10_f10_mg37.I2000Clm50Fates.derecho_gnu.clm-FatesCold BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF
    FAIL SMS_Lm13.1x1_brazil.I2000Clm50FatesCruRsGs.derecho_gnu.clm-FatesCold BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF

 
pr2079-fates-2_int: 34 tests
    FAIL ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdNoCompFixedBioGeo BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF
    FAIL ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdNoComp BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF
    FAIL ERS_Ld60.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLogging BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF
    FAIL ERS_Lm12.1x1_brazil.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesFireLightningPopDens BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF
    FAIL ERS_Lm13.f45_f45_mg37.I2000Clm50Fates.derecho_intel.clm-FatesColdNoComp BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF
    FAIL ERS_P128x1_Lm25.f10_f10_mg37.I2000Clm51Fates.derecho_intel.clm-FatesColdNoComp BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF
    FAIL SMS_Lm13.1x1_brazil.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesCold BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF
    FAIL SMS_Lm6.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-Fates BASELINE fates-sci.1.69.0_api.31.0.0-ctsm5.1.dev159: DIFF

```

All other diffs are expected `FIELDLIST` differences and otherwise b4b.

Derecho location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr2079-fates-2`

PR #1043, opened 2023-06-21
fix typo in tools/FatesPFTIndexSwapper.py

ckoven:
Tiny bugfix to correct a typo in tools/FatesPFTIndexSwapper.py that is preventing an interpretable error message to be written.

Right now if you call `FatesPFTIndexSwapper.py` without the argument `pft-indices`, instead of giving an error message saying that that argument needs to be set, it crashes with an error message that says that variable `donor_pft_indices_str` is not set. This change fixes that, so it should give a clearer / more actionable error message.

### Description:
fix one typo in the line that puts in an initial value into a variable

### Collaborators:
Mingjie Shi discovered this issue.

None

### Checklist:

### Test Results:
Not tested, but only affects python tool, which isn't part of test suite.

### DIFF
--- a/tools/FatesPFTIndexSwapper.py
+++ b/tools/FatesPFTIndexSwapper.py
@@ -85,7 +85,7 @@ def interp_args(argv):
     input_fname = "none"
     output_fname = "none"
     donor_pft_indices = -9
-    donot_pft_indices_str = ''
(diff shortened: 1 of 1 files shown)


rgknox (approved):

glemieux:
Tested the fix and this works as expected.

PR #1072, opened 2023-08-26
Fix error with large size initialisation in nocomp mode

### LINKED ISSUE #1071: nocomp large tree initialisation failing
I was looking at this again ahead of updating the docs (https://github.com/NGEET/fates-docs/issues/47) and a short test failed to run with the error: 

demotion did not trim area within tolerance                                                                                           
 arealayer:   505.48050811974696                                                                                                        
 patch%area:   482.72740789018195                                                                                                       
 ilayer:            1                                                                                                                   
 bias:   22.753100229565007                                                                                                             
 rel bias:   4.5012814270921124E-002                                                                                                    
 demote_area:   22.753100229565007                                                                                                      
 spread:    5.0000000000000003E-002                                                                                                     
 ENDRUN:ERROR in /global/u2/j/jneedham/E3SM_latest/E3SM/components/elm/src/external_models/fates/biogeochem/EDCanopyStructureMod.F90 at\
 line 783 

The problem was that crown area is calculated with site spread set to init_spread_inventory (0) giving a total canopy layer that exactly matches the patch area. But then in canopy_spread site spread is increased - I think from bare ground patches. In the following call to canopy_structure the canopy area is greater than patch area because of the larger site spread. DemoteFromLayers should sort it out - but it was failing because cohort height wasn’t set. I’ve added a call to h_allom and also ensured that site spread is set to 0 if any of the pfts have a large size initialisation. There is the possibility that some pfts are being initialised by density and others by dbh, in which case site spread will be incorrect for some of them - sort of related to discussion: https://github.com/NGEET/fates/discussions/990.

JessicaNeedham:
Fixes #1071. Initialisation by dbh was failing in no comp mode because height was not being  initialised. This PR adds a call to h_allom so that trees can be demoted in canopy_structure if site spread has been updated in the first time step. 

nocomop initialisation by dbh should now run, rather than result in an error. 

### Checklist:

### Test Results:

FATES baseline hash-tag:

Test Output:

Compiles and runs on Perlmutter with https://github.com/glemieux/E3SM/tree/lnd/fates-refactor E3SM branch and latest FATES commit.

### DIFF
--- a/main/EDInitMod.F90
+++ b/main/EDInitMod.F90
@@ -61,6 +61,7 @@ module EDInitMod
   use FatesInterfaceTypesMod         , only : nlevage
 
   use FatesAllometryMod         , only : h2d_allom
+  use FatesAllometryMod         , only : h_allom
   use FatesAllometryMod         , only : bagw_allom
   use FatesAllometryMod         , only : bbgw_allom
   use FatesAllometryMod         , only : bleaf
@@ -851,6 +852,14 @@ contains
       patch_in%tallest  => null()
       patch_in%shortest => null()
 
+      ! if any pfts are starting with large  size  then the whole  site needs a spread of 0
+      do pft = 1, numpft
+         if (EDPftvarcon_inst%initd(pft) < 0.0_r8) then   
+            site_in%spread = init_spread_inventory
+         end if
+      end do
+
+      
       ! Manage interactions of fixed biogeog (site level filter) and nocomp (patch level filter)
       ! Need to cover all potential biogeog x nocomp combinations
       ! 1. biogeog = false. nocomp = false: all PFTs on (DEFAULT)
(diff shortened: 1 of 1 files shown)


rgknox (approved):
Looks good, makes sense.

Stretch goal: In the call to create_cohort, we've already updated that named argument to convey a zero aged cohort.  How about that integer value of 1 that we pass. I think it is canopy layer. Perhaps we should use a named constant there too.  Actually, I'm surprised we pass anything in, we shouldn't be defining the layer upon create cohort, that is an emergent phenomonena. It almost seems better to pass nothing and set it to uninitialized.  I'll make an issue instead.

rgknox:
Here is the new issue: https://github.com/NGEET/fates/issues/1082
So my suggestion of changing canopy layer to the create_patch call is out of scope and I see nothing else to do on this PR.

rgknox:
running tests:  /glade/scratch/rgknox/tests_0928-180603ch

Issue #1062, opened 2023-08-15
Clean up inventory initialization files


adrifoster:
Right now the inventory initialization files have a lot of columns that are not used by FATES. This (anecdotally) results in some confusion for new users trying to make init files.

I understand the files are the way they are now so that we can maintain compatibility with ED init files. However, I propose that we clean up the files and make them easier to use, and then we can just write a python tool that will convert an ED init file into a FATES init file so that we can maintain this compatibility. We won't really be able to go from a FATES file to an ED file since the ED file needs more information, but I think this is okay.

glemieux:
Closing per #1104

PR #1104, opened 2023-10-19
Clean inventory initialization files

JessicaNeedham:
This PR address issue #1062  and removes unnecessary columns from the inventory initialisation files. As  discussed in #1062 these extra columns are confusing to new users. As part of work on the NGEET FATES tutorial I want to simplify running with inventory data. 

Specifically, I removed water, fsc (fast soil carbon), stsc  (structural soil carbon), stsl (structural soil lignin), ssc (slow soil carbon), psc (passive soil carbon), msn (mineralized soil nitrogen), and fsn (fast soil nitrogen) from the patch file. From the cohort file I removed avgRG (average radial growth), balive (live biomass per individual of the cohort), bdead (the dead biomass per indiv of the cohort), cohort (unused string) and height. 

I was a bit unsure of height -  the code currently doesn’t use it and instead calculates height from DBH. It would be possible to update it to handle a case where the file only had height, and use the height to calculate DBH. But I’m not sure how many sites would have only height - maybe remote sensing data? I’ve removed it for now but could add it back in if people think it might be useful. 

As suggested by @adrifoster  I made a python script that takes an ED file (pss or css) and creates a new file in the new FATES format. The new file will have _fates appended to the file name.

### Collaborators:
@adrifoster identified the issue and suggested a path forward. Discussed with @glemieux. 

This should be bfb in runs with inventory initialisation both on and off as the only information removed from the inventory files was not being used anyway. 

### Checklist

*Contributor*
- [X] The in-code documentation has been updated with descriptive comments

*Integrator*

### Documentation

- [Technical Note](https://github.com/NGEET/fates-docs) update:  no updates needed
- [User's Guide](https://github.com/NGEET/fates-users-guide) update:   https://github.com/NGEET/fates-users-guide/pull/48

### Test Results:

Test suite  not run but  code compiles and runs with inventory init on Perlmutter. 

*CTSM (or) E3SM (specify which) test hash-tag:* b8be65d

*FATES baseline hash-tag:* a3048a6

*Test Output:*

### DIFF
--- a/main/FatesInventoryInitMod.F90
+++ b/main/FatesInventoryInitMod.F90
@@ -168,10 +168,12 @@ contains
       character(len=patchname_strlen), allocatable :: patch_name_vec(:)    ! vector of patch ID strings
       real(r8)                                     :: basal_area_postf     ! basal area before fusion (m2/ha)
       real(r8)                                     :: basal_area_pref      ! basal area after fusion (m2/ha)
-      real(r8)                                     :: min_patch_age
-      real(r8)                                     :: max_patch_age
-      real(r8)                                     :: min_cohort_dbh
-      real(r8)                                     :: max_cohort_dbh
+      real(r8)                                     :: n_pref
+      real(r8)                                     :: n_postf
+ !     real(r8)                                     :: min_patch_age
+!      real(r8)                                     :: max_patch_age
+!      real(r8)                                     :: min_cohort_dbh
+!     real(r8)                                     :: max_cohort_dbh
       
 
       ! I. Load the inventory list file, do some file handle checks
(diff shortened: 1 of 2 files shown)


JessicaNeedham:
After discussion at the software meeting today I have added height back into the cohort file. It isn't currently used but might be useful in future cases where  people use lidar to initialise a run.

JoshuaRady:
@JessicaNeedham,
I agrees with the decision to retain the height despite the potential for confusion (which is real).  An additional reason to keep it is that small trees are sometimes measured by height.

main/FatesInventoryInitMod.F90
```diff
@@ -938,13 +915,12 @@ subroutine set_inventory_edcohort_type1(csite,bc_in,css_file_unit,npatches, &
       integer,  parameter :: recruitstatus = 0
 
      
-      read(css_file_unit,fmt=*,iostat=ios) c_time, p_name, c_name, c_dbh, c_height, &
-            c_pft, c_nplant, c_bdead, c_balive, c_avgRG
+      read(css_file_unit,fmt=*,iostat=ios) c_time, p_name, c_dbh, &
+            c_height, c_pft, c_nplant
```
mpaiao, line 919: I guess c_name is not really critical, but it may be a good idea to keep it as a way of indexing the cohorts to the original trees in an inventory (e.g., the tree tag number).

main/FatesInventoryInitMod.F90
```diff
@@ -1267,8 +1243,8 @@ subroutine write_inventory_type1(currentSite)
        open(unit=pss_file_out,file=trim(pss_name_out), status='UNKNOWN',action='WRITE',form='FORMATTED')
        open(unit=css_file_out,file=trim(css_name_out), status='UNKNOWN',action='WRITE',form='FORMATTED')
 
-       write(pss_file_out,*) 'time patch trk age area water fsc stsc stsl ssc psc msn fsn'
-       write(css_file_out,*) 'time patch cohort dbh height pft nplant bdead alive Avgrg'
+       write(pss_file_out,*) 'time patch trk age area'
+       write(css_file_out,*) 'time patch cohort dbh height pft nplant'
```
mpaiao, line 1247: I guess if  we decide to delete `c_name` from above, then we ought to delete `cohort` from here.

mpaiao: The review options don't let me comment on other lines, but we'd need to change the additional lines of code below:

```fortran
-           write(pss_file_out,*) '0000 ',trim(patch_str),' 2 ',currentPatch%age,currentPatch%area/AREA, &
-                 '0.0000    0.0000    0.0000    0.0000    0.0000    0.0000    0.0000    0.0000'
+           write(pss_file_out,*) '0000 ',trim(patch_str),' 2 ',currentPatch%age,currentPatch%area/AREA
```

and 

```fortran
-               write(cohort_str,'(A7,i4.4,A)') '<coh_',icohort,'>'
-               write(css_file_out,*) '0000 ',trim(patch_str),' ',trim(cohort_str), &
-                     currentCohort%dbh,0.0,currentCohort%pft,currentCohort%n/currentPatch%area,0.0,0.0,0.0
+               write(css_file_out,*) '0000 ',trim(patch_str),' ',&
+                     currentCohort%dbh,0.0,currentCohort%pft,currentCohort%n/currentPatch%area

```

(or keeping `cohort_str` if we decide to keep the cohort index).

tools/ed2_to_fates_inventory_init.py
```diff
@@ -0,0 +1,49 @@
...
+    dsin = pd.read_csv(args.fnamein, delim_whitespace=True)
+
+    # if patch file delete unnecessary patch columns
+    if filetype == 'pss' :
+        keep_col = ['time', 'patch', 'trk', 'age', 'area']
+        newds = dsin[keep_col]
+        
+
+    # if cohort file delete unnecessary cohort columns
+    elif filetype == 'css' :
```
mpaiao, line 35: Flagging here in case we decide to keep the cohort indexing.

mpaiao (changes requested):
@JessicaNeedham Thanks for working on the simpler version of the initialization files! I went through your changes, and they look good, I just pointed out to another place that may require some code changes too. 

My only question is whether to or not to keep the cohort name / cohort index in text file. That can be useful for tracking the source of the cohort, but it is not really critical for FATES.

JessicaNeedham:
Thanks @rgknox for adding in functionality to initialise with a cohort file that has only height but no dbh measurements. Now either height or dbh needs to be negative in the cohort file, and the positive one will be used to calculate the other. 

It works fine in my testing except in a specific case where I set dbh negative and all the height values to be identical. In that case the age, and the size structure of all the patches were equal, and somehow the sum of all patch areas was decreased during patch sorting (somewhere [here](https://github.com/JessicaNeedham/fates/blob/6dfe90a595a0521bb9f08dabe2a834b6144b2a9b/main/FatesInventoryInitMod.F90#L406)) triggering fails in update_hlm_dynamics. 

If people want to initialise a plantation with all trees the same size and all patches the same age, they could have one big patch. But I think FATES should handle a case where someone uses subplot info for patches and tree info for cohorts. I wonder about putting a check here, so that the original order in which patches are read in is used if patches are identical. The patches are fused right after the cohorts are fused, so it shouldn’t matter. Does that sound reasonable @rgknox and  @mpaiao ?

rgknox:
"If people want to initialize a plantation with all trees the same size and all patches the same age, they could have one big patch. But I think FATES should handle a case where someone uses subplot info for patches and tree info for cohorts."  

So in the later case, the different subplots would have trees of the same size, and thus the same demographic as the other subplots?

main/FatesInventoryInitMod.F90
JessicaNeedham: My preference is to remove columns not used. If inventory data changes then people can remake the FATES files which wouldn't require indexing by tree tag. But I don't feel too strongly about this. I could put it back if people would miss it.

main/FatesInventoryInitMod.F90
JessicaNeedham: Oh, good catch! I've removed the unused columns from the write sections. Thanks @mpaiao !

JessicaNeedham:
I ran some 10 year tests using BCI inventory initialisation data - both dbh, and then calculating height from dbh (outside of FATES) and using height to initialise. Also a 10 year run with all  cohorts starting at 5 m hight. Results are [here](https://github.com/JessicaNeedham/FATES-plotting-scripts/blob/main/Inventory_init_check.ipynb). They mostly look as expected except that the two runs using BCI data (from height and dbh) are slightly different. Maybe we wouldn't expect them to be bfb?

mpaiao (approved):

JoshuaRady:
I disagree with the requirement that only DBH _or_ height be used and that DBH or height be negative as the way to select this behavior.

Inventory initialization should handle real inventories, which may contain DBH and/or heights for different trees.  It is common for large trees to be missing heights and for small trees to be missing DBHs.  The better behavior would be to use a preferred measure (probably DBH by default) and use the other when it is missing.  A setting parameter could be provided to choose the primary measure.  Otherwise one missing DBH/height in an entire dataset prevents its accurate import.

rgknox:
@JoshuaRady , I believe we have it setup in the way you desire.  Any plant can be described by dbh or height, it does not need to be consistent throughout the dataset.  As long as, for each plant or cohort, one and only one of either dbh or height is positive.  But for each plant or cohort in the dataeset, it must be one of the two, not both. My motivation for promoting a negative to specify a not-applicable instead of a separate field, is that we may want to use this dataset convention for large grids on HPC machines, so there would potentially be lots of reading and writing involved, and thus cut down on any absolutely unnecessary fields.

JoshuaRady:
@rgknox, thanks for the clarification.  That seems like a reasonable compromise.  I was only considering the use of inventory initialization for single points but if it is to be used on larger scales that makes sense.  The cost of adding the needed logic when converting inventory data into the FATES import format will be small so I can get on board with this.

glemieux:
Regression testing on cheyenne is underway.

glemieux:
All expected regression tests are returning B4B against `fates-sci.1.68.3_api.31.0.0-ctsm5.1.dev153`.

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1104-fates`

PR #1107, opened 2023-10-26
[WIP] Landuse x PFT static data tool

glemieux:
Tool chain for creating the landuse x pft static input dataset

### Description:
This PR adds the `landusepft` tool to the fates tool directory.  This tool has been developed with a [project package structure](https://packaging.python.org/en/latest/tutorials/packaging-projects/).  It also includes unit tests for the tool functions using the [`pytest`](https://docs.pytest.org/en/7.1.x/index.html) package.

### Collaborators:
@ckoven 

### Expectation of Answer Changes:
None as this is new tooling

### Checklist

*If this is your first time contributing, please read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/main/CONTRIBUTING.md) document.*

All checklist items must be checked to enable merging this pull request:

*Contributor*
- [x] The in-code documentation has been updated with descriptive comments
- [x] The documentation has been assessed to determine if updates are necessary

*Integrator*
- [ ] FATES PASS/FAIL regression tests were run
- [ ] Evaluation of test results for answer changes was performed and results provided

### Documentation

- [Technical Note](https://github.com/NGEET/fates-docs) update: N/A
- [User's Guide](https://github.com/NGEET/fates-users-guide) update:  https://github.com/NGEET/fates-users-guide/issues/49

### Test Results:

*CTSM (or) E3SM (specify which) test hash-tag:*

*CTSM (or) E3SM (specify which) baseline hash-tag:*

*FATES baseline hash-tag:*

*Test Output:*

Issue #1080, opened 2023-09-05
Issues and workflow around initialization of FATES for global historical runs with land use change


ckoven:
Hi All,

I wanted to raise a few questions around initialization of land use in FATES for global historical runs with land-use change. As we get closer to being able to do this, some of the complexities of doing so are becoming more apparent to me. This is a long post, so no need for rapid response from anyone, but hopefully it can provide some basis for discussion going forward.

Traditionally in the big-leaf model, a global transient historical run has three parts: 

1. Spin-up under 1850 (really early 20th century) climate, 1850 CO2, and constant 1850 land-use with either accelerated decomposition or matrix spinup turned on to equilibrate vegetation and soils
2. Spin-up under 1850 (really early 20th century) climate, 1850 CO2, and constant 1850 land-use with normal soil dynamics turned on to re-equilibrate vegetation and soils to slightly different dynamics.  
3. Transient dynamics of land-use, climate, and/or CO2 from 1850 onwards

In FATES with land-use change, this is going to have to work differently. There are two reasons for this. The first is that from FATES's perspective there really isn't any such thing as "constant 1850 land-use". Since harvest of any primary land creates a one-way transfer of primary land to secondary land, there can't ever be any steady state as long as there is any harvest from primary land or other nonzero net transition rates.  The second reason is that the secondary land is intentionally meant to be out of equilibrium, and with an age structure that reflects its disturbance history. So if we want to start considering model results at, e.g., 1850, we need to initialize the model in such a way that the age structure is already present at that point. The GFDL group, which shares much of the underlying conceptual structure and assumptions of FATES, has been working on this for a long time and written extensively on the point, e.g.: https://doi.org/10.1175/2011EI401.1, which basically says that one needs to initialize land use in the model 100-150 years before the time point where one wants to start looking at the behavior.

So the question I want to pose is on how to set things up to do this for FATES in as simple and reproducible a way as possible. What I think a general workflow would look like might be one of the following general strategies:

**Strategy 1**

This would be analogous I think to the GFDL paper I linked to above:

1. Spin-up under 1850 (really early 20th century) climate, 1850 CO2, and no land-use (and, if nocomp configuration is active, a prescribed potential-vegetation distribution map) with either accelerated decomposition or matrix spinup turned on to equilibrate vegetation and soils.
[...]

PR #1116, opened 2023-11-10
V2 Land Use Change

### LINKED ISSUE #1084: Logging fractions to product pools parameters
Currently FATES uses a [single parameter ](https://github.com/NGEET/fates/blob/main/parameter_files/fates_params_default.cdl#L1668)to describe what happens to the wood product that is generated from logging, which is applied across al woody PFTs. Classic CLM and ELM both share a set of PFT-dependent parameters and logic.  @lawrencepj1 tells me that these numbers come from the "Grand Slam protocol", which appears to be this [Houghton et al. 1983 paper](https://esajournals.onlinelibrary.wiley.com/doi/10.2307/1942531)

The CLM & ELM parameters table by PFT, which roughly follow numbers from table 3 of that paper, are pasted below. The first three parameters (pconv, pprod10, and pprod100) are the wood fractions into instantaneous-loss, 10-year, and 100-year product pools resulting from land use change, and the last is the 10-year pool partitioning for forest harvest (so 1-pprodharv10 is the 100-year-pool harvest partitioning). In order to have the global runs with land-use comparable to the non-FATES ones, we should probably update the FATES parameters to at least match this level of complexity, and once there, perhaps there might be some effort to modernize further.  Som reorganization to the code logic (e.g., making the [`site_mass%wood_product` pools](https://github.com/NGEET/fates/blob/main/main/EDTypesMod.F90#L208) pft-indexed) would be needed to do this.

| pftname | pconv | pprod10 | pprod100 | pprodharv10 |
| --- | --- | --- | --- | --- |
| needleleaf_evergreen_temperate_tree   | 0.6 | 0.3 | 0.1 | 0.75 |
| needleleaf_evergreen_boreal_tree  | 0.6 | 0.3 | 0.1 | 0.75 |
| needleleaf_deciduous_boreal_tree  | 0.6 | 0.3 | 0.1 | 0.75 |
| broadleaf_evergreen_tropical_tree   | 0.6 | 0.4 | 0 | 1 |
| broadleaf_evergreen_temperate_tree  | 0.6 | 0.3 | 0.1 | 0.75 |
| broadleaf_deciduous_tropical_tree   | 0.6 | 0.4 | 0 | 1 |
| broadleaf_deciduous_temperate_tree  | 0.6 | 0.3 | 0.1 | 0.75 |
| broadleaf_deciduous_boreal_tree   | 0.6 | 0.3 | 0.1 | 0.75 |
| broadleaf_evergreen_shrub   | 0.8 | 0.2 | 0 | _ |
| broadleaf_deciduous_temperate_shrub   | 0.8 | 0.2 | 0 | _ |
| broadleaf_deciduous_boreal_shrub  | 0.8 | 0.2 | 0 | _ |
| c3_arctic_grass   | 1 | 0 | 0 | _ |
| c3_non-arctic_grass   | 1 | 0 | 0 | _ |
| c4_grass  | 1 | 0 | 0 | _ |

ckoven:
This PR represents the second set of changes needed for transient global land use change in FATES. 

### Description:
This is written on top of #1040, so the scope looks larger than the specific code unique to this PR.  The added functionality here is mainly based around running in nocomp+prescribed biogeography mode. So in this configuration, each patch has two distinct categorical variables: land-use class and nocomp PFT class, and the code allows the total patch area associated with both of these to be time-varying, so that land use governs land cover. The basic idea is that the code in #1040 reads in a prescribed timeseries of land-use states and transition rates from LUH2, which determines the rate of land-use-change disturbance on any given day for any given patch, and which happens alongside the existing harvest logic, which determines the transition rate of primary -> secondary lands. The new code in this PR builds on that to allow time-varying nocomp-PFT areas in response. 

The way this works is that the code reads in a land use : PFT mapping file, which determines, for every gridcell, what fraction of each land-use type is associated with each PFT. This mapping is (currently) time-independent.  The underlying dataset for this comes from @lawrencepj1's [files](https://gdex.ucar.edu/dataset/188b_oleson/file.html) that form the basis of the CLM and ELM BGC-configuration model's land use change data and the current FATES nocomp PFT data.  Using that data, this code reallocates nocomp PFT fractions during the disturbance routines for all newly-disturbed patches whose land-use identity changed due to either logging or land-use-change disturbance.  So, e.g., if primary lands are converted into rangelands, the PFT areal fractions of the newly-disturbed patches shifts in response. And the nocomp logic subsequently only allows the relevant PFT to recruit on the newly-changed patch.

In addition to the core functionality, a bunch of other things are also included to make the system work in a more general way. Model initialization had to be substantially changed to extend the nocomp logic to multiple land use types, and I pulled out bare ground fraction as its own variable instead of having an index of zero as in the prior logic. I added a switch for potential vegetation mode (`use_fates_potentialveg`) which allows for spinning up the model under potential vegetation (i.e. all primary lands). If a run with land use is initialized from the restart file of a run with potential vegetation, then there is logic to automatically apply all needed land use transition rates and logging from primary land to other lands on the first day of the simulation, so that the initial land use states align with the driving data (see #1080 for more discussion). The parameters governing the fates of dead material produced during land use change needed to be updated to allow for global simulations (#1084).

This code is tied to E3SM branch https://github.com/glemieux/E3SM/tree/fates-landuse-pft, will coordinate PRs as this moves through the queue. The tool to regrid and reformat the land use : PFT mapping data is in #1107.

The basic functionality appears to work, though there are a few things that still need to be sorted out. But I wanted to post it now that the core code is hopefully there, so that people can start looking at it.

This represents hopefully the last major piece of code development needed to start allowing CMIP-class experiments with FATES in a prescribed biogeography configuration.

fixes #1061.
fixes #1084.
further progress along the plan lain out in #450 .

### Collaborators:
Lots of discussion with lots of people, and written with support of @glemieux.

In principle this might not be answer changing in all other configurations, but I wouldn't be surprised if roundoff level differences were introduced.

### Checklist

*Contributor*
- [x] The documentation has been assessed to determine if updates are necessary.  Relevant changes to documentation are [here](https://fates-users-guide.readthedocs.io/projects/tech-doc/en/latest/fates_tech_note.html#land-use-change).

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


dlawrenncar:
Excellent to see this PR.  Amazing work (in principle, not that I would be
able to really assess the code) Charlie and Greg!

rosiealice:
Yeah, what @dlawrenncar said.  That is an awe-inducing list of commits...

parameter_files/fates_params_default.cdl
```diff
@@ -1597,6 +1624,10 @@ data:
 
  fates_frag_cwd_frac = 0.045, 0.075, 0.21, 0.67 ;
 
+ fates_landuse_crop_lu_pft_vector = -999, -999, -999, -999, 11 ;
+
+ fates_max_nocomp_pfts_by_landuse = 4, 4, 2, 2, 1 ;
```
ckoven, line 1629: noting that these numbers need to change so that they are each <= `fates_maxpatches_by_landuse`

main/FatesInterfaceMod.F90
```diff
@@ -611,13 +619,13 @@ subroutine allocate_bcout(bc_out, nlevsoil_in, nlevdecomp_in)
...
-      allocate(bc_out%ftdd_parb(maxpatch_total,hlm_numSWb))
-      allocate(bc_out%ftid_parb(maxpatch_total,hlm_numSWb))
-      allocate(bc_out%ftii_parb(maxpatch_total,hlm_numSWb))
+      allocate(bc_out%albd_parb(fates_maxPatchesPerSite,hlm_numSWb))
+      allocate(bc_out%albi_parb(fates_maxPatchesPerSite,hlm_numSWb))
+      allocate(bc_out%fabd_parb(fates_maxPatchesPerSite,hlm_numSWb))
+      allocate(bc_out%fabi_parb(fates_maxPatchesPerSite,hlm_numSWb))
+      allocate(bc_out%ftdd_parb(fates_maxPatchesPerSite,hlm_numSWb))
+      allocate(bc_out%ftid_parb(fates_maxPatchesPerSite,hlm_numSWb))
+      allocate(bc_out%ftii_parb(fates_maxPatchesPerSite,hlm_numSWb))
```
glemieux, line 628: What predicated this update?

glemieux (changes requested):
Review Part 1.  I still need to review EDInitMod, EDPhysiologyMod, and FatesLandUseChangeMod

glemieux (commented):
Landuse v2 review part 2

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -30,7 +31,8 @@ module FatesLandUseChangeMod
   public :: get_landuse_transition_rates
```
rgknox, line 31: We've generally had the convention of CamelCase() for procedures in FATES.

main/FatesInterfaceTypesMod.F90
```diff
@@ -124,6 +124,7 @@ module FatesInterfaceTypesMod
                                                          ! bc_in%hlm_harvest_rates and bc_in%hlm_harvest_catnames
 
    integer, public :: hlm_use_luh                   ! flag to signal whether or not to use luh2 drivers
+   integer, public :: hlm_use_potentialveg          ! flag to signal whether or not to use potential vegetation only
```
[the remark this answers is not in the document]
rgknox: I was able to learn through context that this implies a simulation with only primary lands, right?

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -232,62 +239,86 @@ subroutine LoggingMortality_frac( pft_i, dbh, canopy_layer, lmort_direct, &
...
-            ! HARVEST_SH3 = harvest from secondary non-forest (assume this is young for biomass)
-
-            ! Get the area-based harvest rates based on info passed to FATES from the boundary condition
-            call get_harvest_rate_area (patch_land_use_label, hlm_harvest_catnames, &
-                 hlm_harvest_rates, frac_site_primary, secondary_age, harvest_rate)
+      call get_luh_statedata(bc_in, state_vector)
+      site_secondaryland_first_exceeding_min =  (state_vector(secondaryland) .gt. currentSite%min_allowed_landuse_fraction) &
+           .and. (.not. currentSite%landuse_vector_gt_min(secondaryland))
+
+      if (.not. currentSite%transition_landuse_from_off_to_on) then
```
rgknox, line 256: Why is this comparison made? I don't think I understand this flag.

biogeochem/FatesLandUseChangeMod.F90
samsrabin: _[Pushes glasses up nose]_ Technically that's PascalCase!

biogeochem/EDLoggingMortalityMod.F90
ckoven: The `transition_landuse_from_off_to_on` is for handling the special case of the first timestep after leaving potential vegetation mode. In this case, all prior historical land-use, including harvest, needs to be applied on that first day. So logging rates on that day are what is required to deforest exactly the amount of primary lands that will give the amount of secondary lands dictated by the land use state vector for that year, rather than whatever the continuous logging rate for that year is supposed to be according to the transition matrix.  

Will note to add comments in code to that effect.

main/EDInitMod.F90
```diff
@@ -450,80 +463,150 @@ subroutine set_site_properties( nsites, sites,bc_in )
...
+                      if (.not. is_crop(i_landusetype)) then
+                         do hlm_pft = 1,size( EDPftvarcon_inst%hlm_pft_map,2)
+                            do fates_pft = 1,numpft ! loop round all fates pfts for all hlm pfts
+                               sites(s)%area_pft(fates_pft,i_landusetype) = sites(s)%area_pft(fates_pft,i_landusetype) + &
+                                    EDPftvarcon_inst%hlm_pft_map(fates_pft,hlm_pft) * bc_in(s)%pft_areafrac_lu(hlm_pft,i_landusetype)
+                            end do
+                         end do !hlm_pft
+                      else
+                         ! for crops, we need to use different logic because the bc_in(s)%pft_areafrac_lu() information only exists for natural PFTs
+                         sites(s)%area_pft(crop_lu_pft_vector(i_landusetype),i_landusetype) = 1._r8
```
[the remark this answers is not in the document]
ckoven: @samsrabin I guess it would be even more correct to make `crop_lu_pft_vector` an integer array, would that also be straightforward to do?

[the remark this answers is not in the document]
samsrabin: Oh yeah, definitely. I guess I was assuming a reason for it to be real, but seems not. I've pushed the change to the PR.

biogeochem/EDLoggingMortalityMod.F90
rgknox: thanks for the explanation

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -296,7 +327,7 @@ subroutine LoggingMortality_frac( pft_i, dbh, canopy_layer, lmort_direct, &
             if (cur_harvest_tag == 0) then
```
samsrabin, line 327: @glemieux In my CTSM testing (Derecho Intel) with `fates_harvest_mode` 1 ("fates logging via fates logging event codes") and 2 ("fates harvest driven by CLM landuse timeseries data"), I'm getting an error here that says `cur_harvest_tag` "is being used... without being defined." It is obviously _defined_, at line 240, so I think it means uninitialized.

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -335,21 +366,34 @@ subroutine LoggingMortality_frac( pft_i, dbh, canopy_layer, lmort_direct, &
...
+         else
+            lmort_direct     = 0.0_r8
+            lmort_collateral = 0.0_r8
+            lmort_infra      = 0.0_r8
+            if (canopy_layer .eq. 1) then
+               l_degrad         = harvest_rate
+            else
+               l_degrad         = 0.0_r8
+            endif
+         endif
```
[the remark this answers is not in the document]
samsrabin: I think the if block should actually be like this:
```fortran
if (prt_params%woody(pft_i) == itrue)then
   lmort_direct     = harvest_rate
else if (canopy_layer .eq. 1) then
   l_degrad         = harvest_rate
endif
```

main/FatesInterfaceTypesMod.F90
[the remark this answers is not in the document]
ckoven: yes, correct. adding comment to explain.

main/FatesInterfaceMod.F90
```diff
@@ -563,7 +564,13 @@ subroutine allocate_bcin(bc_in, nlevsoil_in, nlevdecomp_in, num_lu_harvest_cats,
          allocate(bc_in%hlm_harvest_catnames(0))
       end if
 
-      allocate(bc_in%pft_areafrac(natpft_lb:natpft_ub))
+      if ( hlm_use_fixed_biogeog .eq. itrue) then
+         if (hlm_use_luh .eq. itrue ) then
+            allocate(bc_in%pft_areafrac_lu(size( EDPftvarcon_inst%hlm_pft_map,2),num_luh2_states-n_crop_lu_types))
```
glemieux, line 569: This is allocating more indices than is necessary as it was using the number of states which is 12
```suggestion
            allocate(bc_in%pft_areafrac_lu(size( EDPftvarcon_inst%hlm_pft_map,2),n_landuse_cats-n_crop_lu_types))
```

biogeochem/EDLoggingMortalityMod.F90
glemieux: Just a quick update on this: I haven't tested this with https://github.com/ESCOMP/CTSM/pull/2507, but I was able to get a 1 year `fates_harvest_mode = 1` case to run without hitting this using the latest commits here and with https://github.com/E3SM-Project/E3SM/pull/6353.

samsrabin: I'm still seeing it, unfortunately. 

It looks like it's possible if
- `.not. currentSite%transition_landuse_from_off_to_on`
- and `site_secondaryland_first_exceeding_min` false
- and `logging_time` true
- and `prt_params%woody(pft_i)` true

and either

- `hlm_use_lu_harvest` false

or:

- `hlm_use_lu_harvest` true
- and not `hlm_harvest_units == hlm_harvest_area_fraction`
- and not `hlm_harvest_units == hlm_harvest_carbon`

Of the either/or, my setup has the former: `hlm_use_lu_harvest` (in CTSM, `do_harvest`) is false. So it seems like we need some sort of guardrail, depending on (respectively) whether that's a valid setup or not:
- In the namelist build check: Do not allow the HLM equivalent of `hlm_use_lu_harvest` to be false if... what? `fluh_timeseries` is pointing to a file?
- In `LoggingMortality_frac()`: Immediately return (or don't get called in the first place) if the HLM equivalent of `hlm_use_lu_harvest` is false?

ckoven: @sshu88 could you help with this debugging of the `cur_harvest_tag1` variable?

samsrabin: In CTSM, it looks like `hlm_harvest_units` is set by looking at the units field of the harvest rate variable on the `flanduse_timeseries` file and/or (?) the `fluh2_timeseries` file. I'm not sure if there's a way to check those units in the build namelist step. If not, we'd want to either (a) add a check as early as possible in the actual model code so that it can fail gracefully with a helpful error message, or (b) add handling to ignore harvest rates given on `flanduse_timeseries`/`fluh_timeseries` if `do_harvest` is off.

ckoven: Hi @samsrabin the prior behavior was to use the units field of the `flanduse_timeseries` file to set `flanduse_timeseries`. This should still be the case only when `fates_harvest_mode = 2`, which we've retained for backwards compatibility.  The new behavior is for `fates_harvest_mode = 3` and `fates_harvest_mode = 4`, which should be setting the `hlm_harvest_units` directly: if 3 then area-based, if 4 then mass-based.

samsrabin: Okay, yes, I do see `fates_harvest_mode` 3 and 4 setting the units directly now. That explains why those aren't failing for me with `do_harvest` false. The problem, I think, is how to handle this situation:
- `do_harvest` false 
- AND [`flanduse_timeseries` and/or (?) `fluh_timeseries`] specified with harvest variable included
- AND `fates_harvest_mode` 1 or 2.

Do we want that to be a valid combination?

samsrabin: Well, I guess the same question applies with `do_harvest` false and `fates_harvest_mode` 3 or 4. Any non-zero `fates_harvest_mode` feels contradictory with `do_harvest` false, so I feel like the namelist build should throw an error in such a situation.

ckoven: @samsrabin I guess that, on some level, the idea of transient land use doesn't really make sense without harvest allowed at least in principle, since only harvest can transition land from primary to secondary categories. So perhaps the simplest solution is to say that if `use_fates_luh == .true.`, then `fates_harvest_mode` must be greater than or equal to 2, and `do_harvest` must be true?

samsrabin: I think that mostly makes sense, @ckoven. But when is `fates_harvest_mode` allowed to be 1?

ckoven: @samsrabin `fates_harvest_mode == 1` is only really appropriate for site-scale simulations (or maybe very highly idealized large-scale ones). But not with any of the global historical or scenario-based land use workflows.

sshu88: > @sshu88 could you help with this debugging of the `cur_harvest_tag1` variable?

Hi @ckoven @samsrabin @glemieux . I tested the global scale run but did not test site-level logging simulation. I put initialization code (https://github.com/NGEET/fates/commit/09aed9fa84a1d3032583cdd6372b763b35ba5928) and the model passed test.

samsrabin: @ckoven So are these the settings we want to fail in the namelist build? (In FATES runs only, of course.)
1. `do_harvest` false AND `use_fates_luh` true — Can't have land use without harvest
2. `do_harvest` false AND `fates_harvest_mode` specified as something other than 0 — User has specified a harvest mode but not enabled harvest (could instead warn and set `do_harvest` true)
3. `do_harvest` true AND `use_fates_luh` false — Can't have harvest without land use
4. `do_harvest` true AND `fates_harvest_mode` specified as 0 — Can't have harvest without harvest
5. `do_harvest` true AND `fates_harvest_mode` specified but mismatched with units on `fluh_timeseries` — We don't want the user's intention to be ignored by the code parsing `fluh_timeseries`
6. `do_harvest` true AND `fates_harvest_mode` 1 AND `fluh_timeseries` specified

Sorry to be such a pill about this, but I know deep in my bones I will get myself into trouble without these kinds of guardrails!

samsrabin: Also:
7. `do_harvest` true in non-transient run

Anything else special for non-transient runs?

glemieux: No worries, there are good questions! 

I think these are correct, although I have followup on the following:
2. Do we have users who just want to use the logging codes and not landuse harvest?
1/3. Are there scientifically relevant simulations to do with landuse but no harvesting?

Also, I think we should move away from using `do_harvest` at all for fates.  I.E. I think `do_harvest` should only be for HLM-specific harvest calls like `CNHarvest`.   I'm working on an update that disentangles this on the HLM side.  I've also started a PR to draft a page in the user's guide explaining the various `fates_harvest_modes` in more detail to help drive this distinction: https://github.com/NGEET/fates-users-guide/pull/63

samsrabin: Are you suggesting we fully deprecate `do_harvest` as a user-facing setting? If so, I'm totally on board. It seems redundant with `fates_harvest_mode`.

I'll let Charlie comment on your follow-up questions.

glemieux: > Are you suggesting we fully deprecate `do_harvest` as a user-facing setting? If so, I'm totally on board. It seems redundant with `fates_harvest_mode`.

Yep, exactly.  Now that fates harvesting options are moving beyond the HLM harvest modes I think it makes it easier to manage.  Started up a PR to your ctsm PR here: https://github.com/samsrabin/CTSM/pull/2

ckoven: @samsrabin I think we should allow harvest without the full land use change drivers, I could imagine someone wanting to do that experiment. So I would allow your number 3 (or the variant of it in which we remove `do_harvest` entirely: to allow harvest mode to be any value of 1-4 when `use_fates_luh == false`. Otherwise I agree with al the above.

ckoven:
@sshu88 I added a fix for the error that you identified recently in the harvest rates, in commit [6080ab6](https://github.com/NGEET/fates/pull/1116/commits/6080ab6032814c5e70547bcf82cd39a028e5253f). Could you take a look and confirm that it is doing the right thing now?

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -449,15 +453,21 @@ subroutine get_harvest_rate_area (patch_land_use_label, hlm_harvest_catnames, hl
...
-           harvest_rate = min((harvest_rate / frac_site_primary),frac_site_primary)
+           harvest_rate = min((harvest_rate / frac_site_primary),1._r8)
         else
            harvest_rate = 0._r8
         endif
      else if (patch_land_use_label .eq. secondaryland) then
-        if (frac_site_secondary .gt. fates_tiny) then
-           harvest_rate = min((harvest_rate / frac_site_secondary), frac_site_secondary)
+        ! the .gt. -0.5 in the next line is because frac_site_secondary returns -1 if no secondary area.
+        if (frac_site_secondary .gt. fates_tiny .and. frac_site_secondary .gt. -0.5_r8) then
```
sshu88, line 465: I'm not pretty sure what does -0.5 mean here?

biogeochem/EDLoggingMortalityMod.F90
```diff
@@ -449,15 +453,21 @@ subroutine get_harvest_rate_area (patch_land_use_label, hlm_harvest_catnames, hl
      !  Normalize by site-level primary or secondary forest fraction
      !  since harvest_rate is specified as a fraction of the gridcell
      !  also need to put a cap so as not to harvest more primary or secondary area than there is in a gridcell
+     !  For secondary, also need to normalize by the young/old fraction.
      if (patch_land_use_label .eq. primaryland) then
         if (frac_site_primary .gt. fates_tiny) then
-           harvest_rate = min((harvest_rate / frac_site_primary),frac_site_primary)
+           harvest_rate = min((harvest_rate / frac_site_primary),1._r8)
```
sshu88, line 459: 1._r8 may remove the whole patch. If you have certain protection for no_comp mode in PatchDyanmicMod.F90 to prevent removal of all primary forest, this makes sense. Otherwise I would suggest to use a less than 1 number.

biogeochem/EDLoggingMortalityMod.F90
ckoven: really just checking that it is greater than -1, since that is the value returned if there is secondary area is less than fates_tiny.

main/EDTypesMod.F90
```diff
@@ -547,5 +549,48 @@ function get_current_landuse_statevector(this) result(current_state_vector)
...
+     class(ed_site_type) :: this
+     real(r8)            :: secondary_young_fraction
+     real(r8)            :: secondary_young_area
+     real(r8)            :: secondary_old_area
+
+     ! !LOCAL VARIABLES:
+     type(fates_patch_type), pointer :: currentPatch
+
+     secondary_young_area = 0._r8
+     secondary_old_area = 0._r8
```
sshu88, line 574: Is "secondary mature" better than "secondary old"? Just want to keep some naming convention to be consistent.

biogeochem/EDLoggingMortalityMod.F90
sshu88: I see. Makes sense.

biogeochem/EDLoggingMortalityMod.F90
ckoven: yes, we have a cap on disturbance rates [here](https://github.com/NGEET/fates/pull/1116/files#diff-28c679efbe804905b8b4fac3ccdbb04f6bb3bcd2794fe1a4556d5541e008a750R217)

sshu88:
> @sshu88 I added a fix for the error that you identified recently in the harvest rates, in commit [6080ab6](https://github.com/NGEET/fates/pull/1116/commits/6080ab6032814c5e70547bcf82cd39a028e5253f). Could you take a look and confirm that it is doing the right thing now?

Codes look good to me and I only made some suggestions. Thank you for the work!

main/EDInitMod.F90
```diff
@@ -450,80 +463,149 @@ subroutine set_site_properties( nsites, sites,bc_in )
...
-                   sites(s)%area_pft(ft)=0.0_r8
-                   ! remove tiny patches to prevent numerical errors in terminate patches
+             use_fates_luh_if: if (hlm_use_luh .eq. itrue) then
+                ! MAPPING OF FATES PFTs on to HLM_PFTs with land use
+                ! add up the area associated with each FATES PFT
+                ! where pft_areafrac_lu is the area of land in each HLM PFT and land use type (from surface dataset)
+                ! hlm_pft_map is the area of that land in each FATES PFT (from param file)
+
+                ! First check for NaNs in bc_in(s)%pft_areafrac_lu. If so, make everything bare ground.
+                if ( .not. (any( isnan( bc_in(s)%pft_areafrac_lu (:,:) )) .or. isnan( bc_in(s)%baregroundfrac))) then
```
rgknox, line 474: I wonder if this isnan will give us trouble.  I wonder if it would be better to initialize at a known unknown, like the  [hlm_hio_ignore_val](https://github.com/NGEET/fates/blob/sci.1.76.3_api.35.1.0/main/FatesInterfaceTypesMod.F90#L77-L80) or something else?

ckoven: Just to update that I looked at a run that @glemieux did with the most recent code, and it doesn't show the same artifact as in the figure above, so whatever I did to cause that doesn't seem to be persistent.

biogeochem/EDLoggingMortalityMod.F90
rgknox: @ckoven , I didn't see fates_tiny defined.  What is the its intent here?  Is that a generic very small number used to prevent div0s? Or is it the same as fortran tiny, ie the smallest possible number for this floating point type?  I very often use nearzero to prevent div0s

ckoven: yes, sorry, `nearzero`

glemieux:
Regression testing is showing mostly expected results, but I'm seeing an exact restart issues with two satellite phenology tests:
```
    FAIL ERP_P128x2_Ld30.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdSatPhen COMPARE_base_rest
    FAIL ERS_Ld30.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdSatPhen COMPARE_base_rest
```

glemieux:
It looks like the seed values for the base are either zero or uninitialized and the restart is uninitialized everywhere:
```
37526  FATES_SEEDLING_POOL   (lon,lat,time)  t_index =     18    18
37527        1077     3312  (    33,     2,     1) (    34,     6,     1) (    33,     2,     1) (    33,     2,     1)
37528                 1083   0.000000000000000E+00  -1.199999999999999E+37 1.2E+37  0.000000000000000E+00 2.5E-01  0.000000000000000E+00
37529                 1083  -1.199999999999998E+37  -1.199999999999999E+37         -1.199999999999999E+37         -1.199999999999999E+37
37530                 3312  (    60,    11,     1) (    33,     2,     1)
37531           avg abs field values:    9.051314643096083E+36    rms diff: 5.2E+36   avg rel diff(npos):  2.5E-01
37532                                    1.199999999999980E+37                        avg decimal digits(ndif):  1.6 worst:  0.0
37533  RMS FATES_SEEDLING_POOL              5.1713E+36            NORMALIZED  4.9131E-01
```
`FATES_SEEDS_IN_LOCAL` are always zero so these are B4B.

Looking at the passing baseline I generated recently the ERS differences are always showing unititialized everywhere for base and rest:
```
65153  FATES_SEEDLING_POOL   (lon,lat,time)  t_index =     31    31
65154                 3312  (    59,    10,     1) (    33,     2,     1)
65155                 1083  -0.120000000000000E+38  -0.120000000000000E+38
65156                 1083  -0.120000000000000E+38  -0.120000000000000E+38
65157                 3312  (    59,    10,     1) (    33,     2,     1)
65158           avg abs field values:    1.199999999999980E+37
65159                                    1.199999999999980E+37
```

glemieux:
The exact restart satellite phenology issue noted above has been corrected per [c9e7271](https://github.com/NGEET/fates/pull/1116/commits/c9e72712f2822665083d3d6f9a0c39e178f2de52).  

The majority of fates testmods with differences are those that engage fixed biogeography, which includes all satellite phenology test and the pre-existing `FatesColdLUH2` test.

```
FAIL ERS_D_Ld15.f45_f45_mg37.I2000Clm50FatesRs.derecho_gnu.clm-FatesColdTwoStreamNoCompFixedBioGeo BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
FAIL ERP_P128x2_Ld30.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdSatPhen BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
FAIL ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLandUse BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
FAIL ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLUH2 BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
FAIL ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdFixedBiogeo BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
FAIL ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdNoCompFixedBioGeo BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
FAIL ERS_Ld30.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdSatPhen BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
FAIL SMS_D.1x1_brazil.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdSatPhen BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
FAIL SMS_Lm1.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdBasic BASELINE fates-sci.1.76.4_api.35.1.0-ctsm5.2.008: DIFF
```

There is also a difference in the `FatesColdLandUse` test which is due to the fix introduced in [bb3d58e](https://github.com/NGEET/fates/pull/1116/commits/bb3d58e0878964490ddc998bfd5593c53dab8b97).  This test exercises the fates logging only mode.  I discovered during review of previous tests that this mode was not actually engaging the logging for the final day of this test.  As such the `FATES_DISTURBANCE_RATE_LOGGING` (and all downstream outputs) are now appropriately resulting in a non-zero logging disturbance on the last timestep of the output.  All other previous timesteps are b4b.

glemieux:
The DIFF associated with the `FatesColdLandUse` testmod has been fixed per https://github.com/ESCOMP/CTSM/pull/2507/commits/a486e86065223a882790a6078647a82c592ffdaf.  The option name change for this testmod was getting truncated and as such, not populating the harvest rates correctly.

glemieux:
I looked into this some more this morning to get a better sense of what the initialization in fbg modes look like (without landuse).  The new way in which we initialize the bareground with this update creates a difference in the initial patch areas.  With this change we directly reference the incoming HLM `wt_nat_patch` for the bareground (via `bc_in(s)%pft_areafrac(0)`) which is correctly being passed as zero bare ground fraction.  That said, since there can be other hlm pfts with non-zero fraction beyond the 14 pfts not being mapped to fates pfts, we can end up with a total fates pft fraction less than one.  This account for this, we normalize against the sum of all the fates pft areas (including bareground).

On `main` currently we generate the bareground fraction (in fbg+nocomp modes) by computing the difference between the sum of the fates pft fractions (not including the bareground) and use that as the bareground fraction.  As a result we will generate some amount of bareground when we have the case above, even if the actual `wt_nat_patch` is zero.

@ckoven should we retain the old method of the bareground generation for nocomp+fbg (without landuse) given the above?

ckoven:
Thanks @glemieux I was also just looking into this and not understanding why there would be a difference. Your answer makes sense to me. To me, the new logic makes more sense than the old logic: we shouldn't be making bare ground area if the datasets say not to, it makes more sense to me to rescale the existing PFT areas proportionally.

But a larger problem remains if there are PFTs beyond the 14 that we have been using, that means that they are crop PFTs and we should be mapping those onto our crop-analog PFT(s). But that also seems to me like a subsequent development for after when this is merged.

glemieux:
@ckoven I think the only thing that might be of concern here, or at least should be noted for posterity, is that this will result in slightly lower `sp_tlai` values since the patch areas that we use to scale the `sp_tlai` will be slightly larger due to the patch area normalization:
https://github.com/NGEET/fates/blob/b8e4eee5ed46daf5c9e710e9ebbe6d20464adbc8/biogeochem/EDPhysiologyMod.F90#L1838-L1845

glemieux (approved):

Issue #899, opened 2022-09-13
fleaf is sensitive to compiler math imprecisionss


rgknox:
When determining the leaf and stem area profiles, we use an intermediate variable called "fleaf" to parse out total area to the leaves and stems in each layer. See its usage here: https://github.com/NGEET/fates/blob/sci.1.59.5_api.24.1.0/biogeochem/EDCanopyStructureMod.F90#L1710-L1729

Essentially, its the fraction of leaf to total stem and leaf in a cohort.   

However in PR #887 @glemieux and I noticed that the calculation of fleaf was strangely sensitive, in that we get different results if we use lai/(lai+sai)  versus tree_lai / (tree_lai + tree_sai).  Note that the difference between lai and tree_lai is that the former the leaf area per canopy area (of the patch), and the latter is the the leaf area per crown area (of the cohort).  

IE   lai = tree_lai *crown_area/canopy_area

However, what makes this really strange, is that the term crown_area/canopy_area term is in both the numerator and denominator of fleaf (when based on lai), and should cancel out.  Thus:  

tree_lai / (tree_lai + tree_sai) == lai / (lai + sai)

However, in the testing of #887, we found that this is not so. In this commit you can see we were testing the two methods. And eventually decided to stay with using the lai formulation, to maintain b4b results with the base.

https://github.com/NGEET/fates/pull/887/commits/bf99a0b6ba0901411875ce566c8a69d2b507c180

This issue is to serve a reminder that there is a difference, which is probably related to numerical imprecision.  My personal take is that we should use tree_lai to form the ratio.  The differences could be due to small crown_area values, perhaps as a result of very small number densities, and since tree_lai and tree_sai is independant of number density, there is less chance for numerical weirdness when calculating a ratio.  Plus, we test if we should do the math on tree_lai and tree_sai, not lai and sai.

Issue #674, opened 2020-07-20
Issue of crown depth


JunyanDing:
In fire module the parameter fates_fire_crown_depth_frac is used to calculate the crown depth
 https://github.com/NGEET/fates/blob/master/fire/SFMainMod.F90#L885-L898

But in allometry module the crown depth is calculated as the smaller value of tree height and 0.1m:
https://github.com/NGEET/fates/blob/master/biogeochem/FatesAllometryMod.F90#L1910 

Should we calculate the crown depth the same way as in fire module? Also, in fire module this parameter is named as EDPftvarcon_inst%crown, which seems informative.

ckoven:
Looking at the code, the subroutine `CrownDepth` is only used by the hydraulics code, does it have the same physical/ecological meaning as in the fire code?  If so, I agree that we ought to unify those calculations.

JunyanDing:
It affects the length of the stem, hence the total hydraulic conductance of the tree, thus the leaf water potential.

rosiealice:
That seems to me like it's -not- the same definition as the fire module,which is trying to calculate the vertical height over which there are leaves (as opposed to stem). This might be an issue of definitions and the hydraulics one sounds more like just 'tree height' to me. Although, admittedly, most water doesn't have to actually reaxh the very top of the tree so you could imagine unifying the concepts a little more....

ckoven:
ok, @rosiealice, I see your point. But I still think there may be an error on https://github.com/NGEET/fates/blob/master/biogeochem/FatesAllometryMod.F90#L1910:
shouldn't it be `crown_depth = max(height,0.1_r8)`?

rosiealice:
Ah yes, indeed...

rosiealice:
That might cause quite a bit of answer-changing...

ckoven:
yeah.  @xuchongang, @bchristo, @rgknox: thoughts on this?

JunyanDing:
@ckoven and @rosiealice , I disagree with both of you 😉 
That line of code is on allometry module, which means FATES treats crown depth conceptually differently for fire and the rest of other processes though ut is only used in hydro currently. 
Stem height is calculated as tree height - crown depth. If set to  max(height, 0.1), the height of stem will either be zero if tree is shorter than 0.1m or negative if the tree is taller than 0.1m

JunyanDing:
For hydro , I will set the height of stem to be the middle of the crown. I think this is the appropriate height(or the length of flow path) for all the leaves in average

[...]

PR #1128, opened 2023-12-08
New allometric modes

mpaiao:
### Description:

1. New allometric function for above-ground biomass (`allom_amode = 4`), which is similar to [Chave et al. (2014)](https://dx.doi.org/10.1111/gcb.12629), except that the exponent for wood density is independent of the one for "tree size" (i.e., DBH*DBH*Height). Although this function could be simplified externally (by incorporating wood density into the scaler), this makes the dependency on wood density more transparent and may make parameter perturbation experiments a bit easier. 
2. New allometric function for leaf biomass and crown area (`allom_lmode = 4`), which uses tree size as the predictor. In the case of leaf biomass, the function also is a function of top-canopy SLA, using a similar formulation as the one for above-ground biomass.
3. Crown depth has now allometric modes too (variable allom_dmode). If `allom_dmode=1`, crown depth is a constant fraction of height (the former default); when `allom_dmode=2`, crown depth is defined as p1 * Height^p2, following [Poorter et al. (2006)](https://dx.doi.org/10.1890/0012-9658(2006)87{$[$}1289:AOMTST{$]$}2.0.CO;2). Former parameter `crown_depth_frac` was replaced with parameter `allom_h2cd1`, and a new parameter `allom_h2cd2` was added for the exponent.
4. All allometric mode variables were converted to integers when they are loaded (parteh/PRTParamsFATESMod.F90). This avoids converting them to integers every time the allometric function is called.
5. Small edits to CheckParams (parteh/PRTParamsFATESMod.F90). 
 - The current implementation of allom_hmode=2 is a bit unusual because the minus sign of the Weibull function is incorporated in parameter p2 (meaning that p2 must be negative). Added a check to make sure the parameter is negative.
 - Instead of aborting the run as soon as an inconsistency in parameters is spotted, the subroutine now continues to check other parameters, and report all the errors before stopping the run. This avoids users recompiling/submitting FATES multiple times to fix one inconsistency at a time.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


rgknox:
awesome set of updates @mpaiao looking forward to getting this integrated

rgknox:
I was looking through an old issue related to fates Hydro and how it perceives and uses crown depth, see here: https://github.com/NGEET/fates/issues/674

I think the things we discuss in that thread are out of scope for this PR, but would be good to re-evaluate and move forward on them though.

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1616,12 +1625,10 @@ subroutine leaf_area_profile( currentSite )
                 ! is obscured by snow.
 
                 layer_top_height = currentCohort%height - &
-                     ( real(iv-1,r8)/currentCohort%NV * currentCohort%height *  &
-                     prt_params%crown_depth_frac(currentCohort%pft) )
+                     ( real(iv-1,r8)/currentCohort%NV * crown_depth )
```
rgknox, line 1628: nice update here @mpaiao 

This will  have a conflict with the new two-stream PR.  I created a new subroutine that calculates the leaf area of crown layers, and it will need to leverage your fixes, see: https://github.com/NGEET/fates/pull/1141/files#diff-4fd8508e4a25680b45ecbfec6b6d80980a5b1e358af731203ee9d29266be46dbR2655

biogeochem/FatesAllometryMod.F90
```diff
@@ -305,7 +306,7 @@ subroutine h2d_allom(h,ipft,d,dddh)
                 p3          => prt_params%allom_d2h3(ipft), &
                 allom_hmode => prt_params%allom_hmode(ipft))
 
-      select case(int(allom_hmode))
```
rgknox, line 308: agreed, smarter to convert the switch on the read in, no need to convert over and over again in the dynamics loop

biogeochem/FatesAllometryMod.F90
```diff
@@ -522,6 +534,12 @@ subroutine carea_allom(dbh,nplant,site_spread,ipft,crowndamage,c_area,inverse)
           call carea_2pwr(dbh_eff,site_spread,d2bl_p2,d2bl_ediff,d2ca_min,d2ca_max, &
                crowndamage, c_area, do_inverse)
           capped_allom = .true.
+       case (4)
+          dbh_eff = min(dbh,dbh_maxh)
```
rgknox, line 538: Noting that height capping is compatible with this scheme and implementation makes sense.

biogeochem/FatesAllometryMod.F90
```diff
@@ -1489,7 +1507,111 @@ subroutine dh2blmax_2pwr(d,p1,p2,dbh_maxh,c2b,blmax,dblmaxdd)
...
+     real(r8), intent(in)            :: p3       ! Log-slope parameter for SLA    [     -]
+     real(r8), intent(in)            :: slatop   ! Top canopy specific leaf area  [ m2/gC]
+     real(r8), intent(in)            :: c2b      ! Carbon to biomass multiplier   [kg/kgC]
+     real(r8), intent(in)            :: dbh_maxh ! dbh at maximum height          [    cm]
+     real(r8), intent(out)           :: blmax    ! Leaf biomass                   [   kgC]
+     real(r8), intent(out), optional :: dblmaxdd ! Leaf biomass derivative        [kgC/cm]
+     !--- Local variables
+     real(r8) :: duse
+     !---~---
+
```
rgknox, line 1576: header and descriptions are nice and verbose, units, refs, great

biogeochem/FatesAllometryMod.F90
```diff
@@ -1489,7 +1507,111 @@ subroutine dh2blmax_2pwr(d,p1,p2,dbh_maxh,c2b,blmax,dblmaxdd)
...
+           ! future.
+           !---~---
+           dblmaxdd = 0._r8
+           !---~---
+        else
+           !---~---
+           !   Find the leaf biomass derivative, noting that height is actually
+           ! a function of DBH.
+           !---~---
+           dblmaxdd = p2 * blmax * ( 2._r8 / duse + dhdd / h )
```
rgknox, line 1605: @mpaiao does this need to swap the 2 for c2b?

biogeochem/FatesAllometryMod.F90
```diff
@@ -1791,6 +1913,90 @@ subroutine dh2bagw_chave2014(d,h,dhdd,p1,p2,wood_density,c2b,bagw,dbagwdd)
...
+     bagw = p1 * (d*d*h)**p2 * wood_density**p3 / c2b
+     !---~---
+
+
+     !---~---
+     !   Compute the above-ground biomass derivative with DBH if needed, noting that
+     ! height is actually a function of DBH.
+     !---~---
+     if (present(dbagwdd)) then
+        dbagwdd = p2 * bagw * ( 2._r8 / d + dhdd / h )
```
rgknox, line 1989: same as above, need to switch 2 with c2b?

biogeochem/EDCanopyStructureMod.F90
mpaiao: Thanks for spotting this, @rgknox. Once the two-stream PR is integrated, I can update this part. I will leave this conversation open so we don't forget.

biogeochem/FatesAllometryMod.F90
mpaiao: No, this 2 comes from the derivative of blmax, because of the DBH<sup>2</sup> term in the main allometric equation. I can add a note to indicate this.

biogeochem/FatesAllometryMod.F90
mpaiao: No, this is also due to the derivative.

biogeochem/FatesAllometryMod.F90
```diff
@@ -2625,7 +2949,265 @@ subroutine cspline(x1,x2,y1,y2,dydx1,dydx2,x,y,dydx)
...
+         return
+      end if
+      !---~---
+
+
+      !--- First guess: use current DBH.
+      adbh  = dbh
+      call h_allom(adbh,ipft,hgt,dhgtddbh)
+      afun  = adbh * adbh * hgt - size
+      deriv = 2.0_r8 * adbh * hgt + adbh * adbh * dhgtddbh
```
rgknox, line 3010: what is 2 here?

rgknox: is it the derivative of x^2?

biogeochem/FatesAllometryMod.F90
```diff
@@ -2625,7 +2949,265 @@ subroutine cspline(x1,x2,y1,y2,dydx1,dydx2,x,y,dydx)
...
+         write(fates_log(),*) 'Size to DBH routine failed to converge!'
+         call endrun(msg=errMsg(sourcefile, __LINE__))
+         !---~---
+      end if
+      !---~---
+
+      return
+   end subroutine size2dbh
+   ! ============================================================================
+
```
rgknox, line 3210: You were very determined to find the inverse solution for dbh and h @mpaiao , not easy, kudos

biogeochem/FatesAllometryMod.F90
mpaiao: Yes, derivative of DBH<sup>2</sup>.

parteh/PRTParamsFATESMod.F90
```diff
@@ -1087,22 +1138,28 @@ subroutine PRTCheckParams(is_master)
...
               write(fates_log(),*) 'reproductive tissues are a special case'
               write(fates_log(),*) 'and therefore should not be included in'
               write(fates_log(),*) 'the parameter file organ list'
               write(fates_log(),*) 'fates_prt_organ_id: ',prt_params%organ_id(:)
-              write(fates_log(),*) 'Aborting'
-              call endrun(msg=errMsg(sourcefile, __LINE__))
+              write(fates_log(),*) '---~---'
+              write(fates_log(),*) ''
+              write(fates_log(),*) ''
+              nerror = nerror + 1
```
rgknox, line 1150: Like this

parteh/PRTParamsFATESMod.F90
```diff
@@ -1526,18 +1698,29 @@ subroutine PRTCheckParams(is_master)
...
      end do pftloop
 
 
+     ! If any error was found, abort. We add a single point to abort the run after all
+     ! checks so users can get all the errors and address them in one go (as opposed to
+     ! multiple submissions).
+     if (nerror > 0) then
+        write(fates_log(),*) 'One or more parameter errors found. Aborting.'
+        call endrun(msg=errMsg(sourcefile, __LINE__))
+     end if
```
rgknox, line 1722: double checked to make sure that all checks removed endrun and added nerror, looks good

biogeochem/FatesAllometryMod.F90
mpaiao: This one is totally borrowed from ED2 solvers, I just adapted it here. This allometric function is well behaved and should quickly converge with the Newton method, but I thought I should keep the full root finding approach in case we want to adapt for less well-behaved functions anywhere in FATES in the future.

rgknox:
@mpaiao these changes look pretty clean to me. The only issue I saw was the use of 2 instead of c2b in a couple of cases. I saw the use of 2 in a derivative also, could you just note in the text that it was the derivative of x^2 (if I got that right).  The root finding method is complicated, and I did not go through it in detail. But I did see that it was borrowed from ed2? If so, that is great, since its already been vetted.

rgknox (approved):

biogeochem/FatesAllometryMod.F90
mpaiao: I opted for writing the derivative this way instead of the more explicit (like in `dh2bagw_chave2014`) because it reduces the number of operations (especially power operations). I could update the one in `dh2bagw_chave2014` too if it interests.

rgknox: i think its fine

biogeochem/FatesAllometryMod.F90
```diff
@@ -1489,7 +1507,111 @@ subroutine dh2blmax_2pwr(d,p1,p2,dbh_maxh,c2b,blmax,dblmaxdd)
...
+     !    La phytomasse epigee d'une foret dense en Guiane Francaise
+     !    Acta Oecol.-Oec. Gen. 4: 237-251.
+     !    URL http://www.documentation.ird.fr/hor/fdi:010005089
+     !
+     ! Longo M, Saatchi SS, Keller M, Bowman KW, Ferraz A, Moorcroft PR,
+     !    Morton D, Bonal D, Brando P, Burban B et al. 2020. Impacts of
+     !    degradation on water, energy, and carbon cycling of the Amazon 
+     !    tropical forests. J. Geophys. Res.-Biogeosci. 125:
+     !    e2020JG005677. doi:10.1029/2020JG005677.
+     !
```
JessicaNeedham, line 1538: The new leaf biomass allometry looks great to me. One minor suggestion is to put suggested parameter values. It looks like the default d2bl1, d2bl2 and d2bl3 will give very high biomass with this equation so some suggested values for users wishing to use this option could be helpful.

biogeochem/FatesAllometryMod.F90
```diff
@@ -2173,7 +2405,99 @@ subroutine carea_2pwr(dbh,spread,d2bl_p2,d2bl_ediff,d2ca_min, &
...
+     !    Calculate area of ground covered by entire cohort. (m2)
+     ! Function of DBH (cm), height (m), canopy spread (m/cm) and number of 
+     ! individuals. 
+     !---~---
+
+     !--- List of arguments
+     real(r8)   , intent(inout) :: dbh         ! Diameter at breast/ref/ height     [   cm]
+     real(r8)   , intent(inout) :: height      ! Height                             [    m]
+     integer(i4), intent(in)    :: ipft        ! PFT index
+     real(r8)   , intent(in)    :: dbh_maxh    ! Minimum DBH at maximum height      [    m]
```
JessicaNeedham, line 2425: should units here be cm?

parteh/PRTParamsFATESMod.F90
JessicaNeedham: Agree.

JessicaNeedham (approved):
Thanks @mpaiao - these are great changes/additions. I was wondering whether we want to put suggested parameter values for various allometry options, since default parameters will be wacky with some of the options. We already do this for the Chave d2h and dh2bagw but we could add it in for some others as well. Maybe out of scope for this PR, but might be helpful in future.

biogeochem/EDCanopyStructureMod.F90
mpaiao: @rgknox Should I go ahead and edit subroutine `VegAreaLayer` now that two-stream has been merged?

biogeochem/FatesAllometryMod.F90
```diff
@@ -1791,6 +1913,90 @@ subroutine dh2bagw_chave2014(d,h,dhdd,p1,p2,wood_density,c2b,bagw,dbagwdd)
...
+     ! -----------------
+     ! d            -- Diameter at breast height           [     cm]
+     ! h            -- Total tree height                   [      m]
+     ! dhdd         -- Height derivative with dbh          [   m/cm]
+     ! p1           -- Parameter 1 (log-intercept)         [     --]
+     ! p2           -- Parameter 2 (power, or log-slope)   [     --]
+     ! p3           -- Parameter 3 (power, or log-slope)   [     --]
+     ! wood_density -- Wood density                        [  g/cm3]
+     ! c2b          -- Carbon to biomass multiplier ~ 2    [ kg/kgC]
+     ! dbh_maxh     -- DBH at maximum height               [     cm]
```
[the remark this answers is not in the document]
mpaiao: Indeed, thanks for spotting this!

biogeochem/FatesAllometryMod.F90
mpaiao: Yes, good catch, thanks!

biogeochem/FatesAllometryMod.F90
```diff
@@ -2634,6 +2966,10 @@ subroutine VegAreaLayer(tree_lai,tree_sai,tree_height,iv,nv,pft,snow_depth, &
     
     tree_vai = tree_lai + tree_sai
 
+    ! Ratio between crown depth and plant height
+    call CrownDepth(tree_height,pft,crown_depth)
+    frac_crown_depth = crown_depth / tree_height
```
mpaiao, line 2971: @rgknox Could this be causing the bit-for-bit mismatch? In the past we would apply the crown depth directly, whereas now we calculate the crown depth and divide by the tree height to get the fraction.

rgknox: Its the same math, so I would only expect round-off level differences here

glemieux: I've tested this out and I can confirm that this isn't causing diffs.

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1625,12 +1633,10 @@ subroutine leaf_area_profile( currentSite )
...
                    layer_top_height = currentCohort%height - &
-                        ( real(iv-1,r8)/currentCohort%NV * currentCohort%height *  &
-                        prt_params%crown_depth_frac(currentCohort%pft) )
-                   
+                        ( real(iv-1,r8)/currentCohort%NV * crown_depth )
+                                           
                    layer_bottom_height = currentCohort%height - &
-                        ( real(iv,r8)/currentCohort%NV * currentCohort%height * &
-                        prt_params%crown_depth_frac(currentCohort%pft) )
+                        ( real(iv,r8)/currentCohort%NV * crown_depth )
```
glemieux, line 1639: After discussing this with @rgknox, I ran a test manually reverting this section of code to a form that is more consistent with the old method like so:
```
 layer_top_height = currentCohort%height - &
                        ( real(iv-1,r8)/currentCohort%NV * currentCohort%height *  &
                          prt_params%allom_h2cd1(currentCohort%pft) )
```
Doing so resulted in b4b comparisons against a baseline built with Jessie's default allometry updates from #1093.

@rgknox insight here was recalling that we have had similar issues with attempting to refactor this section of code in the past.  As such, we feel that the code as written is good to go and the roundoff diffs are acceptable.

glemieux: For posterity I should note that https://github.com/NGEET/fates/issues/899 as a related connection here.

glemieux:
I re-ran the fates regression tests on `derecho` since it had been a while since the last tests Ryan ran.  As expected, all expected tests passed b4b.

Location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1128-fates-2`

PR #1115, opened 2023-11-04
Turn off hydraulic failure mortality in frozen soils

JessicaNeedham:
This PR address issue #1105 and turns off hydraulic failure mortality when soils are freezing, specifically, when any soil layer is below -2 degrees C. This should prevent high mortality of vegetation at high latitudes.  

### Collaborators:
@rosiealice @ckoven 

### Expectation of Answer Changes:
This is an answer changing fix - hydraulic failure will be reduced at high latitudes. 

### Checklist

*If this is your first time contributing, please read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/main/CONTRIBUTING.md) document.*

All checklist items must be checked to enable merging this pull request:

*Contributor*
- [x] The in-code documentation has been updated with descriptive comments
- [x] The documentation has been assessed to determine if updates are necessary

*Integrator*
- [x] FATES PASS/FAIL regression tests were run
- [x] Evaluation of test results for answer changes was performed and results provided

### Documentation

- [Technical Note](https://github.com/NGEET/fates-docs) update: https://github.com/NGEET/fates-docs/pull/61
- [User's Guide](https://github.com/NGEET/fates-users-guide) update: 

### Test Results:

I haven’t tested with the latest commits from main. I branched from FATES a3048a646 (E3SM b8be65d). The figure  below is compared with FATES  a3048a646. It's a no comp run - mean over year 10. 
Screenshot 2023-11-04 at 16 51 41

PR #1130, opened 2023-12-08
Disable hydraulic failure mortality for leafless plants

mpaiao:
### Description:

These changes are mostly to be on the safe side. Hydraulic failure mortality is triggered when btran is too low (non-hydro mode) or when hydraulic conductivity is diminished (hydro mode). This pull request adds a check to see if the PFT is deciduous and is completely leafless. In these cases, hydraulic failure is forced to zero.
1. This change impacts both drought-deciduous and cold-deciduous PFTs.
2. In the future, we may want to coordinate hydraulic failure with phenology, to ensure that deciduous plants abscise leaves instead of dying when under high risk of hydraulic failure.

This may have minimal effect under severely dry conditions.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -89,14 +93,27 @@ subroutine mortality_rates( cohort_in,bc_in,btran_ft, mean_temp,             &
...
+    ! we prevent hydraulic failure mortality to occur as plants are leafless. For now, 
+    ! semi-deciduous plants with senescing leaves may still die of hydraulic failure, but in
+    ! the future we could accelerate senescence to avoid mortality. Note that both drought 
+    ! deciduous and cold deciduous are considered here to be consistent with the idea that
+    ! plants without leaves cannot die of hydraulic failure.
+    is_decid_dormant =                                                            & !
+       ( prt_params%stress_decid(cohort_in%pft) == ihard_stress_decid .or.        & ! Drought deciduous
+         prt_params%stress_decid(cohort_in%pft) == isemi_stress_decid .or.        & ! Semi-deciduous
+         prt_params%season_decid(cohort_in%pft) == itrue                  ) .and. & ! Cold deciduous
+       ( cohort_in%status_coh == leaves_off )                                     ! ! Fully abscised
```
rgknox, line 114: I think this looks good. I started to muse about what guidance we could offer in the future if we ever enable a long window of time for leaf drop, but that would be a full discussion and could not really be encapsulated in a couple of comments here.

rgknox (approved):

ZacharyRobbins (commented):
I think this code looks good. If my review of the btran functions is correct, I believe the explicit case this prevents is in non-hydro mode. When soil moisture is sufficiently low to trigger btran-hydraulic failure mortality in a plant that is not transpiring. I agree that is unlikely this would change much when FATES hydro is on., though I don't think its impossible for PLC to go up under leafless conditions. 

I guess my only larger science concern is that plants lose water even in the absence of leaves, and hydraulic failure would still be possible through stem permeabilty and loss through roots. Given the approximation that is btran, I don't think its unreasonable to make exception for this (as one would assume in a broad sense leaf loss greatly lowers the risk of HF).  But I wanted to voice this concern, particularly  that with 1. being that leafless plants cannot experience cavitation, which I don't think is explicitly true. 

Wolfe, B. T., Sperry, J. S., & Kursar, T. A. (2016). Does leaf shedding protect stems from cavitation during seasonal droughts? A test of the hydraulic fuse hypothesis. New Phytologist, 212(4), 1007-1018.

Thank you.

mpaiao:
Thanks for the feedback @ZacharyRobbins! I'm not very familiar with FATES-Hydro, but if the current FATES-Hydro implementation allows for plants to continue losing water when they are leafless, then we definitely don't want this change, at least when FATES-Hydro is on. Let me know and I can update/close the PR.

rosiealice:
So the loss of water from leafless plants is a big issue for frozen soils, which all else being equal, dehydrate the plants via the roots, leading to massive hydraulic failure. 

To prevent this, in Marius Lambert's work, the (minimum) stomatal conductance and the root conductivity needed to be  dialled down (a lot) as a function of hardening (from his earlier paper). 

Unfortunately this didn't make it to the trunk (Marius started working on river DOC immediately after finishing his thesis), and we are writing a proposal to follow it up. 

But the lessons learned remain valid, and I think maybe transferable to the super dry soil conditions also where we know that Piñon pines become hydraulically isolated when swp <-3Mpa, from all of the work in NM etc. 

- If we turn this option on in fates hydro it might help stop things dying in the off season (if that is desired) but there would be a potentially big pulse of mortality when the leaves came back on... we could prevent that by stopping the leaves coming in if the plc was high, but it would probably be better to stop the (bulk of the) dehydration process in the first place?

mpaiao:
Thanks for the feedback @rosiealice, and good to know that there is some more mechanistic approach to prevent the mortality. So maybe for now the best option is to implement my original suggestion only when FATES-Hydro is turned off, and note that in the future we will incorporate Marius' work?

ZacharyRobbins:
I think that makes sense @mpaiao .

glemieux:
@JessicaNeedham and @mpaiao, in deconflicting `main` against this PR branch, I was unsure how to reconcile the update from https://github.com/NGEET/fates/pull/1115.  The current code on `main` has the following:
https://github.com/NGEET/fates/blob/a9bbfd820bb613afd0c33fe75404c7848b435260/biogeochem/EDMortalityFunctionsMod.F90#L159-L162

If we have time, we can discuss together at the fates software meeting today.

glemieux:
This DIFF starts up 2.5 months into the 6 month tests and is occurring only on one gridcell on the coast of Antarctica for that whole time.  As such, I think this is good to integrate.

PR #1132, opened 2023-12-09
Exponential-based carbon starvation mortality function

mpaiao:
### Description:

The default FATES carbon starvation mortality is a linear function of carbon storage (relative to target leaf biomass). This formulation means that carbon starvation mortality can be significant even when cohorts still have storage (or carbon starvation mortality will be always low, triggering termination mortality too frequently).

The new formulation continues to assume a maximum mortality rate when storage is zero, but quickly declines following an exponential function, similarly to the growth-dependent mortality function by [Camac et al. (2018)](https://dx.doi.org/10.1073/pnas.1721040115), similar to what currently exists in ED2 too. Camac et al. (2018) used growth rates as predictor of carbon starvation mortality. However, the FATES implementation still uses the carbon storage relative to target leaf biomass. This choice was to ensure that deciduous PFTs do not die from carbon starvation when growth is halted because they have no leaves (or high-latitude plants when growth is limited during the winter).

A new parameter `fates_mort_cstarvation_model` was added to the FATES parameter file to switch between FATES' original formulation (`fates_mort_cstarvation_model=1`) or the exponential formulation (`fates_mort_cstarvation_model=2`). In both models, two additional parameters control the mortality:

- `fates_mort_scalar_cstarvation`. Same meaning in both formulations: the maximum carbon starvation mortality [yr<sup>&minus;1</sup> when storage is completely depleted. 
- `fates_mort_upthresh_cstarvation`. This parameter is now used, and its meaning depends on the carbon starvation model. 
   - In the original formulation (`fates_mort_cstarvation_model=1`), parameter `fates_mort_upthresh_cstarvation` defines the maximum storage biomass (relative to the target leaf biomass) above which carbon starvation is zero; the default value (1.0 for all PFTs) is consistent with the previously hard-coded assumption. 
   - In the exponential-based formulation (`fates_mort_cstarvation_model=2`), parameter  `fates_mort_upthresh_cstarvation` defines the "e-folding" storage biomass (relative to the target leaf biomass) for carbon starvation mortality (i.e., at  `fates_mort_upthresh_cstarvation`, carbon starvation mortality is e<sup>&minus;1</sup> &times; `fates_mort_scalar_cstarvation`. It is advisable to set `fates_mort_upthresh_cstarvation` much lower than 1 (e.g., 0.1 or so), for the model to work as intended.

### Collaborators:

Previous discussions with @xiangtaoxu 

No change in answer expected if `fates_mort_cstarvation_model=1` and `fates_mort_upthresh_cstarvation=1.0`. Changes expected otherwise.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


biogeochem/FatesAllometryMod.F90
JessicaNeedham, line 1: Maybe the description at the top of the subroutine could be simplified, but still explain that storage > target does not impact linear c starvation mortality (it would be 0) but allows it to continue to decay with the exponential model?

parameter_files/fates_params_default.cdl
```diff
@@ -433,7 +433,7 @@ variables:
 		fates_mort_scalar_hydrfailure:long_name = "maximum mortality rate from hydraulic failure" ;
 	double fates_mort_upthresh_cstarvation(fates_pft) ;
 		fates_mort_upthresh_cstarvation:units = "unitless" ;
-		fates_mort_upthresh_cstarvation:long_name = "threshold for storage biomass (relative to target leaf biomass) above which carbon starvation is zero" ;
+		fates_mort_upthresh_cstarvation:long_name = "It depends on fates_mort_cstarvation_model: 1. upper threshold for storage biomass (relative to leaf biomass) above CS mortality is zero. 2. storage biomass (relative to leaf biomass) at which CS mortality decays to e^-1 of the maximum" ;
```
JessicaNeedham, line 436: Can you put the suggested value here that was in the main PR comment?  E.g. in case 2 set to ~0.1 for the model to work as intended.

JessicaNeedham (approved):
Thanks @mpaiao this looks great! Hope it helps with the understory survival!

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -169,20 +172,32 @@ subroutine mortality_rates( cohort_in,bc_in,btran_ft, mean_temp,             &
...
+             cmort = EDPftvarcon_inst%mort_scalar_cstarvation(cohort_in%pft) * &
+                max(0.0_r8, (EDPftvarcon_inst%mort_upthresh_cstarvation(cohort_in%pft)-frac) / &
+                            EDPftvarcon_inst%mort_upthresh_cstarvation(cohort_in%pft) )
+
+          case (cstarvation_model_exp)
+             ! Exponential model.  The upper threshold really is the e-folding factor for
+             ! frac. We also make sure the mortality is set to zero when tiny.
+             cmort = EDPftvarcon_inst%mort_scalar_cstarvation(cohort_in%pft) * &
+                     exp(- frac / EDPftvarcon_inst%mort_upthresh_cstarvation(cohort_in%pft))
+             if (cmort <= nearzero) then
```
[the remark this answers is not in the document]
mpaiao: Sure, I will move the check outside the select case so it applies to all models.

biogeochem/FatesAllometryMod.F90
mpaiao: I believe this comment is for the calculation of carbon starvation mortality rates? Somehow it is showing in FatesAllometryMod.F90. 

If it was for the carbon starvation mortality, I edited the comments explaining how each model works — though I may have complicated rather than simplified... I will update the PR shortly, let me know if it looks fine or needs simplification.

parameter_files/fates_params_default.cdl
mpaiao: I am adding text in the next update of this PR.

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -166,24 +169,44 @@ subroutine mortality_rates( cohort_in,bc_in,btran_ft, mean_temp,             &
...
-          else
+
+          ! Select the carbon starvation mortality model (linear or exponential)s.
+          select case (mort_cstarvation_model)
+          case (cstarvation_model_lin)
+             ! Linear model. Carbon starvation mortality will be zero when fraction of
+             ! storage is greater than or equal to mort_upthresh_cstarvation, and will
+             ! increase to the maximum mortality (mort_scalar_cstarvation) when frac = 0.
+             cmort = EDPftvarcon_inst%mort_scalar_cstarvation(cohort_in%pft) * &
+                max(0.0_r8, (EDPftvarcon_inst%mort_upthresh_cstarvation(cohort_in%pft)-frac) / &
```
rgknox, line 188: I see EDPftvarcon_inst%mort_upthresh_cstarvation is already in the parameter file, that makes things a little easier: https://github.com/NGEET/fates/blob/sci.1.70.0_api.32.0.0/main/EDPftvarcon.F90#L1030-L1032 with a default of 1, which should give us b4b with base on existing tests.

rgknox: Thinking about how this upthresh parameter will affect results. So if you do set it lower than 1, then cstarvation won't kick in at all, until the plants storage fraction hits that upthresh fraction... but since we now divide by uptresh as well, the rate at which cstarvation ramps up with be increased once it does, so it should converge to the same rate as the storage fraction approaches zero.

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -166,24 +169,44 @@ subroutine mortality_rates( cohort_in,bc_in,btran_ft, mean_temp,             &
...
+                            EDPftvarcon_inst%mort_upthresh_cstarvation(cohort_in%pft) )
+
+          case (cstarvation_model_exp)
```
rgknox, line 191: This also applies to your allometry PR @mpaiao , but we should start talking about expanding our regression tests to capture these different parameter mode settings.  As of the seed-dispersal PR by @glemieux , we can now do this.  Since this PR does have a change to the  parameter file, it will need to come in with a host-side update to the default parameter file. We can add the test in with that PR. Maybe we will integrate this PR and the allometry PR in sequence, and/or maybe make a combo tag of the two, and update the ctsm/e3sm side default files and their tests together.

rgknox (approved):

biogeochem/EDMortalityFunctionsMod.F90
mpaiao: Correct, this is just to make the existing code general and continuous for any value of `mort_upthresh_cstarvation`, whilst keeping the solution identical to the previous code when `mort_upthresh_cstarvation = 1.0`

biogeochem/FatesAllometryMod.F90
JessicaNeedham: I meant it for the comments in  storage_fraction_of_target subroutine but it isn't too important. I just found it confusing that it first says it only calculates  fraction if storage is below target, but then says the cap no longer applies. But it's a minor point. 
I like the comments in EDMortalityMod - very helpful!

mpaiao: Ah, I see, it's just that your original comment was not showing next to any code, so I wasn't sure to where it was referring. I can edit the comment for storage_fraction_of_target too, thanks!

PR #1144, opened 2024-01-02
Revised quadratic equation solver for handling exceptions.

mpaiao:
Slightly revised the solvers for quadratic equations ( <i>a</i> <i>x</i><sup>2</sup> + <i>b</i> <i>x</i> + <i>x</i> = 0). The previous solver would fail under a few cases in which roots may actually exist (most notably, when <i>a</i> = 0). Also, the revised solver is slightly more robust for handling other specific cases, for example, when <i>b</i> = 0. In this case, the numerical recipe approach could lead to FPE even though two valid roots may exist.

The existing solver should be working fine for the cases in which the solver is being currently used, this revision is more to prevent issues if future implementations call the existing solvers. 

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -1790,6 +1790,8 @@ subroutine quadratic_f (a, b, c, r1, r2)
   ! !REVISION HISTORY:
   ! 4/5/10: Adapted from /home/bonan/ecm/psn/An_gs_iterative.f90 by Keith Oleson
   ! 7/23/16: Copied over from CLM by Ryan Knox
+  ! 12/30/23: Instead of issuing errors when a=0, solve the trivial cases too.
+  !           Check determinant sign, and stop the run when it is negative.
```
glemieux, line 1794: Add an explicit reference to the Numerical Solutions book in case the person reading the code isn't familiar with it.

glemieux (approved):
This looks great.  One minor request for updating the revision history comment.

biogeophys/FatesPlantRespPhotosynthMod.F90
mpaiao: Thanks for revising it, @glemieux! There was already a reference to the book in the comments, but for an earlier edition. I updated with the full reference and link.

rgknox:
@mpaiao , code looks good. It certainly handles exceptions more thoroughly and will "keep going" in more situations.  Has this code been tested in other contexts (like ED2 for instance)? Have you tried any long term simulations to stress test it? Its likely slower, given all the logic, the logic on the discriminant and the max call, but I'm fine with that if its not too noticeably slower given the upsides.

mpaiao:
@rgknox This PR brings in some of the exceptions handled in the ED2 code, but ED2 didn't handle one of the cases that FATES already handled. I actually made a PR to the ED2 code to include this exception.

I have not run many long-term simulations with the code, but I did test the routine offline for a broad number of special/challenging cases (e.g., a,b, and/or c being zero, two roots with extremely different orders of magnitude, negative discriminant), and the solver seems to handle all the cases correctly.

rgknox:
I ran a 5 year smoke test just to put it the solver through its paces and make sure it didn't generate any weird behavior. The maps of above ground biomass is roughly the same as that from a base run I created with two-stream a couple weeks back.  Merging

PR #1149, opened 2024-01-11
Alternative vertical scaling of leaf maintenance respiration 

JessicaNeedham:
This PR allows for variable vertical scaling of leaf maintenance respiration when using the Atkin et al. 2017 leaf MR model. The current model calculates top of canopy leaf MR following Atkin et al. 2017 and then scales through the canopy using the leaf nitrogen scaler from Lloyd et al. 2010. Recent observational evidence from [Lamour et al. 2023](https://nph.onlinelibrary.wiley.com/doi/full/10.1111/nph.18901) suggest that respiration might decrease more rapidly than photosynthesis through the canopy. In Lamour et al. 2023 they fit a linear model to observations of leaf MR by LAI, but here to avoid negative leaf MR we use the same negative exponential function as in Lloyd et al. 2010 but allow the curve to vary according to a new parameter fates_maintresp_leaf_decay. This should in principle improve the carbon budget of understory plants, and as a result increase LAI which has been biased low when using the Atkin respiration model.

### Collaborators:
@mpaiao @ckoven @jenniferholm @jennykowalcz

With the new parameter set to its default value of 2.43 results should be bfb with main. Decreasing this parameter decreases understory respiration and should decrease carbon starvation mortality in understory cohorts. I’m running a control and test with a version up to date with API 29 now and will post results.

### Checklist

*Contributor*
- [X ] The in-code documentation has been updated with descriptive comments
- [X ] The documentation has been assessed to determine if updates are necessary

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


JessicaNeedham:
This comparison is the updated parameter file with either the fates_maintresp_leaf_decay set to 2.43 (the current default in FATES) or 0.75 its default value at 1.8225. The run is 40 years in now but we see the carbon budget of understory plants is improved (GPP - respiration) is now more positive in the amazon, understory mortality is generally lower, and LAI has increased in the tropics. 
Screenshot 2024-01-12 at 15 24 08

Screenshot 2024-01-12 at 15 18 17

rosiealice:
That looks promising @JessicaNeedham 
What is the effect on CUE? (like, does it push it out of reasonable range or not?)

JessicaNeedham:
> That looks promising @JessicaNeedham What is the effect on CUE? (like, does it push it out of reasonable range or not?)

In this test with the default parameters, the reduced MR in the understory increases CUE by about 0.06 or so. 

Screenshot 2024-01-15 at 15 19 06
Screenshot 2024-01-15 at 15 19 13


In runs with two stream radiation, Atkin respiration, photosynthesis temperature acclimation and reduced carbon starvation mortality, the difference in CUE is smaller, and CUE in the Amazon is higher (with and and without new vertical scaling of  MR). 

Screenshot 2024-01-15 at 15 25 41

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -2148,34 +2150,50 @@ subroutine LeafLayerMaintenanceRespiration_Atkin_etal_2017(lnc_top, &
...
    ! all figs in Atkin et al 2017 stop at zero Celsius so we will assume acclimation is fixed below that
    r_0 = EDPftvarcon_inst%maintresp_leaf_atkin2017_baserate(ft)
-   r_t_ref = max( 0._r8, nscaler * (r_0 + lmr_r_1 * lnc_top + lmr_r_2 * max(0._r8, (tgrowth - tfrz) )) )
-   
+   rdark_decay = EDPftvarcon_inst%maintresp_leaf_decay(ft)
+
+   ! This code uses the relationship between leaf N and respiration from Atkin et al 
+   ! for the top of the canopy, but then scales through the canopy based on a rdark_scaler.
+   ! To assume proportionality with N through the canopy following Lloyd et al. 2010, use the
+   ! default parameter value of 2.43, which results in the scaling of photosynthesis and respiraiton
```
mpaiao, line 2187:
```suggestion
   ! default parameter value of 2.43, which results in the scaling of photosynthesis and respiration
```

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -2148,34 +2150,50 @@ subroutine LeafLayerMaintenanceRespiration_Atkin_etal_2017(lnc_top, &
...
+   ! being proportional through the canopy. To have a steeper decrease in respiration than photosynthesis
+   ! this number can be smaller. There is some observational evidence for this being the case
+   ! in Lamour et al. 2023. 
+
+   kn = exp(0.00963_r8 * vcmax25top - rdark_decay)
```
mpaiao, line 2192: I know this is borrowed from Lloyd et al. (2010), but I wonder if the slope (0.00963) should be also made a parameter as they are both empirical.

mpaiao (approved):
@JessicaNeedham Thanks for this PR, I think it will be very helpful for predicting an understorey in FATES! I made just one minor suggestion on parameters, though maybe it could be part of a future PR that turns all the empirical terms associated with trait plasticity into parameters.

biogeophys/FatesPlantRespPhotosynthMod.F90
JessicaNeedham: The kn equation is also used in other places in the photosynthesis subroutines. Would we want two sets of parameters - one for the photosynthesis scaling and one for respiration? Or just stick to respiration here and keep photosynthesis scaling parameters hard coded?

parameter_files/fates_params_default.cdl
```diff
@@ -386,7 +386,10 @@ variables:
 		fates_maintresp_leaf_atkin2017_baserate:long_name = "Leaf maintenance respiration base rate parameter (r0) per Atkin et al 2017" ;
 	double fates_maintresp_leaf_ryan1991_baserate(fates_pft) ;
 		fates_maintresp_leaf_ryan1991_baserate:units = "gC/gN/s" ;
-		fates_maintresp_leaf_ryan1991_baserate:long_name = "Leaf maintenance respiration base rate per Ryan et al 1991" ;
+		fates_maintresp_leaf_ryan1991_baserate:long_name = "Leaf maintenance respiration base rate per Ryan et al 1991" ;	
+        double fates_maintresp_leaf_decay(fates_pft) ;
```
rgknox, line 390: I like the introduction of the parameter, but when I see the word "decay" I think of a biological decomposition process. I wonder if "attenuation" would work here, or something similar that conveys how the rate changes over depth.   fates_maintresp_vert_atten?   fates_maintresp_vert_scaler ?

JessicaNeedham: I've changed it to maintresp_leaf_vert_scaler_coeff1 and maintresp_leaf_vert_scaler_coeff2.

biogeophys/FatesPlantRespPhotosynthMod.F90
JessicaNeedham: I've pulled out the other parameter used in the rdark scaling. The other kn parameters are still hard coded.

JessicaNeedham:
I'm adding in those two additional parameters for the nscaler now. 
The respiration parameters are fates_maintresp_leaf_vert_scaler_coeff1 and 2. 
Should these ones be: 
fates_photo_vert_scaler_coeff1 and 2? or fates_leafn_vert_scaler_coeff1?

rgknox:
All tests pass with one exception that is irrelevant, see: https://github.com/ESCOMP/CTSM/issues/2491

Issue #1151, opened 2024-01-22
FATES-Hydro restarts fail with drought-deciduous PFT


jennykowalcz:
As raised in [this discussion](https://github.com/NGEET/fates/discussions/1133), FATES-Hydro fails upon restart with a drought-deciduous PFT, with error messages saying is recruitment = F, LAI = 0, stem_water = 0, w_tot_* = NaN, and LWP = NaN. For example,
```
 42:  Could not find a stable solution for hydro 1D solve
 42:  
 42:  error code:            1
 15:  error diag:   0.000000000000000E+000  0.000000000000000E+000
 15:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 96:  error diag:   0.000000000000000E+000  0.000000000000000E+000
 96:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 15:  lat:   -23.6842105263159      longitidue:   302.500000000000     
 15:  is recruitment:  F
 15:  layer:            4
 15:  wb_step_err =                      NaN
 15:  q_top_eff*dt_step =   0.000000000000000E+000
 15:  w_tot_beg =                      NaN
 15:  w_tot_end =                      NaN
 15:  leaf water:                      NaN  kg/plant
 15:  stem_water:   0.000000000000000E+000  kg/plant
 15:  troot_water:   0.000000000000000E+000
 15:  aroot_water:   0.000000000000000E+000
 15:  LWP:                      NaN
 15:  dbh:    3.49675824928844     
 42:  error diag:   0.000000000000000E+000  0.000000000000000E+000
 42:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 15:  pft:            4
 15:  z nodes:    4.24889198027262        2.09944599013631     
 15:  -0.190993887372315      -0.227329413955693      -0.227329413955693     
 15:  psi_z:                      NaN  2.057457070350210E-002 -1.871740096248686E-003
 15:  -2.227828256764042E-003 -2.227828256765796E-003
 15:  vol,    theta,   H,  Psi,     kmax-
 15:  flux:            0.000000000000000E+000
 15:  l:  0.000000000000000E+000                     NaN                     NaN
 15:                      NaN
 15:                          6.053417126477062E-003
 15:  s:  1.888056051911535E-003  0.000000000000000E+000  -5551.12473567488     
 15:   -5551.14531024558     
 15:                          2.339234678527242E-002
[...]

Issue #701, opened 2020-09-24
FATES_ERRH2O_SCPF fails COMPARE_base_rest


glemieux:
This error was discovered after correcting another error in the `FatesHydro` test mod in which no history files where being generated for the `ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruGs.cheyenne_intel.clm-FatesHydro` test case (see https://github.com/ESCOMP/CTSM/issues/1160).  The test case runs, but fails the restart on `FATES_ERRH2O_SCPF` only:
```
FATES_ERRH2O_SCPF   (lndgrid,fates_levscpf,time)  t_index =      5     5
         12      156  (     0,    41,     1) (     1,     1,     1) (     0,    67,     1) (     0,   145,     1)
                 156   3.761678083316676E-16  -6.113989692399947E-16 1.0E-16  6.578333154084754E-17 6.4E-02 -2.485285346918768E-19
                 156   4.620813013804872E-16  -5.128924256061086E-16         -3.811208864302963E-17          3.013937768871115E-19
                 156  (     0,    41,     1) (     1,     1,     1)
          avg abs field values:    8.833762855196752E-18    rms diff: 1.6E-17   avg rel diff(npos):  6.4E-02
                                   7.583820984915835E-18                        avg decimal digits(ndif):  0.2 worst: -0.3
 RMS FATES_ERRH2O_SCPF                1.5775E-17            NORMALIZED  1.9217E+00
```

rgknox:
I'm wondering if  maybe it is not reasonable for some error checks, particularly the small ones where the absolute values are  close to machine precision, and thus the relative differences can be large.

billsacks:
@rgknox I'm not sure if I understand your thinking, but it shouldn't matter if the absolute value is close to machine precision. Machine precision of ~ 1e-16 means that values of 1 and (1+1e-16) are indistinguishable. But double precision values can represent values down to about 1e-308. If you have a value like 1e-17, then that's indistinguishable from (1e-17 + 1e-33), but it should be very distinguishable from, say, 1.1e-17.

ekluzek:
Yes, this looks like a legit issue to me. We do expect exact results between the baseline and restart cases for all history fields. And actually the error is large compared to the average values of the data. Does FATES_ERRH2O_SCPF need to be added to the restart file?

rgknox:
[...]

PR #1156, opened 2024-02-02
prevent FATES-Hydro from failing upon restart with a drought deciduous PFT

jennykowalcz:
### Description:
As discussed in [issue 1151 ](https://github.com/NGEET/fates/issues/1151), FATES-Hydro was failing upon restart for simulations including a drought-deciduous PFT, with error messages saying is recruitment = F, LAI = 0, stem_water = 0, w_tot_* = NaN, and LWP = NaN. This PR implements a fix suggested by @mpaiao to prevent the failure by preventing the leaf elongation factor from being zero when the Fates-Hydro code is calculating water storage volume.

### Collaborators:
@mpaiao @ckoven 

I tested this PR using a small region in the central Amazon with a control case from the main branch and a test case from this branch, for sets of simulations with a single evergreen PFT and a single drought-deciduous PFT. The control cases comprised of 20 year runs (no restart) while the test cases reached 20 years with restarts every 2 years. Changes are near-zero for the evergreen PFT for the quantities I looked at. For the drought-deciduous PFT quantities change by up to a couple percent.
[Hydro restart test.pdf](https://github.com/NGEET/fates/files/14143156/Hydro.restart.test.pdf)

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeophys/FatesPlantHydraulicsMod.F90
+++ b/biogeophys/FatesPlantHydraulicsMod.F90
@@ -936,7 +936,7 @@ contains
     ! we set as our lower cap on leaf volume
     real(r8), parameter :: min_trim      = 0.1_r8   ! The lower cap on trimming function used
     ! to estimate maximum leaf carbon
-    real(r8), parameter :: min_efleaf = 0.1_r8 ! Lower cap on leaf elongation factor
+
 
     ccohort_hydr => ccohort%co_hydr
     ft           = ccohort%pft
@@ -982,7 +982,7 @@ contains
     ! Lets also avoid super-low targets that have very low trimming functions
 
     call bleaf(ccohort%dbh,ccohort%pft,ccohort%crowndamage, &
-         max(ccohort%canopy_trim,min_trim),max(ccohort%efleaf_coh,min_efleaf), leaf_c_target)
+         max(ccohort%canopy_trim,min_trim),1.0, leaf_c_target)
 
     ccohort_hydr%v_ag(1:n_hypool_leaf) = max(leaf_c,min_leaf_frac*leaf_c_target) * &
          prt_params%c2b(ft) / denleaf/ real(n_hypool_leaf,r8)
--- a/biogeophys/FatesPlantHydraulicsMod.F90
+++ b/biogeophys/FatesPlantHydraulicsMod.F90
@@ -936,6 +936,7 @@ contains
     ! we set as our lower cap on leaf volume
     real(r8), parameter :: min_trim      = 0.1_r8   ! The lower cap on trimming function used
     ! to estimate maximum leaf carbon
+    real(r8), parameter :: min_efleaf = 0.1_r8 ! Lower cap on leaf elongation factor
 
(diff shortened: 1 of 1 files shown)


rgknox:
My understanding is that this change will ensure that all plants have at least some nominal volume to the leaf compartment at all times.  That sounds good, it looks like this was not being applied in all scenarios.

The only thing that is throwing me off a little bit, is that we are saying that minimum allowed value is 0.1 of the target amount of leaf at the actual leaf elongation, which is also protected with a minimum value of 0.1.  So this minimum has 0.1*0.1=0.01 of the target biomass (volume).  I don't know if that is too low or too high, but I think a larger value may be more stable and will have little impact on results.  

Essentially, this compartment volume "v_ag" dictates the capacity for holding water. And we are accepting that when leaves are dropped, we are maintaining kind of a ghost volume that stores water there even when the leaves are gone.  We do this for numerical stability. The conductance should still be low or zero and more representative of any losses through the petiole connection points (if we even represent that).  In light of this, I think we might do better by making this relationship a little simpler, and maybe allow a larger ghost volume? Maybe have the minimum volume be keyed off the leaf target at full extension and then apply min_leaf_frac to that?

```
  call bleaf(ccohort%dbh,ccohort%pft,ccohort%crowndamage, &
         max(ccohort%canopy_trim,min_trim),ccohort%efleaf_coh, leaf_c_target)
         max(ccohort%canopy_trim,min_trim),1.0, leaf_c_target)

    ccohort_hydr%v_ag(1:n_hypool_leaf) = max(leaf_c,min_leaf_frac*leaf_c_target) * &
         prt_params%c2b(ft) / denleaf/ real(n_hypool_leaf,r8)
```

mpaiao:
Good point, @rgknox I think your suggestion would be a good implementation. This is likely a bug I introduced when I made efleaf_coh an argument in bleaf, before that change the code probably used to implicitly assume fully flushed leaves for this calculation.

jennykowalcz:
Thanks @rgknox and @mpaiao , this makes sense and sounds like a good idea. I gave it a quick test and the differences vs control look similar to what I posted earlier.

rgknox (approved):
looks good to me too

glemieux:
Regression testing on `derecho` results against `fates-sci.1.72.4_api.34.0.0-ctsm5.1.dev174` baseline shows b4b results for all expected tests.  Note that the current hydro regression tests are failing due to a fix waiting to come in via#1164.

glemieux:
In an effort to test hydro b4b-ness of this, I tested this against `fates-sci.1.72.0_api.33.0.0-ctsm5.1.dev170` using `ERS_D_Ld5.1x1_brazil.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdHydro`.  While the the exact restart issue recorded in #701 is still true, it is consistent and the comparison against the baseline was b4b. 

This is ready to integrate.

Issue #1163, opened 2024-02-22
SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruRsGs.izumi_intel.clm-FatesColdHydro fails when t_soisno is initialized to 272K instead of 274K


olyson:
A change is planned for the host model CTSM such that soil temperature for a cold start will be initialized to 272K instead of the current 274K.  Purpose is to avoid locking in large amounts of deep soil carbon at high latitudes.  Testing indicated a failure for SMS_Lm3_D_Mmpi-serial.1x1_brazil.I2000Clm50FatesCruRsGs.izumi_intel.clm-FatesColdHydro, caused by a divide by zero in FatesHydroWTFMod.F90 at this line:

`psi = this%psi_sat*(th/this%th_sat)**(-this%beta)`

where th is zero.
Please see https://github.com/ESCOMP/CTSM/issues/2373 for more details.
Plan is that @rgknox has graciously agreed to provide a fix for this that @olyson will test in his current configuration.  For the purposes of integrating the soil temperature change in CTSM, the test in question will be labeled as an ExpectedFail.

olyson:
I have tested a fix provided by @rgknox and it resulted in the test passing.

Issue #1154, opened 2024-01-29
FATES-Hydro: NaN/Inf error for soil matric potential


jennykowalcz:
I'm encountering this error 2 years into a pan-Amazon no-comp FATES-Hydro run with evergreen PFTs:
```
  0: PIO: FATAL ERROR: Aborting... An error occured, Writing variables (number of variables = 429) to file (./nocomp_hydro_
2pfts_test2.elm.h0.0002-11.nc, ncid=142) using PIO_IOTYPE_PNETCDF iotype failed. Non blocking write for variable (FATES_ROO
TWGT_SOILMATPOT, varid=288) failed (Number of subarray requests/regions=1, Size of data local to this process = 192). NetCD
F: Numeric conversion not representable (err=-60). Aborting since the error handler was set to PIO_INTERNAL_ERROR... (/glob
al/u2/j/jkowalcz/E3SM-test/E3SM/externals/scorpio/src/clib/pio_darray_int.c: 395)
```

Uncommenting [this check on capillary pressure](https://github.com/NGEET/fates/blob/a9bbfd820bb613afd0c33fe75404c7848b435260/main/FatesHistoryInterfaceMod.F90#L5102) didn't help - I still get the above error. So I am confused about what's happening. Has anyone else encountered this error? @xuchongang @ZacharyRobbins and others who use FATES-Hydro, which [soil water retention function](https://github.com/NGEET/fates/blob/a9bbfd820bb613afd0c33fe75404c7848b435260/biogeophys/FatesPlantHydraulicsMod.F90#L400) are you using?

jennykowalcz:
[PR 1164](https://github.com/NGEET/fates/pull/1164) with fates_hydro_solver=2 resolves this error.

PR #1164, opened 2024-02-22
Small number protections to CCH water transfer functions

rgknox:
A cap has been created on the smallest suction that we are willing to simulate with CCH (Campbell Clapp-Hornberger) water transfer functions. These methods determine the equivalent water content, so that if values below this water content are passed to the functions, it operates at the lowest threshold instead of breaking.  The current threshold is -15 MPa. This is incredibly small, and should generate conductances that indistinguishable from conductances generated with near zero water contents.

A further modification was made in a second round of development here. All water retention curves are now designed to have a minimum value. For CCH it is -15 MPa, for Van Genucten it is the residual water content (existing parameter). We now force the fraction of maximum conductivity to be zero at the minimum value, this currently includes stomatal conductance.  This is applied by determining a weighting function that blends in a zero value. The weighting is 1 at the minimum suction, and decreases to zero on an exponential. I used an exponent that attenuates to a value of about 0.175 after increasing 1 MPa above minimum, and 0.025 after 2 MPa above minimum.  This function makes the derivative of conductance with suction continuous and smooth, but most importantly, it prevents the flow of water in plant and soil media that are completely parched.  

This method of ensuring fluxes are zero at these super low water contents will not affect re-hydration, because conductance between nodes is driven by the suction of the up-stream compartment (ie the one with more total potential, or less suction).  However, this will prevent completely desicated compartments from going into negative water content.

### Collaborators:
@olyson @jennykowalcz 

This should prevent crashes, particularly in soil water calls using frozen soils.

Fixes: #1168 
Fixes: #1163

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


rgknox:
This may address: #1154

jennykowalcz:
> This may address: #1154

I tried this out -- in the control case, without this PR, the case fails with this error:

```
128: PIO: FATAL ERROR: Aborting... An error occured, Writing variables (number of variables = 425) to file (./hydro_nocomp
_2pft_campbell_control.elm.h0.0002-11.nc, ncid=141) using PIO_IOTYPE_PNETCDF iotype failed. Non blocking write for variabl
e (FATES_ROOTWGT_SOILMATPOT, varid=274) failed (Number of subarray requests/regions=1, Size of data local to this process 
= 288). NetCDF: Numeric conversion not representable (err=-60). Aborting since the error handler was set to PIO_INTERNAL_E
RROR... (/global/u2/j/jkowalcz/E3SM-test/E3SM/externals/scorpio/src/clib/pio_darray_int.c: 395)

```

In the test case, pulling in the changes to FatesHydroWTFMod.F90, it fails with this error:

```
 27:  Site plant water balance does not close
 27:  delta plant storage:  -4.101476403268849E-005  [kg/m2]
 27:  integrated root flux:  -2.537539597419707E-002  [kg/m2]
 27:  transpiration flux:   3.019638043594088E-005  [kg/m2]
 27:  end storage:   1.310177109332794E-002
 27:  pre_h2oveg  1.314278585736063E-002
 27:  ENDRUN:
 27:  ERROR in FatesPlantHydraulicsMod.F90 at line 2808  
```                            

In both cases I'm using [this E3SM branch](https://github.com/E3SM-Project/E3SM/commits/master/) with [this FATES branch](https://github.com/NGEET/fates/tree/0b105bec7a90c5e02d1098f9bbdcf255f358aaa4) (with the change to FatesHydroWTFMod for the test case). I can try also testing with this PR and the [hydro stability PR](https://github.com/NGEET/fates/pull/851).

The region used in these runs is a lat/lon rectangle from 15 N - 25 S and 30 W - 85 W so it includes very dry regions outside the Amazon basin where FATES-Hydro is unhappy.

glemieux (approved):

glemieux:
Running regression tests on derecho I found that while this resolves #1163, I'm seeing a run failure due to the 1D hydro solver failing with a large `wb_step_error`:
```
2461  Could not find a stable solution for hydro 1D solve
2462 
2463  error code:            1
2464  error diag:    0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000
2465  lat:   -7.0000000000000000      longitidue:   305.00000000000000
2466  is recruitment:  F
2467  layer:            7
2468  wb_step_err =    1.6512024553098321E+139
2469  q_top_eff*dt_step =    0.0000000000000000
2470  w_tot_beg =    102.64162586485499
2471  w_tot_end =    1.6512024553098321E+139
2472  leaf water:    6.3082104539614020E-002  kg/plant
2473  stem_water:    5.2044616695225525E-002  kg/plant
2474  troot_water:    1.6178663624130458E-002
2475  aroot_water:    1.8182979624968509E-002
2476  LWP:   -2.1266161551233220
2477  dbh:   0.40361081540638905
2478  pft:            6
2479  z nodes:    1.2616446404957407       0.60582232024787031      -0.19099388737231493      -0.79999999999999993      -0.79999999999999993
2480  psi_z:    1.2364117476858283E-002   5.9370587384291323E-003  -1.8717400962486863E-003  -7.8399999999999997E-003   0.0000000000000000
2481  vol,    theta,   H,  Psi,     kmax-
2482  flux:             0.0000000000000000
2483  l:   8.1613119481785556E-005  0.77294073477601488        3.3911760124701393        3.3788118949932811
2484                           4.3843160661903065E-004
2485  s:   8.0363738039128391E-005  0.64761318929546885       -3.6111596928010109E-002  -4.2048655666439241E-002
2486                           1.2058111921054136E-003
2487  t:   2.4966046503811633E-005  0.64802665578870955       -3.6564678574268254E-002
2488                           8.0381337022338061E-004
2489  a:   1.3824320409604611E-006  0.75035604020636570       -2.2154089112156015E-003
2490                     in:   1.6033112452206815E-003
2491                    out:   3.0709050878900186E-006
2492  r1:  0.21880148501536775       0.46850359885155290        6.9272904988199870E+156
```
I'm guessing that this is due to something causing `dth_node` to blow up within `Hydraulics_Tridiagonal`?

Note that the log stops writing early due to #1168.

jennykowalcz:
> > This may address: #1154
> 
> I tried this out -- in the control case, without this PR, the case fails with this error:
> 
> ```
> 128: PIO: FATAL ERROR: Aborting... An error occured, Writing variables (number of variables = 425) to file (./hydro_nocomp
> _2pft_campbell_control.elm.h0.0002-11.nc, ncid=141) using PIO_IOTYPE_PNETCDF iotype failed. Non blocking write for variabl
> e (FATES_ROOTWGT_SOILMATPOT, varid=274) failed (Number of subarray requests/regions=1, Size of data local to this process 
> = 288). NetCDF: Numeric conversion not representable (err=-60). Aborting since the error handler was set to PIO_INTERNAL_E
> RROR... (/global/u2/j/jkowalcz/E3SM-test/E3SM/externals/scorpio/src/clib/pio_darray_int.c: 395)
> ```
> 
> In the test case, pulling in the changes to FatesHydroWTFMod.F90, it fails with this error:
> 
> ```
>  27:  Site plant water balance does not close
>  27:  delta plant storage:  -4.101476403268849E-005  [kg/m2]
>  27:  integrated root flux:  -2.537539597419707E-002  [kg/m2]
>  27:  transpiration flux:   3.019638043594088E-005  [kg/m2]
>  27:  end storage:   1.310177109332794E-002
>  27:  pre_h2oveg  1.314278585736063E-002
>  27:  ENDRUN:
>  27:  ERROR in FatesPlantHydraulicsMod.F90 at line 2808  
> ```
> 
> In both cases I'm using [this E3SM branch](https://github.com/E3SM-Project/E3SM/commits/master/) with [this FATES branch](https://github.com/NGEET/fates/tree/0b105bec7a90c5e02d1098f9bbdcf255f358aaa4) (with the change to FatesHydroWTFMod for the test case). I can try also testing with this PR and the [hydro stability PR](https://github.com/NGEET/fates/pull/851).
> 
> The region used in these runs is a lat/lon rectangle from 15 N - 25 S and 30 W - 85 W so it includes very dry regions outside the Amazon basin where FATES-Hydro is unhappy.

Unfortunately I also get this error with a domain restricted to the Amazon Basin and when including @xuchongang's hydro stability branch. Also, I forgot to mention that I am using  fates_hydro_solver = 2 .

glemieux:
I should note that the regression tests for this where all b4b otherwise.

derecho location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1164-fates-correctbase`

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -751,8 +754,12 @@ function th_from_psi_cch(this,psi) result(th)
     if(psi>this%psi_max) then
         ! Linear range for extreme values
         th = this%th_max + (psi-this%psi_max)/this%dpsidth_max
-    else
-        th = this%th_sat*(psi/this%psi_sat)**(-1.0_r8/this%beta)
+     else
+        if(psi<this%psi_min) then
+           th = this%th_min
```
mpaiao, line 759: Wouldn't this artificially add water to soils?

rgknox: yes, but the intention is to go from a value of zero to an inconsequentially small non-zero value. @ckoven and I were talking, this could potentially be problematic if we artificially add small amounts of water and that enables or drives non-zero fluxes that would had otherwise been zero. One possibility would be to force fractional conductivity to zero where it hits theta min.  That seems messy though. Honestly, I think if we just use a suitably small minimum value, one that is an incredibly small water content, yet can still accomodate the math needed to switch between psi and theta, this should be sufficient. Perhaps we can try minimum capping at -15 MPa or smaller, perhaps -9 is too high.

rgknox:
These changes seem to be adding some stability.  I ran a 25 year smoke test on the F45 grid with ELM-FATES-Hydro.  Vegetation was initialized as a cold start using the specified stem diameter (which I specified as the diameter at 75% maximum height) and closed canopy assumption, along with fixed biogeography and no-competition assumptions.   All parameters were defaults, with the one exception being the use of the Picard hydraulics solver. 

The model completed the smoke test. I show some mean fluxes and states from the final year of simulation. I interpret from the output that the hydraulics model continued to generate fluxes within the realm of realism through the final year of simulation.  Note that contrary to the Picard solver,  the 1D Taylor solve (default) did crash after several years, having trouble generating a solution within the error tolerance. 
[f45-hydrotest-biogeonocomp-yr23.pdf](https://github.com/NGEET/fates/files/15004091/f45-hydrotest-biogeonocomp-yr23.pdf)

rgknox:
I think this PR is ready for another round of review, and has changed significantly since the first round.  I plan no further changes beyond what comes out of review.

jennykowalcz:
Wow, this is great @rgknox ! When perlmutter comes back up I'll try rerunning my various failed hydro cases with this branch and hopefully close several of my hydro Issues.
Should we make the Picard 2D solver the default?

jennykowalcz:
> Wow, this is great @rgknox ! When perlmutter comes back up I'll try rerunning my various failed hydro cases with this branch and hopefully close several of my hydro Issues. Should we make the Picard 2D solver the default?

This PR with fates_hydro_solver=2 allows the cases to run successfully. Like you found, with fates_hydro_solver=1, we still get "Could not find a stable solution for hydro 1D solve" errors in dry gridcells

bchristo:
Chiming in here.
This is fantastic news and sounds to me like the Picard solver should be
the default.

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -44,7 +50,8 @@ module FatesHydroWTFMod
                                             ! elastic-caviation region
 
 
-  real(r8), parameter :: min_theta_cch = 0.01_r8 ! Minimum theta (matches ctsm)
+  !real(r8), parameter :: min_theta_cch = 0.05_r8 ! Minimum theta (matches ctsm)
+  real(r8), parameter :: min_psi_cch = -20.0_r8 ! Minimum theta (matches ctsm)
```
bchristo: update comment to say 'Minimum psi'

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -713,9 +781,11 @@ subroutine set_wrf_param_cch(this,params_in)
     this%psi_max     = this%psi_from_th(this%th_max-tiny(this%th_max))
     this%dpsidth_max = this%dpsidth_from_th(this%th_max-tiny(this%th_max))
 
-    this%th_min      = min_theta_cch
-    this%psi_min     = this%psi_from_th(min_theta_cch+tiny(this%th_max))
-    this%dpsidth_min = this%dpsidth_from_th(min_theta_cch+tiny(this%th_max))
+    !this%th_min      = min_theta_cch
+    !this%psi_min     = this%psi_from_th(min_theta_cch+tiny(this%th_max))
+    this%psi_min     = min_psi_cch
+    this%th_min      = this%th_from_psi(min_psi_cch+tiny(this%th_max))
```
bchristo: what's the purpose of the tiny(this$th_max)? Wondering if this is an error, since the input to this function should be only water potential.

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -1248,21 +1353,22 @@ end function dpsidth_from_th_smooth_cch
...
+    real(r8) :: pc
+    real(r8) :: kr
+    real(r8) :: dkr_dP
+    real(r8) :: sat_res
+    real(r8) :: alpha
+    real(r8) :: lambda
+    real(r8) :: Se
+    real(r8) :: deltaPc
+    real(r8) :: dSe_dpc
+    real(r8) :: dkr_dSe
```
bchristo, line 1374: Just a minor comment. My memory is a bit fuzzy on my Brooks-Corey model without looking it up; can comments on these variables be made to ease interpretability? And an overall comment citing a companion source for implementing the three main different regimes (full, smoothing, saturated).

biogeophys/FatesHydroWTFMod.F90
rgknox: originally, this tiny() function was used to prevent a problem with order of operations. Inside these functions we have "if statements" that compare theta and psi to these very psi_min and th_min thresholds we are calculating. By adding or subtracting a tiny, we were preventing the intering of logical blocks that were not yet initialized.  But with these changes, we actually don't really use these tiny's as much

biogeophys/FatesPlantHydraulicsMod.F90
```diff
@@ -2767,6 +2784,20 @@ subroutine hydraulics_bc ( nsites, sites, bc_in, bc_out, dtime)
            end do
 
            ! Second pass, apply normalized weighting factors for fluxes
+
```
xuchongang, line 2787: Good catch for low water content.  If the sum of weight is close to zero, then the layer-specific water uptake is proportion of the root x water content here.  It might be better to  explain more on the new disaggregation

xuchongang (commented):
I finished my review. It looks great to me (on minimum water content and forced zero conductance) and I think it should resolve the issue of very dry conditions.

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -341,7 +354,36 @@ function th_linear_res(this,psi) result(th)
...
+    ! This method is to be used with any and all water transfer functions
+    
+    real(r8),intent(in) :: psi                  ! MPa suction
+    real(r8),intent(in) :: psi_min              ! Minimum value of psi we track, value that pins to ftc_min
+    real(r8),intent(out):: min_ftc_weight       ! weighting factor for min_ftc
+    real(r8),intent(out):: dmin_ftc_weight_dpsi ! derivative of weighting factor wrt psi
+
+    ! If the difference between psi and psi-min is greater than 10MPa
+    ! just assume there is no effect of the minimum function (ie weight 0)
+    min_ftc_weight = exp(min_ftc_scalar*max(psi_min-psi,-10._r8))
```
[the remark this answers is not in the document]
rgknox: When psi gets to about 2 MPA from psi_min, the weighting factor gets incredibly small, the 10 MPa here is just to avoid numerical problems with that exponent function.  We could test the min_ftc_scalar though.. I don't expect the choice of min_ftc_scalar to have any noticable impact on model output, but I could try a few different values to see if anything happens.  We have the utility parameter  [fates_dev_arbitrary](https://github.com/NGEET/fates/blob/main/parameter_files/fates_params_default.cdl#L720-L722) that could aid in this.

biogeophys/FatesHydroWTFMod.F90
rgknox: I don't believe this is a mass conservation liability. We use these functions to drive fluxes, they never actually over-write the original masses of each pool.  It is the fluxes that update the pools.  What will happen is that water contents below the psi_min or theta_min values will just end up having zero flux out of that pool, and thus remain unchanged.

For instance here is where we update the actual soil water content, it is based off of fluxes:
https://github.com/NGEET/fates/blob/main/biogeophys/FatesPlantHydraulicsMod.F90#L2729-L2730

biogeophys/FatesHydroWTFMod.F90
rgknox: I created an issue here: https://github.com/NGEET/fates/issues/1193
I would just add the changes in here, but I'm not familiar with that model.

biogeophys/FatesHydroWTFMod.F90
```diff
@@ -44,6 +44,8 @@ module FatesHydroWTFMod
                                             ! elastic-caviation region
 
 
+  real(r8), parameter :: min_psi_cch = -9._r8 ! Minimum psi we are willing to track in cch
```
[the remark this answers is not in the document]
rgknox: I think this was actually updated, seems like a github glitch

functional_unit_testing/hydro/HydroUTestDriver.py
```diff
@@ -150,18 +150,60 @@ def __init__(self,index,p50,avuln):
...
+    zsapric= 0.5
+    om_watsat         = max(0.93 - 0.1   *(zsoi/zsapric), 0.83)
+    om_bsw            = min(2.7  + 9.3   *(zsoi/zsapric), 12.0)
+    om_sucsat         = min(10.3 - 0.2   *(zsoi/zsapric), 10.1)
+
+    #om_sucsat = SuctionMMtoMPa(om_sucsat)
+    
+    return(om_watsat,om_sucsat,om_bsw)
+    
+def CCHParmsCosby84T5(zsoi,om_frac,sand_frac,clay_frac):
```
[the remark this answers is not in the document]
rgknox: added

biogeophys/FatesPlantHydraulicsMod.F90
rgknox: I updated the comment a little, should be more accurate. Does this look good @xuchongang ?

rgknox:
tests ok, tests_0501-150922de

PR #1171, opened 2024-03-13
Small fire refactor (fire weather abstract class), plus unit test

adrifoster:
I am planning to create a separate fire weather type (Canadian Fire Weather Indices) so in preparation for that I created a fire weather abstract class to use instead of the current method (subroutine modifying a site-level variable). 

### Description:
The fire weather class is still a site-level variable, but has its own methods for updating the fire weather index.

I also added some unit tests for the Nesterov Index fire weather class.
One note, in order to `use` the parameters for the Nesterov calculation inside the unit test, I had to make those specific values no longer `protected`. Otherwise we would have to read in the whole parameter file for the unit test - which I don't think is a good idea. 

Another option would be to make the fire weather index subroutine take in a fire parameter class, and then we could just set it up by hand in the unit test. This would be a pretty big change (i.e. making the fire parameter class), but is maybe a good idea anyway. 

Open to thoughts by @rgknox or @glemieux or others. 

None, all should be b4b

### Checklist

*Integrator*

### Documentation
- [Technical Note](https://github.com/NGEET/fates-docs) update: None needed
- [User's Guide](https://github.com/NGEET/fates-users-guide) update:  None needed

### Test Results:

*CTSM (or) E3SM (specify which) test hash-tag:* `ctsm5.1.dev171`

*CTSM (or) E3SM (specify which) baseline hash-tag:* `ctsm5.1.dev171`

*FATES baseline hash-tag:* `fates-sci.1.72.1_api.33.0.0`

*Test Output:*
izumi: all `fates` tests were PASS except for some `EXPECTED FAILS`
derecho: currently running `fates` test suite


fire/SFNesterovMod.F90
```diff
@@ -0,0 +1,95 @@
...
+      ! LOCALS:
+      real(r8) :: yipsolon ! intermediate variable for dewpoint calculation
+      real(r8) :: dewpoint ! dewpoint
+
+      if (precip > min_precip_thresh) then ! NI is 0.0 if it rains
+        calc_nesterov_index = 0.0_r8
+      else 
+        ! Calculate dewpoint temperature 
+        yipsolon = (SF_val_fdi_a*temp_C)/(SF_val_fdi_b + temp_C) + log(max(1.0_r8, rh)/100.0_r8) 
+        dewpoint = (SF_val_fdi_b*yipsolon)/(SF_val_fdi_a - yipsolon)
```
ckoven, line 84: This point is out of scope for this PR, but I just wanted to note that the parameters relevant to this approximation for dew point (which are [here](https://github.com/NGEET/fates/blob/main/parameter_files/fates_params_default.cdl#L1669) and [here](https://github.com/NGEET/fates/blob/main/parameter_files/fates_params_default.cdl#L1673)) seem more as approximations of thermodynamic constants than as tunable parameters, and also are very close to but not exactly [what wikipedia at least lists them as ](https://en.wikipedia.org/wiki/Dew_point#Calculating_the_dew_point). Given that, maybe it makes more sense to pull them out of the parameter file and move to [the physical constants file](https://github.com/NGEET/fates/blob/main/main/FatesConstantsMod.F90)?

CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+cmake_minimum_required(VERSION 3.4)
```
ckoven, line 1: Might be useful here to add some comments on what this file is for, i.e. that it is just for building unit tests and not the main model code itself.

ckoven (approved):
Thanks @adrifoster this looks really great. As you can see, I just went down a bit of an internet rabbit hole on dew point calculations; maybe we can revisit the question about moving some of those parameters out of the parameters file and into the physical constants file after this change is integrated?

rgknox:
@adrifoster and I talked a little about the idea of migrating this work to a new base folder called "unit_tests".  This will allow us to to clearly see what is part of this new unit testing framework, and which tests are still in the older deprecated framework in "functional_unit_tests". Any pushback or thoughts on this? (i'd be happy to make the changes to help familiarize myself with the code)

biogeochem/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+list(APPEND fates_sources
+  FatesLitterMod.F90
```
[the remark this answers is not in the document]
adrifoster: Sure can do

fire/SFFireWeatherMod.F90
```diff
@@ -0,0 +1,32 @@
...
+  end type fire_weather
+
+  abstract interface
+    subroutine initialize_fire_weather(this)
+      import :: fire_weather 
+      class(fire_weather), intent(inout) :: this
+    end subroutine initialize_fire_weather
+
+    subroutine update_fire_weather(this, temp_C, precip, rh, wind)
+      use FatesConstantsMod, only : r8 => fates_r8
```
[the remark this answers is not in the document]
adrifoster: no it's not needed, I can remove it.

fire/SFNesterovMod.F90
adrifoster: Sure sounds good, happy to do this.

CMakeLists.txt
adrifoster: Can do!

fire/SFFireWeatherMod.F90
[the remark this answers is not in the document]
adrifoster: Actually nevermind, because it is in an abstract interface it does need to be there.

fire/SFNesterovMod.F90
ckoven: I guess we could migrate the exact constant values that we have in the parameter file now, and it would still be bit for bit. I'm still not sure I understand the source of the exact values of the constants we have now, but since they only differ from what is in the Wikipedia (and the reference pointed to there, which is https://doi.org/10.1175/BAMS-86-2-225)  in the fourth or so digit, I think it should be fine to leave as-is. Or we can change the numbers slightly so that they have a clear provenance, but then it won't be bit for bit. Or first the bit-for-bit change and then the non-bit-for-bit change.

adrifoster: Okay yeah, I'm not sure what the provenance of them is either. The Thonicke paper approximates dewpoint temperature as Tmin-4, so we must have done this ourselves at some point.

adrifoster: Okay I pulled those values out of the parameter file (but left them at what they were - we can change later?)

I also refactored this subroutine a bit to specifically calculate dewpoint as it's own function.

main/EDTypesMod.F90
```diff
@@ -31,6 +31,8 @@ module EDTypesMod
   use EDParamsMod,           only : nclmax, nlevleaf, maxpft
   use FatesConstantsMod,     only : n_dbh_bins, n_dist_types
   use shr_log_mod,           only : errMsg => shr_log_errMsg
+  use SFFireWeatherMod,      only : fire_weather
+  use SFNesterovMod,         only : nesterov_index
```
glemieux, line 35: This can be removed since `fireWeather` is being defined using the base abstract class, correct?
```suggestion
```

fire/SFFireWeatherMod.F90
```diff
@@ -0,0 +1,32 @@
...
+  use FatesConstantsMod, only : r8 => fates_r8
+
+  implicit none
+  private 
+
+  type, abstract, public :: fire_weather
+    real(r8) :: fire_weather_index   ! fire weather index
+    contains 
+    procedure(initialize_fire_weather), public, deferred :: Init
+    procedure(update_fire_weather),     public, deferred :: Update
```
glemieux, line 12: I think I like this style of using `PascalCase` for the public facing call, but using `snake_case` for the type-specific hidden procedure.

main/EDTypesMod.F90
adrifoster: oh hmm maybe, I thought I tried this and it wouldn't compile but I will try again.

fire/SFFireWeatherMod.F90
adrifoster: yeah, I like having a "private" version and a "public" version?

glemieux (approved):
Part 2: unit tests

This looks great.  Thank you very much for getting in the cmake infrastructure including for this and future unit testing.  I made one new issue to remind us to make the top level CMakeLists.txt more agnostic to HLM directory structure in the future: #1196

glemieux:
Note that the testing directory layout will be refactored in #1186.

glemieux:
>Another option would be to make the fire weather index subroutine take in a fire parameter class, and then we could just set it up by hand in the unit test. This would be a pretty big change (i.e. making the fire parameter class), but is maybe a good idea anyway.

@adrifoster I think this would be worth a github discussion post to hash out the pros/cons and keep track of for a future update.

adrifoster:
> > Another option would be to make the fire weather index subroutine take in a fire parameter class, and then we could just set it up by hand in the unit test. This would be a pretty big change (i.e. making the fire parameter class), but is maybe a good idea anyway.
> 
> @adrifoster I think this would be worth a github discussion post to hash out the pros/cons and keep track of for a future update.

Yeah sure. though this specific change/issue is no longer relevant for this PR because these were the parameters we moved off of the parameter file.

glemieux:
Regression testing on derecho against `fates-sci.1.76.0_api.35.0.0-ctsm5.2.002` shows all expected tests passing b4b.

Location: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1171-fates`

PR #1173, opened 2024-03-15
Update which should enable the emission parameters to be properly read 

rosiealice:


rosiealice:
Did not mean to make this PR here!

Issue #1170, opened 2024-03-13
normalize crown area weighted height history diagnostic


rgknox:
The calculation of crown area weighted canopy height is not complete. It needs to be normalized by  total crown area of all cohorts participating in the site level sum.  Here is where we accumulate the mean height:

https://github.com/NGEET/fates/blob/sci.1.72.1_api.33.0.0/main/FatesHistoryInterfaceMod.F90#L3604-L3605

There needs to be a sum of crown area, and division through by that area, following the cohort/patch loop.

PR #1174, opened 2024-03-19
refactor of LAI-by-patch-age history (FATES_LAI_AP)

rgknox:
### Description:

This is a bug fix.  The FATES_LAI_AP diagnostic did not seem to be written correctly.  Also, the variables should match FATES_LAI, when summed over patch area weights.

FATES_LAI = sum( FATES_LAI_AP * FATES_PATCHAREA_AP)

Fixes: #1170 
Fixes: #1183 

### Collaborators:
@samsrabin

This should change both FATES_LAI and FATES_LAI_AP, and a new variable FATES_ELAI is now available.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -5014,6 +5014,8 @@ contains
                   if_notnew: if ( .not. ccohort%isnew ) then
 
                      ! scale up cohort fluxes to the site level
+                     ! these fluxes have conversions of [kg/plant/timestep] -> [kg/m2/s]
+                     
                      hio_npp_si(io_si) = hio_npp_si(io_si) + &
                           ccohort%npp_tstep * n_perm2 * dt_tstep_inv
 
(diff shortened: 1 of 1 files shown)


samsrabin:
Looks like this works for me!
screenshot_0211

rgknox:
I might also add the fix to #1170 into this, since it's also trivial

main/FatesHistoryInterfaceMod.F90
```diff
@@ -2922,6 +2927,10 @@ subroutine update_history_dyn1(this,nc,nsites,sites,bc_in)
             hio_lai_secondary_si(io_si) = 0._r8
          end if
 
+         ! Normalize crown-area weighted height
+         if(site_ca>nearzero)then
+            hio_ca_weighted_height_si(io_si) = hio_ca_weighted_height_si(io_si)/site_ca
+         end if
```
JessicaNeedham, line 2933: BA weighted height is set to zero if site_ba is not > nearzero. Seems like it would be a very small number anyway but do we want the same thing for CA weighted height? 

Also the comment on 2867 could be updated to say that ca_weighted_height and ba_weighted_height are normalized outside of the patch loop.

JessicaNeedham (approved):
Looks good to me. Thanks @rgknox and @samsrabin

main/FatesHistoryInterfaceMod.F90
rgknox: I left it out because we zero everything in an automated function anyway, we don't actually need to zero the ba weighted output.  But now that you bring the point up, its probably confusing to anyone reading the code why we do things differently. I think its better to have these outputs processed consistently to avoid confusion. I'll do a look through and see if I can make things more consistent in other places as well.

JessicaNeedham: Thanks, that makes sense.

rgknox:
tests ok on derecho, some diffs, and the diffs are related only to the changed output variables

glemieux (approved):

PR #1186, opened 2024-04-17
Allometry unit test

adrifoster:
Adds a "functional unit test" for the FATES allometry equations, using Fortran code and then a python wrapper script

### Description:
Adds some unit test framework for reading in a parameter file, and for using the allometry module
Python wrapper scripts build and execute the code, and then plot output.

### Collaborators:
lots of help from @billsacks along the way
@rgknox 

None - does not affect production code

### Checklist
*Contributor*

*Integrator*

### Documentation
- [User's Guide](https://github.com/NGEET/fates-users-guide) update: should do this, we need to have a way to document how to run this on a personal computer, though it should run easily on an ncar/doe machine (?)

### Test Results:

Shouldn't affect anything but will run tests to make sure when we are ready to merge.

*FATES baseline hash-tag:*

*Test Output:*


adrifoster:
Before this comes in I need to figure out how to get the netcdf and netcdff locations without hard-coding them.

@billsacks said this should be possible with some CIME functions...?

adrifoster:
> Before this comes in I need to figure out how to get the netcdf and netcdff locations without hard-coding them.
> 
> @billsacks said this should be possible with some CIME functions...?

okay nevermind it wasn't that hard!

billsacks:
Awesome work on this @adrifoster ! 🎉

rgknox:
@adrifoster and I talked a little about the idea of migrating this work to a new base folder called "unit_tests".  This will allow us to to clearly see what is part of this new unit testing framework, and which tests are still in the older deprecated framework in "functional_unit_tests". Any pushback or thoughts on this? (i'd be happy to make the changes to help familiarize myself with the code)

ekluzek (approved):
I really like this and want to learn more about it. From my overview the main thing I wonder about is the name. I'm thinking of Integration rather than functional unit testing. I've long realized this is a  level of testing that is missing from our standard testing.

A lot of this (maybe the majority of it)  is starting a framework that can allow similar testing to other smaller parts of FATES for similar purposes. We've been using a similar CMake build system for unit-tests in CTSM and I think that makes sense here as well.

I like the idea of providing python plotting capability here. As a scientific development tool, the visualization of behavior for a scientific process is such an important tool. 

See this discussion I opened in CTSM: https://github.com/ESCOMP/CTSM/discussions/2508

I see the need for bringing something like into CTSM as well.

ekluzek:
@rgknox I'm curious what your directory naming suggestion would look like? I think you are saying to rename this to "unit_tests" which I think is a bad name -- because I think this is more on the integration level testing and hence distinct from "unit tests". So I think there should be a distinction between the two. But, would like to hear more about what you mean here...

rgknox:
Everything in that "unit_tests" directory, is intended to test "units" of the code.

ekluzek:
@rgknox and @adrifoster thanks for the discussion. @rgknox for "integration testing" I'm going for a specific meaning of the term in Software Engineering. This definition from the UCAR SEA "Integration testing builds upon unit testing by examining interactions between larger-scale software components, packages, and modules rather than individual functions and methods". You are right we also can use the term "Integration testing" to be the set of different testing that needs to happen to integrate changes into the code. In that case it usually means the different test suites that are run to make a tag (which will include different categories: unit and system tests with now the addition of functional).

After thinking about it I would categorize @adrifoster new additional as functional, because it's about exploring the scientific behavior of allometry. The fact that it has plotting, rather than pass/fail also tips it toward that direction. In this case it also happens to just explore a single unit of code, which is why @adrifoster is using both functional and unit in the name.

See this discussion...

https://github.com/ESCOMP/CTSM/discussions/2508

adrifoster:
Okay I updated this so now we can run the unit tests here too, though right now we can't say run "functional" vs. "unit" tests. That can come in now or in a future PR.

rgknox (approved):
This PR look pretty tight, I can't see much to pick at.  Its big, yes, so I'm sure that I missed things. But the style is great, good comments, we already dialed in on naming conventions, and we already debated the communication method (netcdf io) between python plotting and the fortran test programs.

glemieux:
Regression testing against `fates-sci.1.76.1_api.35.1.0-ctsm5.2.004` shows all expected tests passing b4b, as expected. Note that this was tested in tandem with #1191 

results: /glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1191-1186-fates

PR #1197, opened 2024-05-10
Move growth respiration to daily timestep

### LINKED ISSUE #1194: issue with growth respiration calculation
Last week, @JessicaNeedham and I were discussing growth respiration, and we noticed that the growth respiration calculation may not be doing exactly what is intended. the key line is here:

https://github.com/NGEET/fates/blob/main/biogeophys/FatesPlantRespPhotosynthMod.F90#L998-L999

The issue is the `max(0._r8, gpp-mr)` part. while it makes sense to not let growth respiration go negative at night, adding the zero floor to this means that, integrated over a whole day/night cycle, GR !=  grperc*(GPP-MR), it only equals that over the part of the day when GPP > MR. And this high frequency growth respiration gets integrated through the `npp_tstep` term over the day to result in the total carbon available to grow with. So as a result, I think we are likely overestimating the growth respiration and thus underestimating the NPP relative to what we intend for a given value of the `grperc` parameter. Unless the `grperc` parameter is only intended to be applied during the day, but I don't believe that is the case. So if the intention is that the grperc parameter is meant to be applied over both day and night, we should reformulate this expression (somehow, not sure exactly best how to, that can be a later part of the discussion).

### LINKED ISSUE #1229: Plant "excess respiration" not removed from NEP/NPP in FATES
Hi team, 

In our simulations with FATES at FACE experiments and transient runs at Lambir we've found that the output variables **FATES_NEP and FATES_NPP do not include plant excess respiration**. Presumably this also affects other NPP output variables.  

Excess respiration is respiration in FATES-CNP when allocatable carbon is available but nutrients limit any further growth and carbon storage pools are at their maximum, [calculated here](https://github.com/NGEET/fates/blob/1982b0032c3cab6278892eccb85f643114ffb1af/parteh/PRTAllometricCNPMod.F90#L1918-L1934).  

The FATES_NPP and the plant side of FATES_NEP [are calculated here](https://github.com/NGEET/fates/blob/1982b0032c3cab6278892eccb85f643114ffb1af/main/FatesHistoryInterfaceMod.F90#L5071-L5076) as a function of `ccohort%npp_tstep`.

`ccohort%npp_tstep` is [calculated here](https://github.com/NGEET/fates/blob/1982b0032c3cab6278892eccb85f643114ffb1af/biogeophys/FatesPlantRespPhotosynthMod.F90#L1002-L1005) in FatesPlantRespPhotosynthMod.F90. Pretty clear that this is just taking maintenance and growth respiration into account. 

**Flagging here for thoughts and what to watch out for in any fix.** It seems like it could just be done in the history variable processing but excess resp would also need to be calculated with the same indexing as all the various NPP variables. 

**Are there other things that rely on the `ccohort%npp_tstep` variable in their calculation? Might it be better to update that in the parteh subroutines when they execute?** If yes, presumably this woudl affect the C balance error checking on the ELM side, and maybe FATES too.     

Interestingly, both for FATES and Lambir simulations the NPP values **without** excess respiration removed give a better fit to some of the data. Probably for different reasons. But that's a conversation perhaps for a different thread.  

Note: NEE on the ELM side does have excess respiration included. Not checked NPP.
Also note: the C balance checks are fine, this is from running versions of the model with no errors.

ckoven:
This PR moves the calculation of growth respiration (GR) to the daily timestep, as a tax on daily integrated `max(0,(GPP - MR))`, instead of half-hourly `max(0,(GPP-MR))`. This will lead to a decrease in growth respiration because it includes the nighttime respiration in the loss term.

### Description:
By moving the GR to daily, it means that NPP is also only defined at the daily timestep as well. With one exception, I moved all history variables that depend on growth respiration to also be daily.  The one exception was I kept NEP as a high frequency output for the case where a user wants to compare directly to NEE data; however this requires (1) making the assumption that GR is constant on timescales less than a day, and (2) using a given cohort's GR from the prior day (which thus introduces some slight error since cohort _n_ will have changed since the GR was actually applied).

fixes #1194
potentially:
fixes #1229

### Collaborators:
Discussed on 4/29 FATES SW call, and further discussions with @JessicaNeedham and @rgknox.

This will be answer-changing.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


biogeochem/FatesCohortMod.F90
```diff
@@ -142,13 +142,15 @@ module FatesCohortMod
...
     real(r8) :: npp_acc_hold
 
-    real(r8) :: resp_tstep                ! Autotrophic respiration (see above *)
-    real(r8) :: resp_acc
-    real(r8) :: resp_acc_hold
+    real(r8) :: resp_m_tstep              ! Maintenance respiration (see above *)
+    real(r8) :: resp_m_acc
+    real(r8) :: resp_m_acc_hold
+
+    real(r8) :: resp_g_acc                ! Growth respication can only be calculated at the daily timestep
```
rgknox, line 152: @ckoven , do we need resp_g_acc?  The <var>_acc variable was created to track the accumulation of the flux through the fast timesteps over the course of the day. Since we don't accumulate growth respiration anymore, since it is defined only at the end of the day, I believe we only need resp_g_acc_hold, right?

biogeophys/EDAccumulateFluxesMod.F90
```diff
@@ -82,15 +82,13 @@ subroutine AccumulateFluxes_ED(nsites, sites, bc_in, bc_out, dt_time)
 
                    if ( debug ) then
```
rgknox, line 83: we should remove this debug statement.  we can add these types of write statements in when needed, but not maintained in the main codebase

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -987,23 +987,15 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
 
                         ! convert from kgC/indiv/s to kgC/indiv/timestep
                         currentCohort%resp_m        = currentCohort%resp_m  * dtime
+                        currentCohort%resp_m_tstep  = currentCohort%resp_m   ! these two things are the same. we should get rid of one them
```
rgknox, line 990: agreed, I vote we get rid of resp_m  since resp_m_tstep is more descriptive

main/FatesHistoryInterfaceMod.F90
```diff
@@ -2814,6 +2815,15 @@ subroutine update_history_dyn1(this,nc,nsites,sites,bc_in)
                ! have any meaning, otherwise they are just inialization values
                notnew: if( .not.(ccohort%isnew) ) then
```
rgknox, line 2817: As part of an upcoming refactor, we will be adding calls to daily fluxes that occur before we change plant number densities.  As already acknowledged, the number of plants present in this point of the call sequence is not the same number of plants that generated the growth flux.  This is a known issue, but it outside the scope of this PR.

biogeochem/FatesCohortMod.F90
ckoven: right, this was a thing I wasn't sure about. I was guessing that it is as you describe, but wanted to include the `_acc` variable to be on the safe side. If you don't think it is needed, then happy to remove.

main/FatesHistoryInterfaceMod.F90
rgknox: Here is the last place in the dynamics call sequence where number density has not changed yet: https://github.com/NGEET/fates/blob/sci.1.76.3_api.35.1.0/main/EDMainMod.F90#L751

Here is where we make a call to write daily nutrient fluxes, which also happens before we change number density: https://github.com/NGEET/fates/blob/sci.1.76.3_api.35.1.0/main/EDMainMod.F90#L698

rgknox:
This may or may not fit into this PR @ckoven, but I did want to point out that we have another carbon flux that is triggered daily.  When the nutrient model is active, if carbon is not the limiting resource, any carbon that was available for growth but was not allocated to tissues and storage will be respired off.  The variable is cohort%resp_excess, and it is resolved at the end of the calls to growth, here:

https://github.com/NGEET/fates/blob/main/main/EDMainMod.F90#L595

Ideally, the plant's dynamic root allocation schemes should nudge the plant's resource allocation balance so that it is not wasting resources, but nonetheless this is a term that should be accounted for if anyone is tracking NEE or other carbon flux balancing while the nutrient model is active.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -2377,6 +2375,9 @@ subroutine update_history_dyn1(this,nc,nsites,sites,bc_in)
          hio_sum_fuel_si         => this%hvars(ih_sum_fuel_si)%r81d,  &
          hio_litter_in_si        => this%hvars(ih_litter_in_si)%r81d, &
          hio_litter_out_si       => this%hvars(ih_litter_out_si)%r81d, &
+         hio_npp_si              => this%hvars(ih_npp_si)%r82d, &
```
JessicaNeedham, line 2378: should be r81d

main/EDMainMod.F90
JessicaNeedham, line 1: There's a resp_acc on line 675 that needs to be updated

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -735,7 +735,7 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
 
                            ! Zero cohort flux accumulators.
```
JessicaNeedham, line 736: This npp_tstep needs to be deleted now.

JessicaNeedham (commented):
Thanks @ckoven ! This mostly looks good to me. It compiled after I fixed a few very minor things.

JessicaNeedham:
Results from a global nocomp simulation with the default parameter file. Growth respiration decreased by about 10% with this change. This is just the mean over a 10 year simulation from bare ground so definitely not spun  up yet. 


Screenshot 2024-06-28 at 19 58 11
Screenshot 2024-06-28 at 20 01 12

rgknox:
@JessicaNeedham , that 10% decrease seems about what I would had expected, do you see it the same way?

JessicaNeedham:
> @JessicaNeedham , that 10% decrease seems about what I would had expected, do you see it the same way?

I think so

main/EDMainMod.F90
```diff
@@ -514,14 +515,30 @@ subroutine ed_integrate_state_variables(currentSite, bc_in, bc_out )
...
              ! -----------------------------------------------------------------------------
 
-             currentCohort%npp_acc_hold  = currentCohort%npp_acc  * real(hlm_days_per_year,r8)
              currentCohort%gpp_acc_hold  = currentCohort%gpp_acc  * real(hlm_days_per_year,r8)
-             currentCohort%resp_acc_hold = currentCohort%resp_acc * real(hlm_days_per_year,r8)
+             currentCohort%resp_m_acc_hold = currentCohort%resp_m_acc * real(hlm_days_per_year,r8)
+
+             ! at this point we have the info we need to calculate growth respiration
+             ! as a "tax" on the difference between daily GPP and daily maintenance respiration
+             if (hlm_use_ed_prescribed_phys .eq. itrue) then
```
JessicaNeedham, line 523:
```suggestion
             if (hlm_use_ed_prescribed_phys .eq. ifalse) then
```

main/FatesHistoryInterfaceMod.F90
JessicaNeedham, line 2378:
```suggestion
         hio_npp_si              => this%hvars(ih_npp_si)%r81d, &
```

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -735,7 +735,7 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
...
                            currentCohort%npp_tstep  = 0.0_r8
```
JessicaNeedham, line 737:
```suggestion
                           
```

JessicaNeedham (approved):

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -987,23 +987,15 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
```
rgknox, line 987: Noting that resp_m_unreduced is in units of per second here.  This is only used as a history diagnostic.  In the history interface, it is assuming that the units here have been multiplied by dtime (ie that it is per timestep).  Will correct.

rgknox (approved):

rgknox:
I believe this PR is content complete and has had a thorough examination.

glemieux:
Initial regression testing underway on derecho.

glemieux:
Testing on derecho against `fates-sci.1.79.2_api.36.1.0-ctsm5.3.008` shows no unexpected test failures.  As expected all fates test mods result in DIFFs.

Spot checking the `FatesColdAllVars` differences shows what I believe is expected behavior with the following variables being the only differences in the first fates dynamics timestep (i.e. `t_index = 2`):
```
FATES_AUTORESP
FATES_AUTORESP_CANOPY
FATES_GROWTH_RESP
FATES_MAINT_RESP_UNREDUCED
FATES_NEP
FATES_NPP
FATES_NPP_AP
FATES_GROWAR_CANOPY_SZ
FATES_AUTORESP_SZPF
FATES_GROWAR_SZPF
```
Beyond this timestep, more differences show up in additional variables.  The differences for the variables above are on the order of `E-09`.

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1197`

glemieux:
Checking the `FatesColdSatPhen` tests, the only diffs I'm seeing are with `FATES_MAINT_RESP_UNREDUCED` which are around `E-08`.

PR #1206, opened 2024-05-22
Implement grass allometric equations and update default allometric parameters for grass PFT

XiulinGao:
### Description:
This pull request implements new leaf and AGB allometric equations using basal diameter (cm) and plant height (m) as predictors for grass PFT. In addiiton, a new sapwood option was added in which I set bsap = bagw + bbgw - bdead and bdead is always zero by setting agb_frac to 1. This change will skip the bsap_ltarg_slatop routine for sapwood biomass calculation for grass, and basically make bsap = bagw and there won't be any dead biomass produced for grasses if agb_frac is set to be 1. Lastly, all default allometric parameters, including reproduction allocation, are updated for all three grass PFTs, based on data from Gao et al., 2024. (https://esajournals.onlinelibrary.wiley.com/doi/pdfdirect/10.1002/eap.2976)

### Collaborators:
I have talked with @ckoven, @glemieux , and @mpaiao about implementing this new grass allometry and had discussions with them. 

This change will result in different model results for grass PFTs in terms of carbon allocation and the resulting impacts on PFT competitions. 
### Checklist

*Contributor*
- [ x ] The in-code documentation has been updated with descriptive comments

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


biogeochem/FatesAllometryMod.F90
```diff
@@ -541,6 +547,11 @@ subroutine carea_allom(dbh,nplant,site_spread,ipft,crowndamage,c_area,inverse)
           call carea_3pwr(dbh_eff,height,ipft,dbh_maxh, site_spread,d2bl_p2, &
                d2bl_ediff, d2ca_min,d2ca_max,crowndamage, c_area, do_inverse)
           capped_allom = .true.
+       case(5)
+          dbh_eff = min(dbh,dbh_maxh)
+          call carea_2pwr(dbh_eff,site_spread,d2bl_p2,d2bl_ediff,d2ca_min,d2ca_max, &
+               crowndamage, c_area, do_inverse)
+          capped_allom = .true.
```
mpaiao, line 554: I suggest removing `case (5)` and replacing `case (3)` with `case(3,5)`. This way future changes to `care_2pwr` don't need to be applied twice.

mpaiao (commented):
@XiulinGao Thanks for submitting this pull request! I went through it and I think it looks great! I just have a couple of comments that are probably something for us to think in future pull requests:

1. Many of these modes (especially for crown area and leaf/AG biomass) really could be simplified into 1 generic allometric mode. This may make the code a bit simpler overall.
2. All the derivatives are fine, but in most cases, it could be rewritten in terms of the biomass itself, which saves come floating point operations. I've done this for a few allometric modes. I don't think it's high priority because it's not something computed multiple times a day, but it could save a few milliseconds ;).

biogeochem/FatesAllometryMod.F90
XiulinGao: Thanks Marcos! I have made changes in response to you comments on derivative calculation and the crown area function. But I haven't changed the agb and bleaf allometry to generic function yet. Hope now the code looks less redundant.

mpaiao: @XiulinGao Thanks for the follow up. My comment about the generic function was something for a future pull request, not yours. This would likely require significant changes in the parameter files for most users, so it's probably something we want to discuss with the broader FATES community before implementing.

mpaiao (approved):

biogeochem/FatesAllometryMod.F90
```diff
@@ -530,7 +536,7 @@ subroutine carea_allom(dbh,nplant,site_spread,ipft,crowndamage,c_area,inverse)
           call carea_2pwr(dbh,site_spread,d2bl_p2,d2bl_ediff,d2ca_min,d2ca_max, & 
                crowndamage, c_area, do_inverse)
           capped_allom = .false.
-       case(3)
+       case(3,5)
```
rgknox, line 539: @XiulinGao , is this case 5 intentional? I don't see any discussion or comments related to a new case 5 leaf allometry.

rgknox:
@XiulinGao : "addiiton, a new sapwood option was added in which I set bsap = bagw + bbgw - bdead and bdead is always zero by setting agb_frac to 1."

My understanding of the code is that bdead is set to zero because allom_smode = 2.  Is that correct?

This is a fantastic addition to the code, thanks for submitting this work.  Looking forward to getting this in.  Originally, I had one request, that we add more description to these new allometry functions in the parameter file, and/or the module that reads in the parameters. However, we don't really have a precedent for that, so we will have to get back to this.

biogeochem/FatesAllometryMod.F90
```diff
@@ -1036,6 +1042,35 @@ subroutine bsap_allom(d,ipft,crowndamage,canopy_trim,elongf_stem, sapw_area,bsap
...
        end if
 
+    case(2) ! this is a 'sapwood' function specifically for grass PFT that do not produce
+            ! dead woody biomass. So bsap = bagw. Might remove the bsap and bdead for grass
+            ! in the future as there is no need to distinguish the two for grass above- and belowground biomass
+
+       call bagw_allom(d,ipft,crowndamage,elongf_stem,bagw,dbagwdd)
+       call bbgw_allom(d,ipft, elongf_stem,bbgw,dbbgwdd)
+       bsap = bagw + bbgw
+
```
rgknox, line 1052: if this is replicating code, we should encapsulate this in a routine and call it in the different locations

biogeochem/FatesAllometryMod.F90
```diff
@@ -1036,6 +1042,35 @@ subroutine bsap_allom(d,ipft,crowndamage,canopy_trim,elongf_stem, sapw_area,bsap
...
+       ! replicate the crown damage code
+       ! Do we really need this for grass? I would think this can be helpful for
+       ! grazing in the future. --XLG
+       if(crowndamage > 1)then
```
rgknox, line 1056: tagging @JessicaNeedham

JessicaNeedham: Agree @XiulinGao that we should think ahead about possible grazing functionality. But as written the crown damage module doesn't change plant height, only crown area and the biomass of woody tissues,  so I'm not sure how applicable it is to grasses. Maybe we could skip this for now, but think about making the damage code more general so it could be used in the future?

glemieux: Per discussion in the fates software meeting today @rgknox will make a PR to this branch to remove this and include a graceful failure.

XiulinGao: I see. In that case then I think this is not that helpful for grass grazing. I agree that we can remove this for grasses for now.

XiulinGao:
> @XiulinGao : "addiiton, a new sapwood option was added in which I set bsap = bagw + bbgw - bdead and bdead is always zero by setting agb_frac to 1."
> 
> My understanding of the code is that bdead is set to zero because allom_smode = 2. Is that correct?
> 
> This is a fantastic addition to the code, thanks for submitting this work. Looking forward to getting this in. Originally, I had one request, that we add more description to these new allometry functions in the parameter file, and/or the module that reads in the parameters. However, we don't really have a precedent for that, so we will have to get back to this.

Yes. The new sapwood function just calculate bsap as the sum of above- and belowground live woody biomass for grasses. But as we set agb_frac to 1, so bsap = agb. I get the logic backwards in documentation here... it's bdead that calculated as bagw + bbgw - bsap not the other way around.....

rgknox:
ok, thanks for confirming my understanding @XiulinGao . There are some small tweaks I'll submit to your fork as a PR, but we should be able to integrate this PR fairly soon.

parameter_files/fates_params_default.cdl
```diff
@@ -983,55 +983,55 @@ data:
...
 
  fates_allom_d2ca_coefficient_max = 0.2715891, 0.3693718, 1.0787259, 
     0.0579297, 0.2715891, 1.1553612, 0.6568464, 0.6568464, 0.6568464, 
-    0.6568464, 0.6568464, 0.6568464 ;
+    0.0408, 0.0408, 0.0862 ;
 
  fates_allom_d2ca_coefficient_min = 0.2715891, 0.3693718, 1.0787259, 
     0.0579297, 0.2715891, 1.1553612, 0.6568464, 0.6568464, 0.6568464, 
-    0.6568464, 0.6568464, 0.6568464 ;
+    0.6568464, 0.0408, 0.0862 ;
```
rgknox, line 1022: @XiulinGao can you confirm that fates_allom_d2ca_coefficient_min for PFT 10 should **not** be changed?  I see that PFT 11 and 12 are changed, and PFT 10 is changed for the max term.

rgknox (approved):

PR #1215, opened 2024-06-18
Fire weather refactor (non-b4b)

adrifoster:
This provides a second refactor to the SPITFIRE set of modules, adding effective windspeed and methods to calculate it to the fire weather class.

### Description:
 I made effective wind speed (previously a patch-level variable that was set to be the same value across every patch in the same site), an attribute of the (currently site-level) fire weather class.

Future updates can make the fire weather class a patch attribute if needed. 

Tests will NOT be b4b because we average the patch vs. site variables differently in the history output, but *only* the effective wind speed output is (very slightly) different. No other history variables are affected.

### Checklist
*Contributor*

*Integrator*

### Test Results:

Have run a small test case in the Amazon to prove only wind speed was affected. Will run full test suite as well.

*CTSM (or) E3SM (specify which) test hash-tag:* `ctsm5.2.007`

*CTSM (or) E3SM (specify which) baseline hash-tag:* `ctsm5.2.007`

*FATES baseline hash-tag:* `sci.1.76.4_api.35.1.0`

*Test Output:*

DIFFS for only `FATES_EFFECT_WSPEED`, and were minimal

```
 FATES_EFFECT_WSPEED   (lon,lat,time)  t_index =      4     4
       4148    13824  (    82,    78,     1) (   125,    46,     1) (     3,    81,     1) (     3,    81,     1)
                4148   6.762390095557969E+00   1.844060332053566E-01 5.4E-03  5.188808499425933E+00 1.7E-04  5.188808499425933E+00
                4148   6.760903248381456E+00   1.844019126634444E-01          5.183368180908529E+00          5.183368180908529E+00
               13824  (    82,    78,     1) (   125,    46,     1)
          avg abs field values:    1.574091338575881E+00    rms diff: 4.3E-04   avg rel diff(npos):  1.7E-04
                                   1.573816055109091E+00                        avg decimal digits(ndif):  4.0 worst:  3.0
 RMS FATES_EFFECT_WSPEED              4.2606E-04            NORMALIZED  2.7069E-04
```


biogeochem/FatesPatchMod.F90
```diff
@@ -613,6 +617,42 @@ end subroutine Create
 
     !===========================================================================
 
+    subroutine UpdateTreeGrassArea(this)
```
mpaiao, line 620: Thanks for turning this bit into a subroutine!

fire/SFMainMod.F90
```diff
@@ -113,33 +106,44 @@ end subroutine fire_model
...
+    !  Currently we use tree and grass fraction averaged over whole grid (site) to 
+    !  prevent extreme divergence
 
     use FatesConstantsMod, only : tfrz => t_water_freeze_k_1atm
-    use FatesConstantsMod, only : sec_per_day
+    use FatesConstantsMod, only : sec_per_day, sec_per_min
+    use EDTypesMod,        only : CalculateTreeGrassArea
+
+    ! CONSTANTS:
+    real(r8), parameter :: wind_atten_treed = 0.4_r8 ! wind attenuation factor for tree fraction
```
mpaiao, line 119: I see that these attenuation factors are defined as local constants both in here and also at subroutine `UpdateEffectiveWindSpeed` in SFFireWeatherMod. Perhaps turn these into global parameters, just to ensure they are always consistent across subroutines?

fire/SFMainMod.F90
```diff
@@ -113,33 +106,44 @@ end subroutine fire_model
...
     ! is simply using the values associated with the first patch.
-    ! which probably won't have much inpact, unless we decide to ever calculated fire weather for each patch.  
+    ! which probably won't have much impact, unless we decide to ever calculated fire weather for each patch.  
     
     currentPatch => currentSite%oldest_patch
 
     ! If the oldest patch is a bareground patch (i.e. nocomp mode is on) use the first vegetated patch
     ! for the iofp index (i.e. the next younger patch)
-    if(currentPatch%nocomp_pft_label .eq. nocomp_bareground)then
+    if (currentPatch%nocomp_pft_label == nocomp_bareground) then
```
mpaiao, line 146: I like `==` better than `.eq.` too ;)

main/EDTypesMod.F90
```diff
@@ -518,6 +519,43 @@ end subroutine ZeroMassBalFlux
    
   ! =====================================================================================
 
+  subroutine CalculateTreeGrassArea(csite, tree_fraction, grass_fraction, bare_fraction)
```
mpaiao, line 522: It took me a while to see how this differs from `UpdateTreeGrassArea`. One suggestion may be to rename this to `UpdateTreeGrassAreaSite` and the other routine to `UpdateTreeGrassAreaPatch`, just so the difference is clearer from their names (although longer names).

mpaiao (commented):
@adrifoster Many thanks for refactoring this! I went through the changes and they look good, and I liked how you turned many parts into separate subroutines. I just made some minor suggestions on a couple of parameters and subroutines for clarity.

fire/SFMainMod.F90
adrifoster: I think this was left in by mistake. I am going to delete them here because they are not used anywhere else.

main/EDTypesMod.F90
adrifoster: As discussed in the FATES meeting, because the patch method will always be called with patch%UpdateTreeGrassArea I am going to leave that one and only change the site one as you suggested.

adrifoster:
Thanks for the review @mpaiao. I made the requested changes and am running regression testing now.

mpaiao (approved):

adrifoster:
Testing seems to be the same as before - non-b4b but only for `FATES_EFFECT_WSPEED`, e.g.:

```
grep "RMS" /glade/derecho/scratch/afoster/tests_0823-151723de/ERS_D_Ld3.f19_g17.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesCold.GC.0823-151723de_int/run/ERS_D_Ld3.f19_g17.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesCold.GC.0823-151723de_int.clm2.h0.2000-01-01-00000.nc.cprnc.out
 RMS FATES_EFFECT_WSPEED              3.9756E-04            NORMALIZED  2.5913E-04
 RMS FATES_EFFECT_WSPEED              4.2606E-04            NORMALIZED  2.7069E-04

```

PR #1226, opened 2024-07-18
cleaning up nocomp patch indexing

rgknox:
### Description:

This set of changes removes the need for all these checks that determine if a patch is nocomp+fixedbiogeog.  We currently do this check because when in this mode, we construct a bareground patch in the zero-th patch index position.  When we traverse patch loops, sometimes we increment the patch index, and this logic check makes sure that we don't increment.  

However, we don't need to put this special bare-ground patch in the zeroth position.  We can just put it in the first position, and allow it to be in-line index-wise with the what ELM/CLM identifies as the 1st vegetated patch.  This is fine, FATES has patches in the full dynamics mode that have no cohorts, it's used to having patches with no vegetation lined-up with patches in CLM/ELM that are "vegetated".  What happens is we set the weighting to zero, and many of the fluxes are some nominal value.

We also don't need to touch the ELM/ELM side of the code, because the way we create out weightings actaully works for both the indexing schemes described. Why? because we assign the CLM/ELM bare-ground patch fraction based on 1 minus the sum of all patches total canopy area, which doesn't care about index position.

Here are some slides that give a sense of what is going on and what is proposed: https://docs.google.com/presentation/d/1ZwErNVF3B-_jJwpCDud3d-MtR59Idax8YTWe_ErEc8Q/edit?usp=sharing

Further, if someone want's to bypass some subset of physics, or some algorithm, because there is no vegetation, it would be more appropriate anyway to compare against total canopy area, and not if it is a nocomp+bareground patch.

I still need to make sure that we don't exceed our patch counts on the fates-side when in nocomp+fixedbiogeog, in this mode we were using 1 less space as we matched with ELM/CLM arrays.

### Collaborators:

Discussions with @glemieux 

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


glemieux:
> We also don't need to touch the ELM/ELM side of the code, because the way we create out weightings actaully works for both the indexing schemes described.

For the hlm-fates interface mod, we skip the `col%patchi(c)` indexing by mapping fates patch index `ifp` onto hlm patch index `p` via `p = ifp+col%patchi(c)`.  If we make `ifp=1` the bareground patch for fates, wouldn't we want to remove this mapping on the hlm side API?

glemieux:
@rgknox and @glemieux to talk about this this coming Thursday.

main/FatesInterfaceMod.F90
```diff
@@ -808,19 +813,27 @@ subroutine SetFatesGlobalElements1(use_fates,surf_numpft,surf_numcft,param_reade
...
+            if(hlm_use_luh==ifalse) then
+               maxpatches_by_landuse(secondaryland:n_landuse_cats) = 0
+            end if
             
             if(hlm_use_nocomp==itrue) then
-
                maxpatches_by_landuse(primaryland) = max(maxpatches_by_landuse(primaryland),fates_numpft)
-               maxpatch_total = sum(maxpatches_by_landuse(:))
+               !! REVIEWERS: NEED THIS NEXT LINE?
+               !maxpatches_by_landuse(secondaryland) = max(maxpatches_by_landuse(secondaryland),fates_numpft)
```
rgknox, line 824: @ckoven and others..  Don't we need to allocate at least numpft patches for the secondary lu category (and others if nocomp is on?

rgknox: and actually, this shouldn't be a maximum here, it should not be more than fates_numpft, it should be equal to

ckoven: @rgknox there is logic in the nocomp x land use initialization that limits the number of PFTs allowed on secondary (or other LU) patches to be the number of secondary patches allowed, see here: https://github.com/NGEET/fates/blob/main/main/EDInitMod.F90#L554-L578

rgknox: I see, thanks for pointing that out.  Can you confirm that you do want to make sure maxpatches_by_landuse(primary) is greater than or equal to numpft, at this location in the code?  The logic in the above mentioned line did not seem to ignore primary.

rgknox: Discussed this offline with @ckoven, questions addressed

rgknox:
FATES test suite is generating some unexpected fails and diffs with base.

The coldbasic satphen tests show differences in 'PCT_NAT_PFT'
Another satphen tests has many diffs in many variables..
There appears to be a problem building or executing two of the luh enabled tests
One of the fixedbiogeog tests had diffs only in  FATES_PATCHAREA_LU and FATES_DISTURBANCE_RATE_MATRIX_LULU

I think some of this has to do with fire and how we decide which patch to use for input boundary conditions on the site scale?

Either way, plenty to fix here.

```
./cs.status.fails
1210-094439de_gnu: 10 tests
    FAIL ERS_D_Ld15.f45_f45_mg37.I2000Clm50FatesRs.derecho_gnu.clm-FatesColdTwoStreamNoCompFixedBioGeo COMPARE_base_rest (EXPECTED FAILURE)
    FAIL ERS_D_Ld15.f45_f45_mg37.I2000Clm50FatesRs.derecho_gnu.clm-FatesColdTwoStreamNoCompFixedBioGeo BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF
    FAIL PEM_D_Ld15.5x5_amazon.I2000Clm50FatesRs.derecho_gnu.clm-FatesColdSeedDisp COMPARE_base_modpes (EXPECTED FAILURE)
    FAIL SMS_D_Ld3.f09_g17.I2000Clm60FatesSpCruRsGs.derecho_gnu.clm-FatesColdSatPhen_prescribed RUN time=94 (EXPECTED FAILURE)
    PASS SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Fates.derecho_gnu.clm-FatesPRISM--clm-NEON-FATES-YELL SHAREDLIB_BUILD time=162 (UNEXPECTED: expected FAIL)
    PASS SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Fates.derecho_gnu.clm-FatesPRISM--clm-NEON-FATES-YELL RUN time=103 (UNEXPECTED: expected FAIL)

 
1210-094439de_int: 38 tests
   --->  FAIL ERP_P128x2_Ld30.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdSatPhen BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF
    FAIL ERP_P256x2_Ld30.f45_f45_mg37.I2000Clm60FatesRs.derecho_intel.clm-mimicsFatesCold RUN time=49 (EXPECTED FAILURE)
    PEND ERP_P256x2_Ld30.f45_f45_mg37.I2000Clm60FatesRs.derecho_intel.clm-mimicsFatesCold COMPARE_base_rest
    FAIL ERS_D_Ld15.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesColdTwoStream COMPARE_base_rest (EXPECTED FAILURE)
    --> PEND ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLandUse SHAREDLIB_BUILD (UNEXPECTED: expected FAIL)
    --> FAIL ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLUH2 BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF
    --> FAIL ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLUH2HarvestArea BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF
    --> FAIL ERS_D_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdLUH2HarvestMass BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF
    --> FAIL ERS_Ld30.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdNoCompFixedBioGeo BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF
    FAIL ERS_Ld30.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdSatPhen BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF
    FAIL ERS_Ld60.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdST3 RUN time=26 (EXPECTED FAILURE)
    PEND PVT_Lm3.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesLUPFT RUN
    PASS SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Fates.derecho_intel.clm-FatesFireLightningPopDens--clm-NEON-FATES-NIWO SHAREDLIB_BUILD time=137 (UNEXPECTED: expected FAIL)
    FAIL SMS_Lm1.f45_f45_mg37.I2000Clm60FatesSpCruRsGs.derecho_intel.clm-FatesColdBasic BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF

 
1210-094439de_nvh: 1 test
    FAIL SMS.f45_f45_mg37.I2000Clm60FatesSpRsGs.derecho_nvhpc.clm-FatesColdSatPhen BASELINE fates-sci.1.80.4_api.37.0.0-ctsm5.3.012: DIFF
```

rgknox:
I'm realizing there is a subtle flaw in the approach to this PR. When in nocomp and fixed-biogeo modes (including and excluding SP mode), there is an assumption that the patch index is equivalent to the PFT index of that patch. While the most of logic in the changes hold, this discontinuity affects output such as PCT_NAT_VEG, and other variables that have this assumption.  
I'm going to circle back and take on a slightly different approach. We will continue to allow a zero index patch, but we will also have a global integer that states if the first index is a zero or a one.  This will allow us to get rid of  these "if  patchno==nocomp_bareground" filters.

rgknox:
As per the previous comment (Dec 28th 2024), this PR has been re-structured.  The idea of shifting indexes and abandoning the zero patch created too many issues.   The current set of changes seeks only to make our methods more consistent.  

1) Instead of iterating a patch number index while looping patches, we now always use the patch%patchno to get a boundary condition index for our bc_in and bc_out arrays.
2) We no longer filter nocomp patches based on patchno == 0, and instead compared if the patch's nocomp_pft_label equals nocomp_bareground.
3) There was an inconsistency in how we calculated vegetation temperatures. Originally I thought this inconsistency was causing issues related to nocomp bareground patches, where those patches had uninitialized temperatures.  After investigation, those patches appear to have vegetation temperatures initialized from air temperature. So these changes are not as much of a bug fix related to the other changes, and may be more appropriate stand-alone as an improvement.  See this diff: https://github.com/NGEET/fates/pull/1226/commits/36a86e66c90324a2d3f195ce715d56ded02ce452

mpaiao (approved):
Sorry for taking so long to review this one! I went through it and the implementation looks good and I really like the approach using patchno. Thanks! 

I made only one really minor comment.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -966,6 +957,8 @@ subroutine phenology( currentSite, bc_in )
     integer  :: gddstart          ! beginning of counting period for growing degree days.
     integer  :: nlevroot          ! Number of rooting levels to consider
     real(r8) :: temp_in_C         ! daily averaged temperature in celsius
+    real(r8) :: temp_wgt          ! canopy area weighting factor for daily average
```
glemieux, line 960: Looks like #1306 snuck in here via 36a86e66c90324a2d3f195ce715d56ded02ce452.  Just needs a `git revert`.

main/EDMainMod.F90
```diff
@@ -835,6 +835,9 @@ subroutine ed_update_site( currentSite, bc_in, bc_out, is_restarting )
     type (fates_patch_type) , pointer :: currentPatch
     !-----------------------------------------------------------------------
 
+    ! check patch order (set second argument to true)
+    ! call set_patchno(currentSite,.true.,1)
+
```
glemieux, line 840: Did you want to keep this in just as a check?  I'm assuming it technically can be removed since you're calling `set_patchno` at the end of `ed_ecosystem_dynamics` via `terminate_patches` (which is a good move IMO).

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5065,27 +5059,23 @@ subroutine update_history_hifrq1(this,nc,nsites,sites,bc_in,bc_out,dt_tstep)
...
-
                hio_c_stomata_si(io_si) = hio_c_stomata_si(io_si) + &
                     cpatch%c_stomata * cpatch%total_canopy_area * mol_per_umol * site_area_veg_inv
 
                hio_c_lblayer_si(io_si) = hio_c_lblayer_si(io_si) + &
                     cpatch%c_lblayer * cpatch%total_canopy_area * mol_per_umol * site_area_veg_inv
 
                ! Only accumulate the instantaneous vegetation temperature for vegetated patches
-               if (cpatch%patchno .ne. 0) then
+               !if (cpatch%nocomp_pft_label.ne.nocomp_bareground)then
```
glemieux, line 5074: Should this be reinstated to avoid potentially adding the bareground patch into `hio_tveg`?

radiation/FatesRadiationDriveMod.F90
```diff
@@ -282,17 +279,15 @@ subroutine FatesSunShadeFracs(nsites, sites,bc_in,bc_out)
...
        do while (associated(cpatch))
 
-          if_notbareground:if(cpatch%nocomp_pft_label.ne.nocomp_bareground)then !only for veg patches
-             ! do not do albedo calculations for bare ground patch in SP mode
-             ! and (more impotantly) do not iterate ifp or it will mess up the indexing wherein
-             ! ifp=1 is the first vegetated patch.
-             ifp=ifp+1
+          ifp = cpatch%patchno
+          
+          nocomp_bareground:if(cpatch%nocomp_pft_label.ne.nocomp_bareground)then !only for veg patches
```
glemieux, line 287: The `end if` side of this block still shows `if_notbareground`.

main/FatesInterfaceMod.F90
```diff
@@ -774,8 +774,9 @@ subroutine SetFatesGlobalElements1(use_fates,surf_numpft,surf_numcft,param_reade
       integer,                    intent(in) :: surf_numpft  ! Number of PFTs in surface dataset
       integer,                    intent(in) :: surf_numcft  ! Number of CFTs in surface dataset
       class(fates_param_reader_type), intent(in) :: param_reader ! HLM-provided param file reader
-  
       integer :: fates_numpft  ! Number of PFTs tracked in FATES
+
+      logical, parameter :: preserve_b4b = .true.
```
glemieux, line 779: Do you want to keep this in for a future update?

glemieux (changes requested):
Nice work on the clean up.  I really like getting rid of the iterators and adding the calls to `set_patchno` at the end of `terminate_patches`, which makes more sense to me.

I'm recommending reverting 36a86e66c90324a2d3f195ce715d56ded02ce452 which you brought in with #1306.

biogeochem/EDPhysiologyMod.F90
glemieux: Actually I just test merged `sci.1.80.11_api.37.0.0` and it didn't seem to cause any issue.   Reverting this isn't necessary.

radiation/FatesRadiationDriveMod.F90
rgknox: actually, this will not compile on gnu compilers, I will def have to fix this

main/FatesHistoryInterfaceMod.F90
rgknox: yeah, need to clean some things up

main/EDMainMod.F90
rgknox: yes, or maybe we could put it behind a local debug flag. It is not needed, it is more of a developer level check

rgknox:
your earlier comment about tveg in history @glemieux  is correct. While debugging some things for the two-stream cleanup, I had merge in this branch, and that was one of the things that stood out.

main/EDTypesMod.F90
```diff
@@ -566,14 +568,51 @@ module EDTypesMod
...
   public :: CalculateTreeGrassAreaSite
+  public :: set_patchno
+  
+contains
 
-  contains
-      
-    ! =====================================================================================
+  ! ============================================================================
+
```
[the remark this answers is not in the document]
rgknox: Added description to set_patchno

main/FatesHistoryInterfaceMod.F90
rgknox: this is cleaned up now

radiation/FatesRadiationDriveMod.F90
rgknox: yeah, gnu compilers gave me trouble if I used a label that is also a variable. I think its a reasonable complaint. Reverted this loop label back to if_bareground.

main/FatesInterfaceMod.F90
rgknox: I wasn't sure when CTSM PR: [2898](https://github.com/ESCOMP/CTSM/pull/2898) would go through, but since its through, we can now remove this

rgknox: actually, lets keep this in there and implement the change in a separate PR after this one goes through. This will affect changes in results because we will have different (correct) patch counts.  I'd rather have this PR b4b

glemieux: Sounds good :+1:

glemieux (approved):

glemieux:
Regression testing on `derecho` against `fates-sci.1.80.13_api.37.0.0-ctsm5.3.024` is underway.

glemieux:
Regression testing is showing b4b results against `fates-sci.1.80.13_api.37.0.0-ctsm5.3.024` on `derecho`.

PR #1239, opened 2024-08-20
Add external land use data tool as a submodule to fates

glemieux:
### Description:
This simple PR adds the [land use data tool](https://github.com/glemieux/tools-fates-landusedata) as a submodule to fates in the `tools` directory.   The intent of this is to provide more "natural" discovery of the land use data tool, while allowing the tool to live in its own repository.  It also removes the older `luh2` directory which is superseded by the submodule. 

### Collaborators:
@ckoven 

No expected answer changes as this affects tooling only.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:
Testing in the following way TBD:

- [x] successful recursive `git submodule` checkout using CTSM and E3SM host models


ekluzek:
@glemieux looking at this, this should work fine with git-fleximod on the CESM/CTSM side of things. But, we'll want to make sure. There might be some additional fields that should (or could) be added for it. We should think about that a bit. I think the way this works with submodules is that it gives you the latest version on master for the submodule -- which is maybe what's desired here. But, that does mean that behavior of the tool can change out from under you as the main branch undergoes development...

glemieux:
@ekluzek I was just testing out the `git-fleximod` on my local ctsm repo.  Conducting the update procedure as I understand it (i.e. update `.gitmodules` and checkout and commit the test fates branch) appears to have worked correctly.  I also tested locally checking out within fates a different branch and `git-fleximod` appropriately checked out the PR branch on my test ctsm branch.  That said it did report the following:
```
e    tools/landusedata has no fxtag defined in .gitmodules, module at v0.1.0-24-g88108f3
               fates updated to tool-landuse-submodule
```
I'm not sure if this will be confusing to ctsm users, so perhaps it should be defined in the ctsm `.gitmodules` file?

Your point about the behavior is well taken.  For now we just really want to get the latest commit from the tool repo main branch as there really isn't any need to coordinate the tool's version relative to the fates version, but there could come a time in the future in which that would be desirable.

glemieux:
Testing note: using `git submodule update --init --recursive` in e3sm works as expected and pulls in the land use data tool contents.

ekluzek:
@glemieux we do see a similar message for cprnc in cime, so this is maybe OK. 

You could add the following two that would add a little more clarity for git-fleximod in CESM and be ignored by submodules in E3SM:

fxrequired = AlwaysRequired
# Standard Fork to compare to with "git fleximod test" to ensure personal forks aren't committed
fxDONOTUSEurl = https://github.com/NGEET/tools-fates-landusedata

But, then this might be confusing on the E3SM side of things. And it might not help enough on the CESM side.

If you lifted it outside of FATES and handled by the HLM that would allow it to be done differently between the two. But, if what we want for both is to get the head of the main branch does that help enough?

glemieux:
@ekluzek adding this to the fates `.gitmodules` sounds like a good idea, but I tried implementing this on a [fork from this branch](https://github.com/glemieux/fates/blob/submodule-fxtest/.gitmodules) and it didn't see to clear up that particular output.  Am I understanding that was the intent of your suggestion?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tools/landusedata"]
+	path = tools/landusedata
+	url = https://github.com/NGEET/tools-fates-landusedata
```
ekluzek, line 3: I suggest adding the following which would only be used by git-fleximod for CESM:

``` yaml
            # Two settings only used when git-fleximod is used (for CESM)
            # Always require this submodule since it's small so doesn't hurt to have it
            fxrequired = AlwaysRequired
            # Standard Fork to compare to with "git fleximod test" to ensure personal forks aren't committed
            fxDONOTUSEurl = https://github.com/NGEET/tools-fates-landusedata
```

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tools/landusedata"]
```
ekluzek, line 1: I suggest adding a header to explain what is going on here. Adapting the CTSM header I suggest something like this...

``` yaml
# This is a git submodule file to include additional submodules when FATES is checkout out.
#
# It includes optional additional support for
# git-fleximod (https://github.com/ESMCI/git-fleximod)
# that will be used when git-fleximod is used to check it out (i.e. for CESM)
#
# The additional flags supported by git-fleximod are
# fxtag - the tag associated with the submodule, this tag can be tested for
#         consistancy with the submodule hash using git-fleximod status
#         the hash can be updated to the tag using git-fleximod update
#
# fxrequired - indicates if a given submodule should be checked out on install
#              submodules can be toplevel or internal and required or optional
#              toplevel means that the submodule should only be checked out if the
#              module is the toplevel of the git repo (is not a submodule itself)
#              internal means that the submodule is needed by the component whether
#              the component is toplevel or the submodule of another repo
#              required means that the submodule should always be checked out
#              optional means that the submodule should only be checked out if the
#              optional flag is provided to git-fleximod or the submodule name is
#              explicitly listed on the git-fleximod command line.
#
# fxsparse -   this is a path to a git sparse checkout file indicating that the
#              submodule should be checked out in sparse mode
#
# fxDONOTUSEurl    -   this field is used by git-fleximod test to insure that the url is pointing
#              to the official url of the repo and not to an unofficial fork.
#              It is intended for use of github workflows to test commits to protected
#              repository branches.
#
```

ekluzek (approved):
This just moves the landusedata tool from directly in FATES to it's own submodule so that it can be an external and also go into conda forge etc.

I suggest adding some comments to explain what's going on, and to add some additional  optional support for git-fleximod. I clicked Approve though as you ditch that idea and also just so you don't have to have me reapprove later.

.gitmodules
glemieux: Thanks @ekluzek, I've added a trucated version of this.

.gitmodules
glemieux: Thanks.  I also added the `fxtag` since the landuse tool has tags.  Testing this out with `git-fleximod` resolves the issue I was seeing above.

PR #1236, opened 2024-08-17
Adding two new arctic shrub PFTs, and some arctic grass parameter updates

jenniferholm:
### Description:
To help improve the representation of arctic traits we are introducing two new shrubs that have been parameterized with an initial set of traits that more closely matches arctic observations, but improvements could still be made.  We have also updated a subset of parameter traits for the default c3 arctic grass PFT. 

These updates are based on the work done by Yanlan Liu and her 2024 parameter sensitive manuscript "Large divergence of projected high latitude vegetation composition and productivity due to functional trait uncertainty" https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2024EF004563
We introduce the new "broadleaf_evergreen_arctic_shrub" and "broadleaf_colddecid_arctic_shrub" PFTs, and update the proportional breakdown of only the "broadleaf deciduous boreal shrub" in the hlm_pft_map. 
This is so that non-arctic shrubs can still exist and compete with arctic shrubs, and to distinguish the traits/characteristics between the arctic environments that are pretty distinct.

This PR addresses the discussion in Issue #1204 

### Collaborators:
Help from @rgknox, @YanlanLiu, and conversations with @ckoven, and others. 

### Expectation of Answer Changes:
Will change the results of shrubs in high latitudes. 

### Checklist

*If this is your first time contributing, please read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/main/CONTRIBUTING.md) document.*

All checklist items must be checked to enable merging this pull request:

*Contributor*
- [x] The in-code documentation has been updated with descriptive comments
- [x] The documentation has been assessed to determine if updates are necessary

*Integrator*
- [x] FATES PASS/FAIL regression tests were run
- [x] Evaluation of test results for answer changes was performed and results provided

### Documentation

- [Technical Note](https://github.com/NGEET/fates-docs) update:
- [User's Guide](https://github.com/NGEET/fates-users-guide) update: 

### Test Results:

*CTSM (or) E3SM (specify which) test hash-tag:*

*CTSM (or) E3SM (specify which) baseline hash-tag:*

*FATES baseline hash-tag:*

*Test Output:*

PR #1255, opened 2024-10-02
API 36.1.0 merge

glemieux:
Integrating this PR will merge in #1136 ad #1236.  To be associated with ESCOMP/ctsm#2700.

### Description:
This is a pull request to coordinate the update of multiple parameter file updates at the same time.  Whereas #1136 and #1236 were tested separately with their own special parameter file updates, this PR will combined the parameter file updates into the single default source.  Part of the reason for setting things up this way is to provide more clarity in the testing, but also to capture the workflow necessary to update the default parameter file as it requires a sequential calls of `UpdateParamAPI` followed by `BatchPatchParams` tool on the #1136 and #1236 patch file updates respectively.

### Collaborators:
@JessicaNeedham @jenniferholm @rgknox @XiulinGao 

Yes, answers are expected to change due to the parameter file updates.  See the respective pull requests for more detail.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


glemieux:
-[ ] Update the default parameter file with the correct names for the new pfts

I think this requires a more up-to-date version of numpy than I'm currently using (`numpy==1.24.4`) as modify_fates_paramfile.py returns a warning and skips the update:
```
../tools/modify_fates_paramfile.py --fin /tmp/tmp.hdMg69N3Mi --fout /tmp/tmp.hdMg69N3Mi --var fates_pftname --silent  --val " broadleaf_colddecid_arctic_shrub" --nohist --overwrite --pft 11
output variable not interpretable as real. trying array
../tools/modify_fates_paramfile.py:84: DeprecationWarning: string or file could not be read to its end due to unmatched data; this will raise a ValueError in the future.
  outputval = np.fromstring(args.val, sep=',', dtype=np.float64)
Traceback (most recent call last):
  File "../tools/modify_fates_paramfile.py", line 355, in <module>
    main()
  File "../tools/modify_fates_paramfile.py", line 162, in main
    raise ValueError('variable dimension not handled in this script')
ValueError: variable dimension not handled in this script
```

rgknox:
@glemieux , if your stuck, can you print out the value of the string:

```
print("args.val: --{}--".format(args.val))
outputval = np.fromstring(args.val, sep=',', dtype=np.float64)
```

glemieux:
> @glemieux , if your stuck, can you print out the value of the string:
> 
> ```
> print("args.val: --{}--".format(args.val))
> outputval = np.fromstring(args.val, sep=',', dtype=np.float64)
> ```

Weirdly, the `np.fromstring` call on the new pft varname string doesn't result in a length of zero, which spot testing with a toy case suggests it should.  It's actually returning a single element array with a value of `-1`, which is odd :thinking:, and printing that deprecation warning.  As such, it never triggers an exception to cascade down into the next `try` level here and set `rename_pft` to `True`:
https://github.com/NGEET/fates/blob/e06e0dfcaf6347993d47dc29944e952d3e4412f7/tools/modify_fates_paramfile.py#L84-L91

That said, I think I have a fix in which just involves checking to see if we're modifying the `fates_pftname` a bit early in that routine.

glemieux:
@rgknox looks like the fix I brought in via 204a061 this works.  Would you take a look and give this PR an approval if it looks good to you?

glemieux:
Regression testing on `derecho` is complete.  There are no unexpected failures.  All testmods have DIFFs against the baseline (with the exception of the one ctsm-only testmod), which is expected given the extent of the parameter file changes and noted in https://github.com/NGEET/fates/pull/1136#issuecomment-2291348549.  Spot checking the satellite phenology SMS test shows results similar to those found in https://github.com/NGEET/fates/pull/1236#issuecomment-2389505278. 

This PR should be good to go.

glemieux:
Testing on izumi is underway.

glemieux:
Izumi testing is complete.  There are no unexpected failures.  

Location: `/scratch/cluster/glemieux/ctsm-tests/tests_1004-141329iz`

PR #1269, opened 2024-10-23
handle case where LAI+SAI exceeds array by clipping instead of crashing

ckoven:
There has for a long time (always?) been an issue where certain leaf and lai parameters can generate absurdly large LAI values within the crown of a plant (where plant-level LAI equals leaf area / crown area), and which can crash the model, e.g. as per https://github.com/NGEET/fates/discussions/1190.  Now that our default crown area and leaf area allometric exponents are different, this can happen mid-run, as the plants start to get larger. And I am finding that in rare edge cases in a global run, this does in fact happen.

We've discussed for a long time just putting an upper limit to how large LAI can get, in order to allow the model to keep going rather than crash. The argument being that for these plants, their lowermost leaf area is in the dark and so not going to be doing a lot of photosynthesizing, so it should in principle be self-correcting.  This code does that: it just says, basically, that beyond a certain point, a plant can put on more leaf biomass, but it won't have any effect on the LAI. 

I've only tested it to the extent that I can confirm it prevents the model from crashing under whatever the edge case that I had run into that was previously causing the model to crash.  Further investigation is needed to assess the effect of this on simulations.

fixes #1190.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


rgknox:
Notes from SE Meeting:
1) add a wrapper to update tree lai and tree sai in the cohort module
2) try to see if we can update this in one location

rgknox:
tested up to commit 2c654d9 against base and is b4b using fates test suite

glemieux (approved):
Nice work reusing the `treelai` and `treesai` functions.

Issue #1276, opened 2024-10-31
remote but possible singularity in two-stream code


rgknox:
We have some code that avoids a mathematical singularity in two-stream, see here:

https://github.com/NGEET/fates/blob/main/radiation/TwoStreamMLPEMod.F90#L952-L977

I just realized that it's possible that a correction that is applied to prevent a singularity in the NIR band (second loop index) could actually generate conditions for a singularity in the visible band (the first loop index).  

This seems incredibly unlikely, but when you do this billions of times in and ESM, a one in a billion probability could crash your model.

PR #1277, opened 2024-10-31
comprehensive correction for singularity in two-stream

rgknox:
### Description:

This adds a more thorough correction to avoid singularities in the two-stream radiation. To be specific, we make sure that a correction made in the near-infrared band does not create a singularity in the visible band.

Fixes: #1276

Its possible but unlikely we would see differences in a test. Any test that was vulnerable would probably had failed.  This is more about future-proofing the code.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/radiation/TwoStreamMLPEMod.F90
+++ b/radiation/TwoStreamMLPEMod.F90
@@ -993,7 +993,7 @@ contains
                      call endrun(msg=errMsg(sourcefile, __LINE__))
                   end if
                   ! Test to see if there is a singularity and make corrections if needed
-                  if any((abs(Kb_sing(:) - Kb_eff)) < sing_tol) then
+                  if (any((abs(Kb_sing(:) - Kb_eff)) < sing_tol)) then
                      Kb_eff = Kb_eff + sing_tol
                      is_sing = .true.
                   end if
--- a/radiation/TwoStreamMLPEMod.F90
+++ b/radiation/TwoStreamMLPEMod.F90
@@ -979,10 +979,12 @@ contains
                iter_sing = 0
 
                ! Compute the singularity for all bands
-               Kb_sing(:) = this%band(:)%scelb(ican,icol)%a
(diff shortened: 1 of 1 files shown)


glemieux (approved):

glemieux:
Regression testing underway on derecho.

glemieux:
Regression testing against `sci.1.80.1_api.37.0.0` is complete.  All expected tests pass B4B with the exception of the `FatesColdTwoStream` tests.  The difference show up on the first time step after initialization (`t_index = 2`) and appear to be round off (`< E-13`).  The following variables are different:

```
FATES_GPP
FATES_GPP_CANOPY
FATES_HET_RESP
FATES_LBLAYER_COND
FATES_MAINT_RESP
FATES_NEP
FATES_STOMATAL_COND
```

Note that of the above, only `FATES_STOMATAL_COND` and `FATES_LBLAYER_COND` are not round off.  After this time step, more differences propagate.

I decided to double check the original commit cd33abc9676b5178d5839eddbd347285f3a89abb against the baseline to make sure that my later refactor didn't inject the differences.  The results are showing the same differing variables and similar RMS values/

@rgknox I'm not sure why these particular variables would be impacted by this update.  I spot checked the baseline to make sure it was return b4b, so I'm pretty sure the issue isn't there.  Thoughts?

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1277`

rgknox:
THe differences in FATES_HET_RESP are kind of suspicious... That is soil respiration. I suppose radiation scattering differences could affect soil heating, which affects respiration rates. But it seems unlikely in this case?

rgknox:
After discussing with @glemieux and getting a better sense of the magnitude of the differences, I think they fall within the realm of expectation.

PR #1306, opened 2025-01-09
Fix vegetation temperature weighting during phenology

rgknox:
### Description:

This change improves how we calculate vegetation temperature during phenology.  The previous method included bare ground patches in the mean.  Bare ground temperatures are initialized with the column/site level air temperature, but those values are not as representative of using only patch level vegetation temperatures.  This new method filters in only temperatures from patches with vegetation, and uses the canopy area as the weighting factor, instead of total patch area.

This should precede #1226 

### Collaborators:
@rosiealice 

Yes, this should change answers, at least for tests that are long enough to trigger phenology events.

### Checklist

*Contributor*
- [x] The documentation has been assessed to determine if updates are necessary (NOT NECESSARY)

*Integrator*

### Test Results:

TBD

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -966,6 +966,8 @@ contains
     integer  :: gddstart          ! beginning of counting period for growing degree days.
     integer  :: nlevroot          ! Number of rooting levels to consider
     real(r8) :: temp_in_C         ! daily averaged temperature in celsius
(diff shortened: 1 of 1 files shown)


glemieux (approved):
This looks straightforward.

ckoven (approved):
This looks good to me. Thanks @rgknox!

mpaiao (approved):
Thanks for these edits @rgknox, they look good and reasonable to me too!

Issue #1279, opened 2024-11-04
Error message about PARPROF_PFT_DIR_Z


jennykowalcz:
### Discussed in https://github.com/NGEET/fates/discussions/1274

<div type='discussions-op-text'>

<sup>Originally posted by **jennykowalcz** October 30, 2024</sup>
Hi all! I'm looking for help understanding an error message I have not encountered before. My FATES run failed after ~195 years with:
<img width="835" alt="Screenshot 2024-10-29 at 1 35 54 PM" src="https://github.com/user-attachments/assets/2defdd69-31b5-4fce-8e3e-372768097673">

I rebuilt with DEBUG=TRUE and now it reads:
<img width="899" alt="Screenshot 2024-10-30 at 10 19 41 AM" src="https://github.com/user-attachments/assets/f3335dd8-946e-44ce-8aa0-69bd33d00a45">

This is a full-complexity FATES run over tropical S. America with two evergreen tropical tree PFTs representing early and late successionals. I'm using two-stream radiation and have nclmax set to 3.

</div>

rgknox:
@jennykowalcz thanks for submitting, I think we should be looping over a number of vegetation layers that is governed by "nveg", like we do here when we allocate:

https://github.com/NGEET/fates/blob/main/biogeochem/FatesPatchMod.F90#L314C7-L314C37

And not by this line here:

https://github.com/NGEET/fates/blob/main/radiation/FatesRadiationDriveMod.F90#L446C31-L446C84

PR #1310, opened 2025-01-13
alternative way to adjust area issues in FATES twostream

rgknox:
### Description:

Its possible that there is more horizontal area taken up by the cohorts than there is ground, which is simply a result of numerical imprecision in the rest of FATES. If this is true, then we need to somehow force the total area of the two-stream scattering elements to be exactly 1 and not slightly more. One way is to just chop off some area of the largest scattering element (simple way). This is simple, yet does not conserve total scattering depth.  The other (which we already allow) is to chop off that area but also increase the LAI+SAI. The latter is conservative, but is creating indexing problems when when the LAI and SAI is increased enough such that it uses more depth bins than FATES anticipated.

### Collaborators:

@jennykowalcz 

Good chance there will be very slight differences in two-stream tests.  There should be no scientifically meaningful impacts to results.

### Checklist

*Contributor*

*Integrator*

### Test Results:

TBD

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -2274,6 +2274,13 @@ contains
 
    ! Number of actual vegetation layers in this cohort's crown
    currentCohort%nv =  count((currentCohort%treelai+currentCohort%treesai) .gt. dlower_vai(:)) + 1
+
+   if( currentCohort%nv .ne. minloc(dlower_vai, DIM=1, MASK=(dlower_vai>(currentCohort%treelai+currentCohort%treesai))) ) then
+      write(fates_log(),*) 'We use two methods of finding maximum leaf layers, and they are not equivalent'
+      write(fates_log(),*) 'count method:',currentCohort%nv
+      write(fates_log(),*) 'minloc method:',minloc(dlower_vai, DIM=1, MASK=(dlower_vai>(currentCohort%treelai+currentCohort%treesai)))
+      call endrun(msg=errMsg(sourcefile, __LINE__))
+   end if
    
   end subroutine UpdateCohortLAI
   
--- a/radiation/FatesTwoStreamUtilsMod.F90
+++ b/radiation/FatesTwoStreamUtilsMod.F90
@@ -78,6 +78,20 @@ contains
     integer :: max_elements     ! Maximum number of scattering elements on the site
     integer :: n_scr            ! The size of the scratch arrays
     logical :: allocate_scratch ! Whether to re-allocate the scratch arrays
+    integer  :: icolmax         ! Column index for each layer with largest are footprint
+    real(r8) :: areamax         ! The area footprint of the largest column
+    
+    ! its possible that there is more horizontal area taken up by the cohorts
+    ! than there is ground, which is simply a result numerical and algorithmic
+    ! imprecision in the rest of FATES. If this is true, then we need
+    ! to somehow force the total area of our scattering elements to be exactly
+    ! 1 and not slightly more. One way is to just chop off some area of the
+    ! largest scattering element (simple way), the other is to chop off that
+    ! area but also increase the LAI+SAI. The latter is conservative, but
+    ! could create indexing problems when transfering fluxes back into FATES arrays
+    
+    logical, parameter :: do_simple_area_correct = .true.
+    
     
     ! These parameters are not used yet
     !real(r8) :: max_vai_diff_per_elem ! The maximum vai difference in any element
@@ -119,6 +133,13 @@ contains
          ! an air element is needed for all the non
          ! occupied space, even if the canopy_open_frac
          ! is zero.
+         ! If the area of the elements does not match
+         ! the area of the canopy space within 1.e-7_r8
+         ! then we either add the space in the form of air
+         ! or we compress the space by literally squeezing
+         ! the elements (which consequently increases their
+         ! LAI and SAI to conserve area)
+        
             
          if(patch%total_canopy_area>nearzero)then
             canopy_frac(:) = 0._r8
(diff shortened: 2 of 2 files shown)


biogeochem/EDCanopyStructureMod.F90
jennykowalcz, line 1: Hi Ryan! This is just a question for my own understanding. VAI is measured from the top of the canopy down, right? Is dlower_vai the the lower edge of the bin numerically or lower spatially?

radiation/FatesTwoStreamUtilsMod.F90
```diff
@@ -78,6 +78,20 @@ subroutine FatesConstructRadElements(site,fcansno_pa,coszen_pa)
     integer :: max_elements     ! Maximum number of scattering elements on the site
     integer :: n_scr            ! The size of the scratch arrays
     logical :: allocate_scratch ! Whether to re-allocate the scratch arrays
+    integer  :: icolmax         ! Column index for each layer with largest are footprint
```
jennykowalcz, line 81: Inconsequential typo -- largest area

jennykowalcz (approved):
This looks good. I tested this PR with one of the cases that encountered [this error](https://github.com/NGEET/fates/issues/1279).  Without this fix, the case fails with

`219: forrtl: severe (408): fort: (2): Subscript #3 of the array PARPROF_PFT_DIR_Z has value 12 which is greater than the upper bound of 11`


but with this fix runs successfully well past this point.

radiation/FatesTwoStreamUtilsMod.F90
```diff
@@ -230,24 +251,30 @@ subroutine FatesConstructRadElements(site,fcansno_pa,coszen_pa)
...
                   end if
                end do
                
-               !write(fates_log(),*) 'overfull areas'
-               !twostr%cosz = coszen_pa(ifp)
-               !     call twostr%Dump(1,lat=site%lat,lon=site%lon)
-               !     call endrun(msg=errMsg(sourcefile, __LINE__))
+               ! Test out a simpler way to correct area errors
+               if(do_simple_area_correct) then
+                  twostr%scelg(ican,icolmax)%area = twostr%scelg(ican,icolmax)%area - (canopy_frac(ican)-1._r8)
```
glemieux, line 270: The `canopy_frac(ican)` here is going to be greater than `1` at this point in the code correct?  Is it possible that the value of `canopy_frac(ican)-1._r8` could be greater than `area`?  If so we might end up with a negative final area and should have a limiter against this.

glemieux (commented):
This looks pretty good, although I have one question about a possible necessary check against a negative area.

biogeochem/EDCanopyStructureMod.F90
rgknox: lower spatially, good question!

radiation/FatesTwoStreamUtilsMod.F90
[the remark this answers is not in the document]
rgknox: Good question. I suppose it would be better to define an area error threshold, where exceedance would abort the run.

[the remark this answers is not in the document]
rgknox: @jennykowalcz , I added in an abort if the canopy layer is more than 1% overfull.

radiation/FatesTwoStreamUtilsMod.F90
rgknox: The area being used here is from the element that has the largest area of the group. The group sums to 1. In a worst case scenario, lets say we had 1000 cohorts all sharing one layer, then the area of the largest element would be no less than 0.001.  So in this case, the error would have to be larger than 0.001. Therefore, I don't think this should be a problem, but it shouldn't hurt to add in a check and complain/fail gracefully if this does happen.

rgknox: I added in a check. I put it behind a debug for now. I can run some tests to see if they trigger, and then set the debug flags back to false so that they don't slow down the run.  Note that we really should not have layers that are that overfull, this is because we try to make the cohorts fit correctly during the ppa demotion/promotion scheme. These methods here try to make the area convergence tolerance even tighter, so that it doesn't introduce energy balance errors into the model.

biogeochem/EDCanopyStructureMod.F90
jennykowalcz: Okay that makes sense, thanks!

radiation/FatesTwoStreamUtilsMod.F90
rgknox: 30 day f45 test generated no errors on the checks , switching developer debug flag back to false

rgknox:
Tests pass on derecho:

```
./cs.status.fails
0219-103038de_gnu: 10 tests
    FAIL ERS_D_Ld20.f45_f45_mg37.I2000Clm50FatesRs.derecho_gnu.clm-FatesColdTwoStreamNoCompFixedBioGeo COMPARE_base_rest (EXPECTED FAILURE)
    FAIL PEM_D_Ld20.5x5_amazon.I2000Clm50FatesRs.derecho_gnu.clm-FatesColdSeedDisp COMPARE_base_modpes (EXPECTED FAILURE)
    FAIL SMS_D_Ld3.f09_g17.I2000Clm60FatesSpCruRsGs.derecho_gnu.clm-FatesColdSatPhen_prescribed RUN time=87 (EXPECTED FAILURE)
    PASS SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Fates.derecho_gnu.clm-FatesPRISM--clm-NEON-FATES-YELL SHAREDLIB_BUILD time=123 (UNEXPECTED: expected FAIL)
    PASS SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Fates.derecho_gnu.clm-FatesPRISM--clm-NEON-FATES-YELL RUN time=101 (UNEXPECTED: expected FAIL)

 
0219-103038de_int: 38 tests
    FAIL ERP_P256x2_Ld30.f45_f45_mg37.I2000Clm60FatesRs.derecho_intel.clm-mimicsFatesCold RUN time=44 (EXPECTED FAILURE)
    PEND ERP_P256x2_Ld30.f45_f45_mg37.I2000Clm60FatesRs.derecho_intel.clm-mimicsFatesCold COMPARE_base_rest
    FAIL ERS_D_Ld20.f45_f45_mg37.I2000Clm50FatesRs.derecho_intel.clm-FatesColdTwoStream COMPARE_base_rest (EXPECTED FAILURE)
    FAIL ERS_Ld60.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesColdST3 RUN time=26 (EXPECTED FAILURE)
    FAIL PVT_Lm3.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesLUPFT RUN time=1233 (EXPECTED FAILURE)
    PASS SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60Fates.derecho_intel.clm-FatesFireLightningPopDens--clm-NEON-FATES-NIWO SHAREDLIB_BUILD time=97 (UNEXPECTED: expected FAIL)

 
0219-103038de_nvh: 1 test

 
========================================================================
Non-PASS results for select phases:
TPUTCOMP non-passes: 0
MEMCOMP non-passes: 0
```

PR #1314, opened 2025-01-14
Additional checks for crown depth allometry parameters

mpaiao:
### Description:

This pull request intends to address the discussion in #1308, by adding multiple sanity checks on the parameters for height to crown depth allometry in subroutine `PRTCheckParams`. With this revision, it will issue either an error or a warning, depending on the severity of the problem.

1. **Fatal errors**
  - When `fates_allom_h2cd1 ≤ 0.`
  - When `fates_allom_dmode = 1` and `fates_allom_h2cd1 > 1`.
  - When `fates_allom_dmode = 2`,  `fates_allom_h2cd2 = 1` and `fates_allom_h2cd1 > 1` (which is identical to previous case).
2. **Warning**. 
  - When `fates_allom_dmode = 2`,  `fates_allom_h2cd1 < 1` and `fates_allom_h2cd1` is such that crown depth would exceed height for small cohorts. 
  - When `fates_allom_dmode = 2`,  `fates_allom_h2cd1 > 1` and `fates_allom_h2cd1` is such that crown depth would exceed height for large cohorts. 

In the warning cases, FATES was already applying a cap to prevent crown depth to exceed height, but this was done quietly. The code now issues a warning during initialisation.

### Collaborators:

@glemieux @rgknox @ekluzek

No change in the results expected, except that now the code will not even run if the parameters are out of bounds, or at least issue a warning when some potentially unwanted behaviour occurs. 

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/parteh/PRTParamsFATESMod.F90
+++ b/parteh/PRTParamsFATESMod.F90
@@ -1423,7 +1423,7 @@ contains
         case (2)
            ! Log-linear allometry. Multiplier factor cannot be negative or zero.
            ! -------------------------------------------------------------------------------
-           if ( prt_params%allom_h2cd1 (ipft) <= nearzero ) then
+           if ( prt_params%allom_h2cd1 (ipft) < nearzero ) then
               !   Calculations for the generic case require allom_h2cd1 to be positive. If
               ! not, issue an error.
               ! ----------------------------------------------------------------------------
@@ -1472,8 +1472,8 @@ contains
               ! ----------------------------------------------------------------------------
               call h_allom(prt_params%allom_dbh_maxheight(ipft),ipft,height_max)
 
-              if ( ( prt_params%allom_h2cd2 (ipft)  <  1.0_r8      ) .and. &
-                   ( EDPftvarcon_inst%hgt_min(ipft) <= height_crit ) ) then
+              if ( ( prt_params%allom_h2cd2 (ipft)  <  1.0_r8     ) .and. &
+                   ( EDPftvarcon_inst%hgt_min(ipft) < height_crit ) ) then
                  !   These parameters will cause the code to cap crown depth to height for
                  ! small plants. We print a warning message, but we do not stop the run.
                  ! -------------------------------------------------------------------------
--- a/parteh/PRTParamsFATESMod.F90
+++ b/parteh/PRTParamsFATESMod.F90
@@ -1128,9 +1128,14 @@ contains
      logical :: is_season_decid ! Is the PFT cold-deciduous?
      logical :: is_stress_decid ! Is the PFT drought-deciduous?
      logical :: is_semi_decid   ! Is the PFT drought semi-deciduous?
+     logical :: is_hmode_fine   ! Did the height allometry pass the check?
      integer :: nerror          ! Count number of errors. If this is not
                                 !    zero by theend of the subroutine, stop 
                                 !    the run.
+     real(r8) :: height_crit    ! Critical height where crown depth equals height
+     real(r8) :: height_max     ! Maximum height attainable by PFT.
+
+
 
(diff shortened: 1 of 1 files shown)


parteh/PRTParamsFATESMod.F90
```diff
@@ -1383,8 +1392,154 @@ subroutine PRTCheckParams(is_master)
...
+              ! ----------------------------------------------------------------------------
+              !    General log-linear case. Depending on the parameter values, crown depth 
+              ! may exceed for very small or very large plants. The code has safeguards 
+              ! to prevent this behaviour, but we must at least issue some warning.
+              ! ----------------------------------------------------------------------------
+
+              !   Find the critical height in which crown depth becomes height
+              ! ----------------------------------------------------------------------------
+              height_crit = prt_params%allom_h2cd1 (ipft) ** &
+                           ( 1.0_r8 / (1.0_r8 - prt_params%allom_h2cd2 (ipft)) )
```
glemieux, line 1469: Since we don't end the run till the bottom of this section, it's possible that we could have `is_hode_fine = true` and `allom_h2cd2 = 1.0` which we result in a divide by zero here, correct?

glemieux (commented):
I really like the addition of the warnings checking against the critical height.  That said there is one issue that I think we need to address below.

parteh/PRTParamsFATESMod.F90
mpaiao: > Since we don't end the run till the bottom of this section, it's possible that we could have `is_hode_fine = true` and `allom_h2cd2 = 1.0` which we result in a divide by zero here, correct?

I think this should never happen, because when `prt_params%allom_h2cd2 (ipft) = 1.0_r8`, then the code would enter in the `elseif` block right before this one instead (the one that checks `abs(prt_params%allom_h2cd2 (ipft) - 1._r8 ) < nearzero`).  

I added that previous block to avoid the singularity, and also because that case is numerically the same case as the linear crown depth allometry (`prt_params%allom_dmode(ipft)) =1`), so the code runs the same checks.

glemieux: Gotcha, I see now.  I overlooked that the previous `elseif` statement was in the same `if..end if` chain.  Thanks!

glemieux (approved):

glemieux:
Regression testing on `derecho` against `sci.1.80.9_api.37.0.0-ctsm5.3.020` is B4B.

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1314-fates`

glemieux:
I've confirmed the three different failure modes list above.

Failure mode 1 message:
```
 ---~---
  Incorrect settings for crown depth allometry.
  PFT index:             14
  allom_dmode:            1
  allom_h2cd1:   -1.0000000000000000
  Parameter "allom_h2cd1" must be positive and <= 1
     when allom_dmode = 1 (or allom_h2cd2 = 1).
 ---~---
```

Failure mode 2 message:
```
 ---~---
  Incorrect settings for crown depth allometry.
  PFT index:             14
  allom_dmode:            1
  allom_h2cd1:    1.2000000000000000
  Parameter "allom_h2cd1" must be positive and <= 1
     when allom_dmode = 1 (or allom_h2cd2 = 1).
 ---~---
```

Failure mode 3 message:
```
 ---~---
  Incorrect settings for crown depth allometry.
  PFT index:             14
  allom_dmode:            2
  allom_h2cd1:    0.0000000000000000
  Parameter "allom_h2cd1" must be positive when
     allom_dmode = 2.
 ---~---
```

Issue #1313, opened 2025-01-14
FATES insert_cohort subroutine does not preserve order of cohorts


adrifoster:
The `insert_cohort` subroutine flips the order of cohorts (given identical heights), while `sort_cohorts` preserves order.

Essentially, this means that if we give `insert_cohort` a list of cohorts 1...100 (all the same exact height), the resulting list will be ordered 100...1. 

This is because of [this line](https://github.com/NGEET/fates/blob/1689466cfe449bad5c5370420d7a5e4701ae2e05/biogeochem/EDCohortDynamicsMod.F90#L1392): 

```
    current => pshortest
    exitloop = 0
    !starting with shortest tree on the grid, find tree just
    !taller than tree being considered and return its pointer
    if (associated(current)) then
       do while (associated(current).and.exitloop == 0)
          **if (current%height < tsp) then**
             current => current%taller
          else
             exitloop = 1
          endif
       enddo
    endif
```

if we switch this to  **if (current%height <= tsp) then** we will preserve order.

I think this is not ideal, and is somewhat confusing. I would like to propose we make this change this so that `insert_cohort` preserves the order.

I have a branch that updates our `insert_cohort` and `sort_cohort` routines. Because of this change, it is not B4B. I also propose adding some checks to make sure we don't have any inadvertently disconnected cohorts.

PR #1317, opened 2025-01-23
Update to cohort insertion and sorting routines

adrifoster:
### Description:
Updates the cohort insertion (`insert_cohort`) and sorting (`sort_cohorts`) to be more in line with best-practices for linked list methods. Also makes them a bound method of the `patch` class. 

Also adds some unit testing to demonstrate utility and functioning of methods.

Fixes issue #1313 

### Collaborators:
@billsacks @glemieux @ekluzek 

yes, will not be b4b because it fixes issue #1313, but I updated the existing `sort_cohorts` and `insert_cohort` to fix the issue and this PR is b4b with **that** branch 

/glade/derecho/scratch/afoster/tests_0122-124517de/SMS_Lm6.f45_f45_mg37.I2000Clm50FatesCruRsGs.derecho_intel.clm-Fates.C.0122-124517de_int

### Checklist
*Integrator*

### Test Results:

*CTSM (or) E3SM (specify which) test hash-tag:* ctsm5.3.020

*CTSM (or) E3SM (specify which) baseline hash-tag:* ctsm5.3.020

*FATES baseline hash-tag:* sci.1.80.9_api.37.0.0

*Test Output:*


testing/unit_testing/count_cohorts_test/test_CountCohorts.pf
```diff
@@ -0,0 +1,98 @@
...
+      this%patch%tallest => new_node
+      
+    end subroutine setUp
+    
+    subroutine tearDown(this)
+      class(TestCountCohorts), intent(inout) :: this        ! test object
+      type(fates_cohort_type), pointer       :: cohort      ! cohort object
+      type(fates_cohort_type), pointer       :: next_cohort ! next cohort object
+      
+      ! deallocate cohorts
```
billsacks, line 83: I have never used the finalization feature of Fortran (keyword `final`), but this seems like a possibly-appropriate use for it – i.e., introducing a finalization routine for `fates_patch_type` that deallocates the cohort linked list, so that when tests (or maybe production code, too) deallocates a patch, this underlying cohort deallocation happens automatically (if I understand finalization right).

But that's just an idea - not something you necessarily have to do.

testing/unit_testing/count_cohorts_test/test_CountCohorts.pf
```diff
@@ -0,0 +1,98 @@
...
+  use FatesConstantsMod, only : r8 => fates_r8
+  use FatesCohortMod,    only : fates_cohort_type
+  use FatesPatchMod,     only : fates_patch_type
+  use funit
+
+  implicit none
+  
+  @TestCase
+  type, extends(TestCase) :: TestCountCohorts
+    type(fates_patch_type), pointer :: patch
```
billsacks, line 15: Is there a reason the patch needs to be a pointer here? (Maybe there really is, in which case you can ignore this question!)

testing/unit_testing/count_cohorts_test/test_CountCohorts.pf
```diff
@@ -0,0 +1,98 @@
...
+    contains
+      procedure :: setUp
+      procedure :: tearDown
+  end type TestCountCohorts
+
+  real(r8), parameter :: tol = 1.e-13_r8
+
+  contains 
+  
+    subroutine setUp(this)
```
billsacks, line 25: It's perhaps premature to comment here since you haven't added the actual unit tests for CountCohorts yet, but I'm anticipating that you'll have multiple tests where you want to test different scenarios, so I'm wondering if the stuff currently in setUp should be moved out of here, so each test can test something different. I like the idea of consolidating this setup stuff into a subroutine rather than having it inline in a test; you could do that by having the test call a named subroutine like "call setup_patch_with_8_cohorts", where you put the current code.

Update: Oh, maybe this is just an artifact of starting with what you had for test_insertCohort.pf, and you plan to rework this.

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
```diff
@@ -0,0 +1,331 @@
...
+      allocate(cohort)
+      cohort%height = height
+      call this%patch%InsertCohort(cohort)
+    
+      ! verify that new cohort is now the shortest
+      @assertTrue(associated(cohort, this%patch%shortest))
+      
+      ! verify that new new shortest cohort's taller pointer is the previous shortest
+      @assertTrue(associated(cohort%taller, previous_shortest))
+
```
billsacks, line 142: Perhaps also worth testing that previous_shortest%shorter points to `cohort`?

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
```diff
@@ -0,0 +1,331 @@
...
+      allocate(cohort)
+      cohort%height = height
+      call this%patch%InsertCohort(cohort)
+    
+      ! verify that new cohort is now the tallest
+      @assertTrue(associated(cohort, this%patch%tallest))
+      
+      ! verify that new new tallest cohort's shorter pointer is the previous tallest
+      @assertTrue(associated(cohort%shorter, previous_tallest))
+
```
billsacks, line 166: Perhaps also worth testing that previous_tallest%taller points to `cohort`?

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
```diff
@@ -0,0 +1,331 @@
...
+      real(r8),                parameter     :: height = 35.0_r8 ! cohort height
+        
+      ! allocate patch
+      allocate(patch)
+      
+      ! insert cohorts of same height
+      do i = 1, 100
+        allocate(cohort)
+        cohort%height = height
+        cohort%dbh = i
```
billsacks, line 311: Ah, a clever way to verify this!

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
```diff
@@ -0,0 +1,331 @@
...
+      type(fates_cohort_type), pointer       :: cohort           ! cohort object
+      integer                                :: i                ! looping index
+      real(r8)                               :: dbh              ! cohort diameter
+      real(r8),                parameter     :: height = 35.0_r8 ! cohort height
+        
+      ! allocate patch
+      allocate(patch)
+      
+      ! insert cohorts of same height
+      do i = 1, 100
```
billsacks, line 308: 100 seems like it's probably overkill here: if it works for 2 or 3, it's hard to imagine what would cause it to fail for more. Not a big deal, though, other than the minor performance hit.

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
```diff
@@ -0,0 +1,331 @@
+module test_InsertCohort
```
billsacks, line 1: This test module provides a very nice set of tests of InsertCohort - this is a really valuable addition!

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
```diff
@@ -0,0 +1,425 @@
...
+      
+      allocate(new_node)
+      new_node%height = heights(7)
+      head%taller%taller%taller%taller%taller%taller => new_node
+      new_node%shorter => head%taller%taller%taller%taller%taller
+      
+      allocate(new_node)
+      new_node%height = heights(8)
+      head%taller%taller%taller%taller%taller%taller%taller => new_node
+      new_node%shorter => head%taller%taller%taller%taller%taller%taller
```
billsacks, line 145: (This comment also applies to a number of other places where you set up the list manually like this.)

I'm curious why you need to set things up manually instead of calling InsertCohort; if there's a reason that's needed or you see it as better, it may be worth adding a comment explaining the reason.

If you do keep this manual creation, what would you think about introducing a temporary pointer like `previous_node` where you set `previous_node => new_node` before reallocating `new_node`. That would allow cleaning up these long strings of `taller%taller...` by just referring to `previous_node`, which would be less error-prone.

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
```diff
@@ -0,0 +1,425 @@
...
+      patch%tallest => new_node
+      
+      ! sort cohorts
+      call patch%SortCohorts()
+      
+      ! test that the order is correct
+      i = size(heights)
+      new_node => patch%shortest
+      do while (associated(new_node))
+        @assertEqual(new_node%height, heights(i), tolerance=tol)
```
billsacks, line 224: Here and elsewhere: I don't think you should need a tolerance, since nothing should adjust the heights from their hard-coded values. (But I guess it doesn't really hurt to have this.)

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
```diff
@@ -0,0 +1,425 @@
...
+        new_node => new_node%taller
+        i = i - 1
+      end do
+           
+    end subroutine SmallList_DescendingOrder_Reversed
+    
+    @Test 
+    subroutine SmallList_RandomOrder_SortedCorrectly(this)
```
billsacks, line 232: `RandomOrder` gives the impression that you actually use some randomness, which you don't (and I think it's a good thing that you don't). Perhaps change this to `SmallList_Unordered_SortedCorrectly`?

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
```diff
@@ -0,0 +1,425 @@
+module test_SortCohorts
```
billsacks, line 1: Again, this is an excellent set of tests which adds confidence to this important routine.

testing/unit_testing/validate_cohorts_test/test_ValidateCohorts.pf
```diff
@@ -0,0 +1,145 @@
...
+      patch%shortest => cohort1
+      patch%tallest => cohort3
+      
+      cohort1%taller => cohort2
+      cohort2%shorter => cohort1
+      cohort2%taller => cohort3
+      cohort3%shorter => null()  ! break the forward link
+      
+      ! should fail
+      call patch%ValidateCohorts()
```
billsacks, line 141: I'm thinking there should be an `assertExceptionRaised` call here?

billsacks (changes requested):
This is a really valuable set of changes! Actually, there are (broadly speaking) three sets of changes here, all of which are very valuable: (1) Reworking InsertCohorts and SortCohorts to make them cleaner easier to understand; (2) Adding unit tests for these routines; and (3) Providing a structure for setting up synthetic patches for functional unit testing - a path that I think is going to pay off substantially in facilitating both functional unit testing and more traditional unit testing of FATES.

I have a few relatively minor suggestions, but overall this is excellent!

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
ekluzek: Since, this is in the unit tests, I think you should do overkill. Unless it was an obvious problem in total run length.

When I saw the arbitrary number I thought it might in the code, and something that might need to be tied to come other variable. Or that there might need to be logic to handle the case when it's exceeded. But, being in the unit tester I think 100 is fine and I'm glad it's higher than we could ever expect happening in a simulation. I'd say more than a 100 would be getting into overkill.

ekluzek (approved):
I really love seeing this come in! I've long been nervous that in FATES we have a complex data-structure (linked-lists) being used without unit testing built around it. This is such a critically important part of the FATES Software design that it really needs rigorous testing around it. It's also something that would be easy to go wrong, and without this kind of testing be really difficult to find later. And it could be something to give subtly wrong answers, and not have a clear way to find what went wrong. Those type of debugging problems are difficult to even detect, and very difficult to track down. They are intractable problems that you can't put time estimates to fix on. Having unit testing put in now though means three things.

1. We'll easily see when something breaks immediately
2. Finding the problem when it breaks will be more straight forward
3. So it will save multiple people days or weeks of debugging time down the line

Which is a huge benefit for FATES. I think it's hard to overstate how helpful bringing this is in will be. I'd class it as a game-changer for FATES development. It will allow us to refactor parts of the code that we couldn't before -- because we couldn't have confidence that the changes would be correct. So it allows the design of FATES to improve with time with more of an Agile Software Development Methodology workflow. That's really huge here.

Another benefit is that it adds some infrastructure that will allow doing similar kinds of unit testing easy to add in for other features. This is also a future time saver and so important to have in place.

So thanks to @adrifoster for getting this in for the FATES team.

I will confess that I only skimmed through the code, and didn't see anything obvious to point out. But, I'm grateful that @billsacks did an extensive careful review of the details. Also with the unit-testing I went through the unit-tests and thought if there's anything missing in the testing, and I didn't come up with anything obvious. I also have a lot of confidence here because I know the careful testing @adrifoster has been doing AND the unit testing itself gives a TON of confidence that's done correctly. Without that testing it would be harder to say it's OK.

Everything I'm saying here are just the reasons that unit-testing and, just better testing all around, is an industry standard best practice in the SE industry. It's something we need to shift to doing more of.

biogeochem/FatesPatchMod.F90
```diff
@@ -943,6 +949,279 @@ subroutine FreeMemory(this, regeneration_model, numpft)
       
     end subroutine FreeMemory
 
+    !===========================================================================
+    
+    subroutine InsertCohort(this, cohort)
```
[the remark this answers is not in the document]
adrifoster: thanks!

biogeochem/FatesPatchMod.F90
```diff
@@ -943,6 +949,279 @@ subroutine FreeMemory(this, regeneration_model, numpft)
...
+      type(fates_cohort_type), pointer :: nextCohort
+      type(fates_cohort_type), pointer :: sorted_head
+      type(fates_cohort_type), pointer :: currentSorted
+      
+      if (.not. associated(this%shortest)) then 
+        if (.not. associated(this%tallest)) then 
+          ! empty list
+          return
+        else
+          write(fates_log(),*) "error: one of shortest or tallest is null."
```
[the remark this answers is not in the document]
adrifoster: oh yes thank you! I thought I had updated this as well.

biogeochem/FatesPatchMod.F90
```diff
@@ -943,6 +949,279 @@ subroutine FreeMemory(this, regeneration_model, numpft)
...
+        currentCohort => currentCohort%taller
+      end do
+      
+      this%num_cohorts = cohort_count
+          
+    end subroutine CountCohorts
+    
+    !===========================================================================
+    
+    subroutine SortCohorts(this)
```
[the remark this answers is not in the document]
adrifoster: I think originally I was thinking of making this a merge sort which is more efficient but after realizing we don't really have super huge cohorts lists that often and that almost all the time the list will be nearly sorted already, I went back to insertion sort. 

I agree it would be best to just use the InsertCohort we already use. I will update this!

main/EDMainMod.F90
```diff
@@ -151,6 +149,8 @@ subroutine ed_ecosystem_dynamics(currentSite, bc_in, bc_out)
     !
     ! !LOCAL VARIABLES:
     type(fates_patch_type), pointer :: currentPatch
+    type(fates_cohort_type), pointer :: cohort
+    integer :: i
```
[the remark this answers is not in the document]
adrifoster: whoops thanks, leftover from some debugging

testing/functional_testing/patch/patch_test.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+import xarray as xr
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+from utils import round_up, get_color_palette, blank_plot
+from functional_class import FunctionalTest
+
+
+class PatchTest(FunctionalTest):
```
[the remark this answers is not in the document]
adrifoster: yes, WIP somewhat, it's also just part of the way I set up the functional testing. you have to create a class for the test to run. But right now it doesn't really do much other than demonstrate that you can pull up synthetic patch.

testing/testing_shr/SyntheticPatchTypes.F90
```diff
@@ -0,0 +1,235 @@
...
+    !
+    
+    ! ARGUMENTS:
+    class(synthetic_patch_array_type), intent(in)            :: this       ! array of patch data
+    integer,                           intent(in), optional  :: patch_id   ! desired patch id
+    character(len=*),                  intent(in), optional  :: patch_name ! desired patch name
+        
+    ! LOCALS:
+    integer :: i ! looping index
+
```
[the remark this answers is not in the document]
adrifoster: good catch - thanks will do!

testing/unit_testing/count_cohorts_test/test_CountCohorts.pf
adrifoster: oh interesting. I had thought about making some `final` subroutines for the patch when I did that refactor. I think it is a good idea. I have also never used that feature. There are some other things we deallocate on the production-code patch (arrays, other classes) as well. Perhaps this should be another PR?

testing/unit_testing/count_cohorts_test/test_CountCohorts.pf
adrifoster: oh okay I like this idea for the other tests as well! I have a lot of tests where a bulk of the test is just creating a hard-coded patch.

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
adrifoster: 100 probably is overkill haha... it is pretty fast?

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
adrifoster: So I want to set it up manually to avoid using other production methods so that we know we are only testing one method. But perhaps that is not necessary?

But I like your idea! that is easier to look at and debug.

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
adrifoster: fair point!

adrifoster:
@billsacks and @ekluzek  thank you so much for the review! I appreciate all the comments and suggestions.  Hoping to get to those today and then will re-run testing.

testing/unit_testing/count_cohorts_test/test_CountCohorts.pf
billsacks: Yes, deferring this to later work makes sense.

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
billsacks: I'll add a little more of my thinking here, particularly since @ekluzek and I seem to disagree here: To me, a big purpose of unit tests is their communication to other developers (including your future self) about the important requirements of the function-under-test, and what tests are needed to ensure that these requirements are satisfied. When I saw this test with size 100, it made me wonder, "Is there something interesting going on with the system-under-test that makes it important to have a test with this many of the same-height cohorts?" So there's some communication value in not having unnecessary tests, given your knowledge of the code / algorithm. Also, tests tend to require maintenance, so having unnecessary tests can mean more maintenance.

That said, these are really all minor points in my mind. I wanted to explain a little more the thoughts in the back of my head that led me to make this comment, but if those points don't speak to you, I think it would be totally reasonable to leave this test in place.

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
billsacks: > So I want to set it up manually to avoid using other production methods so that we know we are only testing one method. But perhaps that is not necessary?

Ah, okay. That does make sense. A quick comment explaining that rationale would help.

That said, my thinking would be: I agree that I'd be wary of a situation where all tests were exercising a combination of routines, such that there could be compensating errors in the different routines that led the test to pass. But in this case, it seems like you have good independent test coverage of InsertCohort, so I think it's reasonable to say: we have verified that InsertCohort has worked via this set of tests; so it's safe to use InsertCohort in this other set of tests. (And if, in setting this up, you find yourself realizing that, if a particular usage of InsertCohort might be broken in a way that would interfere with this testing, then that's good guidance on additional testing of InsertCohort that may be needed.) I guess this is at least somewhat (though not exactly) related to the ["solitary vs. sociable"](https://martinfowler.com/bliki/UnitTest.html) debate in unit testing; I come down pretty strongly on the "sociable" side myself, which ties in with my comfort in having a test exercise multiple routines. But others could argue differently, so this is another situation where I'd say: if my thoughts don't resonate with you, then I feel it's perfectly justified to keep this as you have it.

testing/unit_testing/validate_cohorts_test/test_ValidateCohorts.pf
billsacks: > oh weird... yes will update that

And if this test currently passes without that `assertExceptionRaised`, then it's probably worth figuring out why (because it means that other tests may be passing when they shouldn't).

adrifoster: actually I do see this here now. I think perhaps I had added that while you were in the process of reviewing? sorry about that.

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
adrifoster: Oh yeah that's true. I have done that where I created a list with Insert and then moved the heights around, but I am worried that might make it too confusing. Plus, because we don't necessarily run the insertcohort and sortcohorts tests at the same time (we don't necessarily run any of these...) I am leaning more towards the 'solitary' test methods in this case. 

But yes I agree, I will make a test helper.

testing/unit_testing/validate_cohorts_test/test_ValidateCohorts.pf
billsacks: Ah yes, I see it now - I think you're right about what happened (and now GitHub moved this comment to the wrong place in the file, too). No worries.

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
billsacks: > Oh yeah that's true. I have done that where I created a list with Insert and then moved the heights around, but I am worried that might make it too confusing. Plus, because we don't necessarily run the insertcohort and sortcohorts tests at the same time (we don't necessarily run any of these...) I am leaning more towards the 'solitary' test methods in this case.
> 
> But yes I agree, I will make a test helper.

Sounds good - I'm happy with your rationale here.

testing/unit_testing/count_cohorts_test/test_CountCohorts.pf
adrifoster: no... I think I just put that there out of habit haha (the production-code patches actually ARE pointers). thanks for catching that!

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
adrifoster: Thanks for your thoughts, @billsacks. That makes sense to me. Given that I already have a test that inserts a cohort into a list with a cohort of identical heights - is this test redundant?

I do remember that in my testing having the "identical" heights be at the beginning or end did catch other errors, which I think is caught by this test and not the other?

So could I update this to just, say 8 cohorts, which is the same length as the other "large" list tests?

testing/unit_testing/sort_cohorts_test/test_SortCohorts.pf
adrifoster: I actually assumed the `tolerance` argument was required haha. so that is good to know that it isn't!

testing/unit_testing/insert_cohort_test/test_InsertCohort.pf
billsacks: If your experience is that a test like this has caught other errors, then I'd support your idea of just changing this to be smaller.

adrifoster:
@billsacks thanks I just finished bringing in all your requested changes! I am going to kick off regression testing

billsacks (approved):
I have looked through the changes addressing my more substantive comments and am happy with these changes. This looks great to me now! Once again, thank you for all of this valuable work supporting more unit testing in FATES!

adrifoster:
fates testing on derecho all passes as expected (b4b against fixed branch):

`/glade/derecho/scratch/afoster/tests_0131-100214de`

fates testing on izumi running...

adrifoster:
okay finally able to check izumi and it also passes!

PR #1327, opened 2025-02-06
preserve cohort order when cohorts have identical heights

### LINKED ISSUE #1313: FATES insert_cohort subroutine does not preserve order of cohorts

adrifoster:
Fixes #1313 

### Description:

Just makes the cohort order preserved when the cohorts have identical heights

### Collaborators:

@billsacks @glemieux 

Yes - round off diffs

### Checklist

*Contributor*
- [X] The in-code documentation has been updated with descriptive comments

*Integrator*

### Test Results:

*CTSM (or) E3SM (specify which) test hash-tag:* `ctsm5.3.021`

*CTSM (or) E3SM (specify which) baseline hash-tag:* `ctsm5.3.021`

*FATES baseline hash-tag:* `sci.1.80.11_api.37.0.0`

*Test Output:*

### DIFF
--- a/biogeochem/EDCohortDynamicsMod.F90
+++ b/biogeochem/EDCohortDynamicsMod.F90
@@ -1309,10 +1309,10 @@ end subroutine create_cohort
     shortestc => NULL()
     storebigcohort   => null()
     storesmallcohort => null()
-    current_c => current_patch%tallest
+    current_c => current_patch%shortest
 
     do while (associated(current_c))
-       next_c => current_c%shorter
+       next_c => current_c%taller
        tallestc  => storebigcohort
        shortestc => storesmallcohort
        if (associated(tallestc)) then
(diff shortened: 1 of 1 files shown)


glemieux (approved):

adrifoster:
all tests pass :)

PR #1339, opened 2025-02-24
bugfix on harvesting of unoccupied canopy area

ckoven:
FATES is currently underestimating harvest rates relative to the forcing dataset, because it isn't transferring the correct amount of unoccupied space between tree crowns to the newly-logged patch when it applies harvest.  This is because of a bug in some of the logic handling the edge case for when the forcing file specifies too-tiny logging rates, which is mistakenly zeroing out the rate of logging that is applied to the interstitial area. This PR fixes that.

### Description:
There is logic to handle the case when the resulting secondary forest area would be smaller than the minimum FATES patch size. There is a mistaken `else` in that logic that shouldn't be there. 

This will change answers when logging is active.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -411,8 +411,6 @@ contains
              else if (currentPatch%land_use_label .eq. primaryland .and. .not. &
                   site_in%landuse_vector_gt_min(secondaryland)) then
                 harvest_rate = state_vector(secondaryland) / sum(state_vector(:))
-             else
-                harvest_rate = 0._r8
              end if
           else
              call GetInitLanduseHarvestRate(bc_in, site_in%min_allowed_landuse_fraction, &


glemieux (approved):
I think this makes sense.  Writing this out to make sense of it: the `site_in%landuse_vector_gt_min` is only going to be `false` if the state vector for the given landuse is less than or equal to `site_in%min_allowed_landuse_fraction`.  So the only way in which neither of these is true is if `currentPatch%land_use_label` is not primaryland and the secondaryland state vector value is greater than the minimum allowed value.  

If that is the case, coming out of this block, `harvest_rate` should be set to whatever  `get_harvest_rate_area` (or `get_harvest_rate_carbon`) set as the rate

glemieux:
Regression testing against `fates-sci.1.82.1_api.38.0.0-ctsm5.3.30` is showing B4B results with the exception of `FatesColdLUH2HarvestArea` which is as expected.

Issue #303, opened 2017-11-21
Need to change default value of fates_comp_excln


ckoven:
So I've been playing around with some runs where I started from BCI inventory (using datasets described in #297), using the full FATES (i.e. not prescribed physiology mode).  A thing that happens if I do this run with default parameter files is that the initial set of trees all die out in the first few years, and then subsequently get replaced by recruits that grow up and fill in the forest.  What this looks like, if I plot basal area per size bin over time, is the following:

![control_case](https://user-images.githubusercontent.com/10852790/33087007-3b03294c-ce9e-11e7-8a44-7d712a9a4b6c.png)

I did some experiments to try to figure out what was going on.  A clue is that canopy mortality is never high, which means that that initial set of inventory-initted trees is dying because they are in the understory, even though they are large.  But that doesn't make sense, as they ought to stay in the canopy, since they are large.  So I tried playing around with the fates_comp_excln parameter to see if that fixed things. For those who haven't played with this, what it does is control the probability of being in the canopy versus the understory as controlled by DBH (it should be based on height of course, but that's a separate issue, #302).  In code, what this looks like for the case of promotion is: `currentCohort%prom_weight = currentCohort%dbh**ED_val_comp_excln`. First I set the parameter to a value of -1, which triggers the deterministic PPA algorithm instead of the stochastic PPA algorithm for sorting.  That changed the behavior completely, so that it now looked like this:

![determ_ppa_case](https://user-images.githubusercontent.com/10852790/33087025-4376028e-ce9e-11e7-8b36-9aa614156d8a.png)

So from that I concluded that what's going on in the control case is that the number of tiny trees is so large, that they are actually ending up in the canopy relative to the large trees that are already there, which then kills off the big trees.  So obviously this wouldn't make sense that tiny trees are outcompeting the really big trees for light.  But I wanted to see if the stochastic PPA always produced this behavior, so I tried to intermediate cases, with fates_comp_excln set to 1.0 and 3.0.

The fates_comp_excln = 1.0 case looks like this:

![stoch_ppa_cmplexcl_1 0](https://user-images.githubusercontent.com/10852790/33087045-53b1d240-ce9e-11e7-82a9-11ba9684c9e4.png)

And the fates_comp_excln = 3.0 case looks like this:

![stoch_ppa_cmplexcl_3 0](https://user-images.githubusercontent.com/10852790/33087061-5988324a-ce9e-11e7-83e2-a7c3e3c41ba8.png)

[...]

PR #1315, opened 2025-01-16
increase ncl max from 2 to 3

rgknox:
### Description:

This change allows simulations to carry a third canopy layer.

### Collaborators:

many

### Expectation of Answer Changes:

Yes there should be differences in the answers

### Checklist

*If this is your first time contributing, please read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/main/CONTRIBUTING.md) document.*

All checklist items must be checked to enable merging this pull request:

*Contributor*
- [x] The in-code documentation has been updated with descriptive comments
- [x] The documentation has been assessed to determine if updates are necessary

*Integrator*
- [ ] FATES PASS/FAIL regression tests were run
- [ ] Evaluation of test results for answer changes was performed and results provided

### Documentation

- [Technical Note](https://github.com/NGEET/fates-docs) update:
- [User's Guide](https://github.com/NGEET/fates-users-guide) update: 

### Test Results:

*CTSM (or) E3SM (specify which) test hash-tag:*

*CTSM (or) E3SM (specify which) baseline hash-tag:*

*FATES baseline hash-tag:*

*Test Output:*

PR #1358, opened 2025-03-13
change comp_excln parameter to -1 (rank-ordered, deterministic sorting)

ckoven:
This changes the default value of the `comp_excln` parameter to -1. This leads to the use of the rank-ordered sorting scheme, rather than the current-default probabilistic scheme.  The main reason for doing this is that it results in fewer cohorts and therefore makes the model faster, which may be particularly important when increasing the number of canopy layers in #1315. It also has the effect of tending to slightly favor early-successional PFTs, as shown in fig. 13 of Koven et al, 2020, though that is not something that will show up in the kind of global nocomp and SP-mode simulations that this parameter file is currently mainly being aimed at.

This will change answers in all cases other than SP.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/parameter_files/fates_params_default.cdl
+++ b/parameter_files/fates_params_default.cdl
@@ -1732,7 +1732,7 @@ data:
 
  fates_cohort_size_fusion_tol = 0.08 ;
 
- fates_comp_excln = 3 ;
+ fates_comp_excln = -1 ;
 
  fates_damage_canopy_layer_code = 1 ;
 


rosiealice (approved):
I think this is a good idea, as we noted, most of our implementations at the moment do not use the competition feature, leading to large duplication of cohorts betweem canopy layers, so this will likely provide non-trivial efficiencies....

ckoven:
> @ckoven should this close #303?

That is so old I don't really remember the nature of the problem, but probably?

glemieux:
See https://github.com/NGEET/fates/pull/1406#issuecomment-2873372440 for discussion of test results.

glemieux:
See https://github.com/NGEET/fates/pull/1406#issuecomment-2873372440 for discussion of test results.  This PR will be automerged when #1406 will be integrated into `main`.

PR #1357, opened 2025-03-12
SPITFIRE refactor of mortality

adrifoster:
Refactors the SPITFIRE code for tree mortality.

### Description:
 Should be bit-for-bit. Just reorganizes the subroutines and creates some helper functions and unit tests

### Expectation of Answer Changes:
Should be bit 4 bit

### Checklist
All checklist items must be checked to enable merging this pull request:

*Contributor*
- [x] The in-code documentation has been updated with descriptive comments
- [x] The documentation has been assessed to determine if updates are necessary

*Integrator*
- [ ] FATES PASS/FAIL regression tests were run
- [ ] Evaluation of test results for answer changes was performed and results provided

### Test Results:
Ran preliminary tests that were b4b- will run again

*CTSM (or) E3SM (specify which) test hash-tag:* `ctsm5.3.029`

*CTSM (or) E3SM (specify which) baseline hash-tag:* `ctsm5.3.029`

*FATES baseline hash-tag:* `sci.1.81.1_api.38.0.0`

*Test Output:*

PR #1360, opened 2025-03-13
Prescribed fire implementation

XiulinGao:
### Description:
This PR implements prescribed fire [management fire for fuel reduction] into FATES by :
1) adding a subroutine to check weather condition against user defined temperature, relative humidity, and wind speed ranges to  see if there is a burn window 
2) if burn window presents, it then checks patch-level fuel load against user defined fuel load range to see whether a prescribed fire should happen for fuel reduction (with a low fuel load there is no need of using prescribed fire to further reduce fuel) or whether it's safe to burn (when a high fuel load presents it can be risky to conduct prescribed fire without removing extra fuels first, e.g. through mechanical thinning)
3) once fuel load condition is met, it will use SPITFIRE-calculated fire intensity to determine whether a wildfire or prescribed fire will occur by comparing current patch FI to wildfire FI threshold and to user-defined prescribed fire FI range. 
4) Fianlly, it will calculate the burned area for prescribed fire when it happens. Currently, the prescribed fire burned area is a user defined value [fraction per day] to reflect daily burning capacity due to factors such as resource constraints. 

### Collaborators:
@ckoven @adrifoster @samsrabin @glemieux @lmkueppers and Jessica Katz

Expect anwser change when prescribed fire is turned on. 

### Checklist

*Contributor*

*Integrator*
- [x] Multi-PR merge: Check only if PR is not part of a multi-PR merge or if the super PR is ready to merge

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


samsrabin:
It's great to see this PR! Please ping me again for review once it's ready.

rosiealice:
So this is prescribed fire in the sense that land managers set fire to things in a deliberate attempt to reduce fuel loads, as opposed to prescribed fire in the sense that we have a dataset that tells the model how much area should be burned on a given day? 

Either way this is cool...

XiulinGao:
@samsrabin Thanks Sam! It's ready for review now.

XiulinGao:
@rosiealice It's the management fire for fuel reduction purpose. I updated the PR description so there won't be any confusion between the management fire and prescribed fire burned area. Although, we also set a user parameter to prescribe area burned here.

rosiealice:
Cool. Does that mean that we can force the burned area to be a certain extent and override the ROS calculations?

XiulinGao:
> Cool. Does that mean that we can force the burned area to be a certain extent and override the ROS calculations?

yes. It currently has its own subroutine to calculate patch level burned area given the parameter, which defines the cap for site leve total burned fraction. But it still relies on SPITFIRE to calculate fuel condition, fire intensity, and post-fire mortality. I kind ignored the ROS for prescribed fire, not sure if we want to use SPITFIRE derived ROS as a proxy for prescribed fire rate of spread. It might still make sense. Usually for prescribed fire there's just a lot of human control on how much area it will burn, but the ROS should not be influenced, and should depend on fuel and weather condition unleass there's more fuels/energy been manually added to the system, which sometimes can happen during a prescribed fire.

glemieux:
This ideally should be integrated prior to #1357 per @adrifoster: https://github.com/NGEET/fates/pull/1357#issuecomment-2730147745

samsrabin (changes requested):
Hi Xiulin, thanks again for this cool new functionality. My main comment is that variables should be reworked so that the non-prescribed-fire versions are labeled as such, with only their combination being tracked in generic `fire` variables.

XiulinGao:
@samsrabin thanks Sam for your suggestions! I made corresponding changes. The main change is that we now track wildfire, prescribed fire, and the sum of the two with self-explainining variable name. I'm not sure if we need variables to represent the sum of both wildfire and prescribed fire tho, since we can easily get that via post-processing. But some of those will make calculation of other history variables easier without to sum wildfire and prescribed fire repetitively.

fire/SFFireWeatherMod.F90
```diff
@@ -67,4 +69,45 @@ subroutine UpdateEffectiveWindSpeed(this, wind_speed, tree_fraction, grass_fract
...
+    real(r8)             :: ws_check !intermediate value derived from wind speed condition check
+
+    if(.not. rxfire_switch) return   
+
+    t_check   = (temp_C - temp_low)*(temp_C - temp_up)
+    rh_check  = (rh - rh_low)*(rh - rh_up)
+    ws_check  = (wind - wind_low)*(wind - wind_up)
+
+    if(t_check .le. 0.0_r8 .and. rh_check .le. 0.0_r8 .and. &
+    ws_check .le. 0.0_r8)then
```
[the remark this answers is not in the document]
XiulinGao: > This is a clever way to test that values are within a range, but I wonder if it's too clever. It would be clearer for others reading the code to use conditionals instead. Would be interesting to hear if others agree, though—it didn't take me long to figure out.

@samsrabin I simply just added in-code text to explain what it does without making any code change. Hope that makes it clearer.

fire/SFMainMod.F90
```diff
@@ -669,8 +815,15 @@ subroutine  post_fire_mortality ( currentSite )
        if (currentPatch%fire == 1) then 
           currentCohort => currentPatch%tallest
           do while(associated(currentCohort))  
-             currentCohort%fire_mort = 0.0_r8
+             currentCohort%fire_mort = 0.0_r8
```
samsrabin, line 818: This whitespace change can be reverted.

fire/SFMainMod.F90
```diff
@@ -350,21 +368,32 @@ subroutine CalculateSurfaceFireIntensity(currentSite)
...
+    logical                         :: rx_hyb                          ! prescribed fire due to both lightning strike and human ignition
+    logical                         :: managed_wildfire                ! is it a wildfire with FI lower than the max rxfire intensity?[can either be Rx fire or wildfire]
+    logical                         :: true_wildfire                   ! is it a wildfire that cannot be managed?
+    logical                         :: is_wildfire                     ! combine both managed and true wildfire for now
     
     currentPatch => currentSite%oldest_patch 
     do while (associated(currentPatch))
       
-      currentPatch%fuel%frac_burnt(:) = 0.0_r8
+      currentPatch%fuel%frac_burnt(:) = 0.0_r8
```
samsrabin, line 396: This whitespace change can be reverted.

fire/SFMainMod.F90
samsrabin, line 818: This whitespace change can be reverted.

main/FatesRestartInterfaceMod.F90
```diff
@@ -1568,15 +1724,15 @@ subroutine define_restart_vars(this, initialize_variables)
...
 
-   call this%DefineRMeanRestartVar(vname='fates_disturbance_rates',vtype=cohort_r8, &
-        long_name='disturbance rates by donor land-use type, receiver land-use type, and disturbance type', &
-        units='1/day', initialize=initialize_variables,ivar=ivar, index = ir_disturbance_rates_siluludi)
+    call this%DefineRMeanRestartVar(vname='fates_disturbance_rates',vtype=cohort_r8, &
+         long_name='disturbance rates by donor land-use type, receiver land-use type, and disturbance type', &
+         units='1/day', initialize=initialize_variables,ivar=ivar, index = ir_disturbance_rates_siluludi)
 
-   if ( hlm_regeneration_model == TRS_regeneration ) then
+    if ( hlm_regeneration_model == TRS_regeneration ) then
```
samsrabin, line 1735: Whitespace changes here and above can be reverted (and should, to preserve helpful git blame history).

main/FatesRestartInterfaceMod.F90
```diff
@@ -1597,8 +1753,8 @@ subroutine define_restart_vars(this, initialize_variables)
       call this%DefineRMeanRestartVar(vname='fates_sdlng_mdd',vtype=cohort_r8, &
            long_name='seedling moisture deficit days', &
            units='mm days', initialize=initialize_variables,ivar=ivar, index = ir_sdlng_mdd_pa)
-
-   end if
+          
+    end if
```
samsrabin, line 1757: More whitespace changes that can be reverted.

biogeochem/FatesPatchMod.F90
```diff
@@ -211,9 +211,19 @@ module FatesPatchMod
     real(r8)              :: ros_back                ! rate of backward spread of fire [m/min]
     real(r8)              :: tau_l                   ! duration of lethal heating [min]
     real(r8)              :: fi                      ! average fire intensity of flaming front [kJ/m/s] or [kW/m]
```
adrifoster, line 213: so why do we have FI, nonrx_fi, and rx_fi? I'm not sure what FI is now.

fire/SFMainMod.F90
```diff
@@ -376,23 +405,73 @@ subroutine CalculateSurfaceFireIntensity(currentSite)
...
+          currentPatch%FI < SF_val_rxfire_maxthreshold .and. &
+          currentSite%NF > 0.0_r8)
+
+          is_rxfire = (rx_man .or. rx_hyb)
+
+          managed_wildfire = (currentSite%NF > 0.0_r8 .and. &
+          currentPatch%FI > SF_val_fire_threshold .and. &
+          currentPatch%FI < SF_val_rxfire_maxthreshold)
+
+          true_wildfire = (currentSite%NF > 0.0_r8 .and. &
```
adrifoster, line 439: This is a bit complex so I moved it into a new module (see my PR to your branch). Though I am a bit confused by the general logic.

So if we go into this rx_flag == .true. and also pass the fuel loading check, are we guaranteed to either be is_wildfire or is_rxfire? 

I just ask because we are checking a lot of things and some of it feels redundant.

fire/SFMainMod.F90
```diff
@@ -376,23 +405,73 @@ subroutine CalculateSurfaceFireIntensity(currentSite)
...
+              currentPatch%nonrx_fire = 1 
+            end if
+
+          end if
+
+          if (currentPatch%nonrx_fire == itrue) then
             currentSite%NF_successful = currentSite%NF_successful + &
-              currentSite%NF * currentSite%FDI*currentPatch%area / area
+            currentSite%NF*currentSite%FDI*currentPatch%area/area
+            currentPatch%nonrx_FI = currentPatch%FI
```
adrifoster, line 467: I'm not sure what the point of this is, why do we still have a currentPatch%FI?

fire/SFMainMod.F90
```diff
@@ -460,6 +538,74 @@ subroutine CalculateAreaBurnt(currentSite)
...
+
+    ! update total burnable fraction
+    total_burnable_frac = currentSite%rxfire_area_fi / AREA
+   
+    currentPatch => currentSite%oldest_patch;
+
+    do while(associated(currentPatch))
+
+      if(currentPatch%nocomp_pft_label .ne. nocomp_bareground)then
+        currentPatch%fire = 0           ! fire, either rx or non-rx
```
adrifoster, line 574: I had some clean-up to this section suggested in my PR.

adrifoster (changes requested):
This looks nice! I have some suggestions for cleaning up and adding modularity in my PR to your branch. Let me know what you think.

fire/SFMainMod.F90
XiulinGao: @adrifoster in my previous version I had another else statement where there is no fire at all even all prescribed fire condition checks passed. But since we are now zeroing out prescribed fire and wildfire patch variables at the beginning within the patch loop, I'd think adding that is redundant. We only care about the scenario where there's a fire, either wildfire or prescribed fire, and update the fire occurrence flag coordinately.

fire/SFMainMod.F90
XiulinGao: @adrifoster see my response above

fire/SFMainMod.F90
XiulinGao: @adrifoster thanks! Will merge that soon

fire/SFMainMod.F90
XiulinGao: @adrifoster I did feel checking both fuel load and FI can be redundant. I ended up keeping both because 
1) Fuel load can be used purely to control if we should do a prescribed fire (fuel load is too low or too high to burn) without having to check FI. This is useful in the case when we set prescribed fire FI lower threshold to be low, very likely indicating a low fuel load condition too; 2) while fuel load is an input for calculating FI, FI is not only influenced by fuel load, but also by fuel moisture content and other fuel characteristics. Thus, there can be case where we have relatively low fuel load but high FI if, for instance FMC is low or wind speed is high. Since FI is used in calculating post-fire mortality, the direct control on prescribed fire FI thus allow us to control prescribed fire mortality, which is important for fuel management that in most case we only want to remove small trees without affecting large trees.

rgknox:
Could include trivial changes noted here: https://github.com/NGEET/fates/issues/1418
with this PR  , @glemieux ?

glemieux:
@XiulinGao I've brought this branch up to date with `sci.1.84.1_api.40.0.0` and fixed the merge conflicts that were mainly due to the refactor and age-class updates in the history module that @samsrabin brought in with [`sci.1.83.1_api.39.0.0`](https://github.com/NGEET/fates/releases/tag/).  This resulted in the following changes:

- Moving the `site` variable zeroing to `reset_history_dyn_subsite`
- Moving the `rx`/`norx` related  `_si_age` variables and calculations to `update_history_dyn_subsite_ageclass`

I also updated the prescribed history variables that have `BURNFRAC` in the name to be consistent with the name change (e.g. `hio_nonrx_area_si -> hio_nonrx_fracarea_si`

Would you take a look at bf2e0c33f11f05fc890801e10334f5d8d18b06c4 to validate the changes, please?  @samsrabin would you take a look as well?

XiulinGao:
@glemieux Thank you Greg! It looks good to me.

fire/FatesRxFireMod.F90
```diff
@@ -0,0 +1,84 @@
...
+    has_ignitions = wildfire_ignitions > nearzero
+    above_wildfire_thresh = wildfire_FI > wildfire_FI_thresh
+  
+    managed_wildfire = has_ignitions .and. above_wildfire_thresh .and.  &
+      wildfire_FI < rx_max_FI
+
+    true_wildfire = has_ignitions .and. above_wildfire_thresh .and.  &
+      wildfire_FI > rx_max_FI
+      
+    is_wildfire = managed_wildfire .or. true_wildfire
```
samsrabin: What is the purpose of the distinction between "managed wildfire" and "true wildfire"? Is that ever used anywhere?

fire/SFMainMod.F90
```diff
@@ -376,21 +398,57 @@ subroutine CalculateSurfaceFireIntensity(currentSite)
...
+              currentSite%rxfire_area_fi = currentSite%rxfire_area_fi + currentPatch%area ! record burnable area after FI check
+              currentPatch%rx_fire = 1
+              
+            else if (is_wildfire) then
+              currentPatch%nonrx_fire = 1
+            end if
+              
+          else  ! not a patch suitable for conducting prescribed fire or rxfire is not even turned on
+            ! track wildfires greater than kW/m energy threshold
+            if (currentPatch%FI > SF_val_fire_threshold) then
```
samsrabin: Why is `is_wild_fire()` not used here?

parameter_files/fates_params_default.cdl
```diff
@@ -894,6 +894,45 @@ variables:
...
+	    fates_rxfire_wind_upthreshold:units = "m/s";
+		fates_rxfire_wind_upthreshold:long_name= "maximum wind speed threshold for conducting prescribed fire";
+	double fates_rxfire_wind_lwthreshold ;
+	    fates_rxfire_wind_lwthreshold:units = "m/s";
+		fates_rxfire_wind_lwthreshold:long_name= "minimum wind speed threshold for conducting prescribed fire";
+	double fates_rxfire_AB ;
+	    fates_rxfire_AB:units = "fraction/day";
+		fates_rxfire_AB:long_name= "daily burn capacity of prescribed fire";
+	double fates_rxfire_min_threshold ;
+	    fates_rxfire_min_threshold:units = "kJ/m/s or kW/s";
```
samsrabin: kJ/m/s is kW/m, not kW/s.

parameter_files/fates_params_default.cdl
```diff
@@ -894,6 +894,45 @@ variables:
...
+		fates_rxfire_min_threshold:long_name= "minimum energy threshold for conducting prescribed fire";
+	double fates_rxfire_max_threshold ;
+	    fates_rxfire_max_threshold:units = "kJ/m/s or kW/s";
```
samsrabin: kJ/m/s is kW/m, not kW/s.

fire/SFParamsMod.F90
```diff
@@ -37,6 +37,19 @@ module SFParamsMod
    real(r8),protected, public :: SF_val_low_moisture_Slope(num_fuel_classes)
    real(r8),protected, public :: SF_val_mid_moisture_Coeff(num_fuel_classes)
    real(r8),protected, public :: SF_val_mid_moisture_Slope(num_fuel_classes)
+    ! Prescribed fire relevant parameters
+   real(r8),protected, public :: SF_val_rxfire_tpup   ! temprature upper threshold for conducting RX fire
+   real(r8),protected, public :: SF_val_rxfire_tplw   ! temprature lower threshold
```
samsrabin: "temprature" typo.

parameter_files/fates_params_default.cdl
```diff
@@ -894,6 +894,45 @@ variables:
...
+		fates_rxfire_max_threshold:long_name= "maximum energy threshold for conducting prescribed fire";
+	double fates_rxfire_fuel_min ;
+	    fates_rxfire_fuel_min:units = "kgC/m2";
+		fates_rxfire_fuel_min:long_name= "minimum fuel load at the patch level for prescribed fire to occur";
+	double fates_rxfire_fuel_max ;
+	    fates_rxfire_fuel_max:units = "kgC/m2";
+		fates_rxfire_fuel_max:long_name= "maximum fuel load above which prescribed fire can be risky";
```
samsrabin: "Is disallowed" would be clearer than "can be risky." Also, it should be "at or above," not just "above."

fire/SFParamsMod.F90
```diff
@@ -267,6 +345,42 @@ subroutine SpitFireReceiveScalars(fates_params)
...
+	 data=SF_val_rxfire_wdup)
+
+    call fates_params%RetrieveParameter(name=SF_name_rxfire_wdlw, &
+	 data=SF_val_rxfire_wdlw)
+
+    call fates_params%RetrieveParameter(name=SF_name_rxfire_AB, &
+         data=SF_val_rxfire_AB)
+
+    call fates_params%RetrieveParameter(name=SF_name_rxfire_min_threshold, &
+         data=SF_val_rxfire_minthreshold)
```
samsrabin: `SF_val_rxfire_minthreshold` should be `SF_val_rxfire_min_threshold` to enable searching code for instances of this variable across its different names.

fire/SFParamsMod.F90
```diff
@@ -267,6 +345,42 @@ subroutine SpitFireReceiveScalars(fates_params)
...
+
+    call fates_params%RetrieveParameter(name=SF_name_rxfire_max_threshold, &
+         data=SF_val_rxfire_maxthreshold)
```
samsrabin: `SF_val_rxfire_maxthreshold` should be `SF_val_rxfire_max_threshold` to enable searching code for instances of this variable across its different names.

fire/SFParamsMod.F90
```diff
@@ -37,6 +37,19 @@ module SFParamsMod
...
+   real(r8),protected, public :: SF_val_rxfire_tplw   ! temprature lower threshold
+   real(r8),protected, public :: SF_val_rxfire_rhup   ! relative humidity upper threshold
+   real(r8),protected, public :: SF_val_rxfire_rhlw   ! relative humidity lower threshold
+   real(r8),protected, public :: SF_val_rxfire_wdup   ! wind speed upper threshold
+   real(r8),protected, public :: SF_val_rxfire_wdlw   ! wind speed lower threshold
+   real(r8),protected, public :: SF_val_rxfire_AB     ! prescribed fire burned fraction per day
+   real(r8),protected, public :: SF_val_rxfire_minthreshold ! minimum fire energy of rx fire, for management outcomes really
+   real(r8),protected, public :: SF_val_rxfire_maxthreshold ! maximum fire energy
+   real(r8),protected, public :: SF_val_rxfire_fuel_min     ! minimum fuel load at the patch for the need of rx fire management
+   real(r8),protected, public :: SF_val_rxfire_fuel_max     ! maximum fuel load, above which might be risky for conducting rx fire
```
samsrabin: "above which might be risky for conducting rx fire" should be "**at or** above which **prescribed fire is disallowed**."

fire/SFParamsMod.F90
```diff
@@ -37,6 +37,19 @@ module SFParamsMod
...
+    ! Prescribed fire relevant parameters
+   real(r8),protected, public :: SF_val_rxfire_tpup   ! temprature upper threshold for conducting RX fire
+   real(r8),protected, public :: SF_val_rxfire_tplw   ! temprature lower threshold
+   real(r8),protected, public :: SF_val_rxfire_rhup   ! relative humidity upper threshold
+   real(r8),protected, public :: SF_val_rxfire_rhlw   ! relative humidity lower threshold
+   real(r8),protected, public :: SF_val_rxfire_wdup   ! wind speed upper threshold
+   real(r8),protected, public :: SF_val_rxfire_wdlw   ! wind speed lower threshold
+   real(r8),protected, public :: SF_val_rxfire_AB     ! prescribed fire burned fraction per day
+   real(r8),protected, public :: SF_val_rxfire_minthreshold ! minimum fire energy of rx fire, for management outcomes really
+   real(r8),protected, public :: SF_val_rxfire_maxthreshold ! maximum fire energy
```
samsrabin: All the comments for rx fire parameters here and above could be improved: Specify that prescribed fire is only allowed within range.

fire/SFMainMod.F90
XiulinGao: @samsrabin if you take a look at the is_wild_fire(), it's not only based on ignition and energy threshold check, but also that if current wildfire energy is above or less than the max. threshold for prescribed fire. This is_wild_fire() is really just for case where both prescribed fire and wildfire can happen, which fire we'd like to define it. But here we just consider the case where Rx fire is impossible, do we get a wildfire (so independent of rx fire). This is relevant to your previous comment about the difference between "true wildfire" and "managed wildfire" defined in is_wild_fire(). The reason to separate them is that in real world, sometimes managers will also use a mild-moderate wildfire to achieve management goals without suppressing them but rather let them burn under control (which we call "managed wildfire"). So i think it can be useful to leave the possibility here in case in the future we want to classify the "managed wildfire" to be prescribed fire (or beneficial fire in general to be more inclusive) too. It's not used anywhere yet. So if you think we should remove it I'll just remove it.

fire/SFParamsMod.F90
XiulinGao: @samsrabin I totally agree to change to "is disallowed". But still should be "above" not "at or above". In burn window check, when any of the wheather variable is at the boundary conditon, burn window presents, so "above" holds true here unless we change the burn window definition.

XiulinGao: > @samsrabin I totally agree to change to "is disallowed". But still should be "above" not "at or above". In burn window check, when any of the wheather variable is at the boundary conditon, burn window presents, so "above" holds true here unless we change the burn window definition.

Oh never mind! I mis-read this! thats correct, for fuel and energy threshould should be "at or above"! I thought it's about the weather condition check.

fire/SFParamsMod.F90
XiulinGao: @samsrabin Thanks Sam! I have updated the name and parameter description per your comment. But let me know if I also should combine the true_wildfire and managed_wildfire into just wildfire.

fire/SFMainMod.F90
samsrabin: Yeah, I think that the fact that the distinction isn't actually used makes it more confusing than handy to keep around. I think it should changed to remove the unused distinction, which allows substantial simplification: see the PR I just submitted on your repo, XiulinGao/fates#4.

samsrabin (changes requested):
Thanks, Xiulin. I think `is_wild_fire()` should be deleted to keep things simple; see the PR I just submitted on your repo, XiulinGao/fates#4.

XiulinGao:
@samsrabin Thanks Sam! I've merged you PR.

fire/FatesRxFireMod.F90
samsrabin: Resolved with e14c0be.

fire/SFMainMod.F90
samsrabin: Resolved with e14c0be.

biogeochem/FatesPatchMod.F90
adrifoster: okay sounds good!

fire/SFMainMod.F90
adrifoster: okay sounds great

XiulinGao:
@samsrabin Just want to let you know that I added a ignition check for wildfire occurence in SFMain under the CalculateSurfaceFireIntensity subroutine. This is necessary as in 
line 410 it's either non-zero ignition or prescribed fire is on, so we can have a case where ignition is zero but prescribed fire is turned on, then wildfire will never happen even the current potential FI is > the defined threshold for wildfire.

samsrabin:
Oh yup, good catch!

glemieux:
Regression testing on `derecho` is underway.

fire/FatesRxFireMod.F90
```diff
@@ -0,0 +1,52 @@
...
+  ! ============================================================================
+
+  use FatesConstantsMod, only : r8 => fates_r8
+  use FatesConstantsMod, only : nearzero
+
+  implicit none
+  private
+  
+  public :: is_prescribed_burn
+  !public :: is_wildfire
```
glemieux: @XiulinGao during testing I ran into some build errors using the intel compiler on `derecho` one of which was that this `is_wildfire` statement doesn't have an associated procedure.  Is this older code that should be dropped?

XiulinGao: @glemieux yes, I think is_wildfire is deleted based on @samsrabin's earlier comment. So this procedure in FatesRxFireMod should be removed accordingly.

parameter_files/fates_params_default.cdl
```diff
@@ -894,6 +894,45 @@ variables:
 	double fates_q10_mr ;
 		fates_q10_mr:units = "unitless" ;
 		fates_q10_mr:long_name = "Q10 for maintenance respiration" ;
+	double fates_rxfire_switch ;
+	    fates_rxfire_switch:units = "unitless" ;
+		fates_rxfire_switch:long_name = "management fire mode, 1 = use management fire, 0 = turn off management fire" ;
```
glemieux: @XiulinGao I'm going to move this to the host land model as a namelist option since we've been migrating global switches there.  

That said, should how does prescribed fire interact with other existing spitfire modes?  I.e. should it be a new option for `fates_spitfire_mode` or should be a new mode entirely?

cc @adrifoster @samsrabin

XiulinGao: @glemieux good point. Prescribed fire does not rely on any ignition probability, so it does not require the different ignition modes as SPITFIRE does. But it depends on SPITFIRE to calculate fuel conditions, fire behavior, and post-fire mortality. Since fates_spitfire_mode 0 really just skips SPITFIRE routine, it thus also disables prescribed fire or causes error. So mode 0 is the only mode that cannot run when users want to use prescribed fire.  

It does not make sense to set prescribed fire as a new spitfire mode, as what we want is that prescribed fire can interact with SPITFIRE no matter what mode the user is using as long as it's not mode 0. So my final thought is that making a new mode, not fates_spitfire_mode but something like fates_rxfire_mode makes sense to me, and also a note to user that if fates_rxfire_mode is on (e.g. 1), fates_spitfire_mode cannot be 0. In cases where users want to turn off wildfire and only use prescribed fire, they should use fates_spitfire_mode 1 and set ignitions to 0 in parameter file. 

I'd like to see what @adrifoster and @samsrabin think

glemieux: @XiulinGao thanks, that makes sense to me.  During one of our software meetings today, it was noted that the term "prescribed" from a modelling perspective is somewhat overloaded as modeling term.  Is there an alternate term for the switch name that would work for you?

samsrabin: "Controlled" could work.

glemieux: I also realized that using "managed" as the current parameter file switch definition uses would work too.

samsrabin: Yeah, and then `mgd` isn't much longer as an abbreviation than `rx`.

XiulinGao: @glemieux @samsrabin both work fine for me. So whichever you all decide to use.

glemieux:
Regression testing is underway on `derecho`.  The queues have been taking a lot longer than usual (~12+ hours), so this is delaying the process.

glemieux:
Regression testing on `derecho` against `fates-sci.1.86.3_api.40.0.0-ctsm5.3.065` is returning B4B results with the exception of known `FIELDLIST` differences due to the new history output and the change of some older fire outputs that are changed to include `WILDFIRE` in the name.

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1360-fates-1` and `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1360-fates-2`

PR #1364, opened 2025-03-18
vegetation bin indexing

rgknox:
### Description:

This set of changes introduces some consistency improvements to how we determine the number of vegetation layers, and how we access and use those indices.  Of note, the lower bound of this array has been changed to 0.

### Collaborators:

@ckoven 

Not sure. This may be a bug fix as well.  Looking again.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/radiation/FatesRadiationDriveMod.F90
+++ b/radiation/FatesRadiationDriveMod.F90
@@ -13,6 +13,7 @@ module FatesRadiationDriveMod
   use EDTypesMod        , only : ed_site_type
   use FatesPatchMod,      only : fates_patch_type
   use EDParamsMod,        only : maxpft
+  use EDParamsMod       , only : GetNVegLayers
   use FatesConstantsMod , only : r8 => fates_r8
   use FatesConstantsMod , only : fates_unset_r8
   use FatesConstantsMod , only : itrue
@@ -438,11 +439,18 @@ contains
                            if_notair: if (ft>0) then
                               area_frac = twostr%scelg(cl,icol)%area
                               vai = twostr%scelg(cl,icol)%sai+twostr%scelg(cl,icol)%lai
-                              nv = minloc(dlower_vai, DIM=1, MASK=(dlower_vai>vai))
+
+                              nv = GetNVegLayers(vai)
+
                               do iv = 1, nv
                                  
-                                 vai_top = dlower_vai(iv)-dinc_vai(iv)
-                                 vai_bot = min(dlower_vai(iv),twostr%scelg(cl,icol)%sai+twostr%scelg(cl,icol)%lai)
+                                 vai_top = dlower_vai(iv)
+
+                                 if(iv == nv) then
+                                    vai_bot = twostr%scelg(cl,icol)%sai+twostr%scelg(cl,icol)%lai
+                                 else
+                                    vai_bot = dlower_vai(iv)+dinc_vai(iv)
+                                 end if
                                  
                                  cpatch%parprof_pft_dir_z(cl,ft,iv) = cpatch%parprof_pft_dir_z(cl,ft,iv) + &
                                       area_frac*twostr%GetRb(cl,icol,ivis,vai_top)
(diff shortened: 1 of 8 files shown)


radiation/FatesRadiationDriveMod.F90
```diff
@@ -438,11 +439,18 @@ subroutine FatesSunShadeFracs(nsites, sites,bc_in,bc_out)
                            if_notair: if (ft>0) then
                               area_frac = twostr%scelg(cl,icol)%area
                               vai = twostr%scelg(cl,icol)%sai+twostr%scelg(cl,icol)%lai
-                              nv = minloc(dlower_vai, DIM=1, MASK=(dlower_vai>vai))
+
+                              nv = GetNVegLayers(vai)
```
rgknox, line 443: This is a fix: Since the older version of dlower_vai had a non-zero first index, if vai is smaller than dlower_vai(1) but non-zero, then the mask on dlower_vai is all FALSE. So the result of minloc will be zero and there will be no looping here (which would be incorrect).

rgknox: The good part is that these are not prognostic, only diagnostics variables in this loop.

radiation/FatesRadiationDriveMod.F90
```diff
@@ -438,11 +439,18 @@ subroutine FatesSunShadeFracs(nsites, sites,bc_in,bc_out)
...
+
                               do iv = 1, nv
                                  
-                                 vai_top = dlower_vai(iv)-dinc_vai(iv)
-                                 vai_bot = min(dlower_vai(iv),twostr%scelg(cl,icol)%sai+twostr%scelg(cl,icol)%lai)
+                                 vai_top = dlower_vai(iv)
```
glemieux, line 447: Is it correct to say that `vai_top` has shifted now that the lower bound value is zero?

radiation/FatesRadiationDriveMod.F90
```diff
@@ -438,11 +439,18 @@ subroutine FatesSunShadeFracs(nsites, sites,bc_in,bc_out)
...
+
+                                 if(iv == nv) then
+                                    vai_bot = twostr%scelg(cl,icol)%sai+twostr%scelg(cl,icol)%lai
+                                 else
+                                    vai_bot = dlower_vai(iv)+dinc_vai(iv)
```
glemieux, line 452: Could this match the change in FatesAllometryMod line 3197: `vai_bot = dlower_vai(iv+1)`

glemieux:
Regression testing on `derecho` is underway.

radiation/FatesRadiationDriveMod.F90
rgknox: vai_top still has the same meaning, but dlower_vai has changed.  Lower used to refer to lower vertical position, now it refers to the numericaly lower edge of the bin.  So vai_top points directly to dlower_vai now since they conceptually align.

radiation/FatesRadiationDriveMod.F90
rgknox: yes, that's true!

glemieux:
Regression testing against `fates-sci.1.82.7_api.39.0.0-ctsm5.3.040` is complete.  This results in mostly targeted DIFFs across all testmods in the `fates_levleaf`:
```
 209  fates_levleaf   (fates_levleaf)
 210          30       30  (    30) (     1) (     1) 
 211                   30  0.290000000000000E+02  0.000000000000000E+00  0.100000000000000E+01       0  0.000000000000000E+00  0.000000000000000E+00
 212                   30  0.300000000000000E+02  0.100000000000000E+01                              1                         0.000000000000000E+00
 213                   30  (    30) (     1)
 214  RMS fates_levleaf                    1.0000E+00            NORMALIZED  6.6667E-02
````

As a reminder, the third line (here line number `211`) is the test branch output and the fourth line (`212`) are the baseline comparison values.  The first and second numbers in each line are the max and min values respectively.  This highlights that the lower bound is indeed zero as expected per this PR.

It appears that there are answer changes in some of the longer term tests only:
```
ERS_Ld397.f45_f45_mg37.I2000Clm50Fates.derecho_intel.clm-FatesColdNoComp
SMS_Lm13.1x1_brazil.I2000Clm50FatesCruRsGs.derecho_intel.clm-FatesCold
ERS_P128x1_Ld765.f10_f10_mg37.I2000Clm60Fates.derecho_intel.clm-FatesColdNoComp
```

PR #1371, opened 2025-03-27
Nocomp seed localization

ckoven:
### Description:

This code makes it so that, in nocomp configurations only, any given PFT's seeds will all end up in patches where that PFT can grow.  The current behavior is that seeds are dispersed around on all patches, but they will only recruit on patches with that PFT's nocomp logic, so that, as a result, a lot of the seeds just decay.  The problem with the current logic is that if you imagine a gridcell where there is a PFT that has some tiny area that it is allowed to grow on, then most of that PFTs seeds end up unable to grow. So at the limit, as the nocomp area gets smaller for a given PFT on a given gridcell, it basically guarantees local extinction of that PFT at some point. My hunch is that this is responsible for some weird dynamics I have been seeing under spinup with constant land use, where PFTs can go extinct in the primary land patches of gridcells that have a lot of cropland on them.  I still have testing in the queue to see if this code actually fixes that problem. But in any case, it just seems wrong that, at the margin, we would penalize the likelihood of a PFT to maintain a viable population based just on the fractional area that we assign to it. I guess one counterargument would be that seed dispersal and recruitment really does degrade at the range limit of a given species due to habitat discontinuity and lack of dispersers. So I've written this as a hard-coded flag in the code for now, but happy to change it to a namelist flag if others think the counterarguments are strong or this introduces any weird behaviors (still TBD), or alternately delete the flag entirely if others think that the current logic is just wrong.

Note that revisiting some broader discussions we've had in the past of within-gridcell seed dispersal across patches is probably warranted, given that the implied length-scale of patches is likely different for patches that differ only in their disturbance history, versus also in their nocomp PFT and/or land-use labels if those are enabled in a given configuration. This is just the simplest modification I could think of to address this, much more complex ones are also possible. E.g., we may also want to have some fractional seed localization between patches with different land use types on a given gricell, or always localize some fraction of seeds to the parent cohort's patch, etc.

### Collaborators:
Discussed with @rgknox.

This will change answers in nocomp configurations, so I suggest not merging it until we have a bit better sense of how it qualitatively changes things.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -2101,7 +2101,7 @@ contains
        ! only on patches that allow that PFT to grow, then we need to add up all the patch areas
        ! for each nocomp PFT to normalize the seed fluxes with later.
        if (nocomp_seed_localization .and. hlm_use_nocomp .eq. itrue ) then
-          nocomp_patch_areas(0:numpft) = 0.r8
+          nocomp_patch_areas(0:numpft) = 0._r8
           currentPatch => currentSite%oldest_patch
           nocomp_patch_loop: do while (associated(currentPatch))
              nocomp_patch_areas(currentPatch%nocomp_pft_label) = nocomp_patch_areas(currentPatch%nocomp_pft_label) &
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -2080,6 +2080,10 @@ contains
     integer  :: el                     ! loop counter for litter element types
     integer  :: element_id             ! element id consistent with parteh/PRTGenericMod.F90
 
+    logical, parameter  :: nocomp_seed_localization  = .true.  ! if nocomp is on, only send a given PFT's seeds to patches of that nocomp PFT
+    real(r8) :: nocomp_seed_scaling    ! scalar to handle case for nocomp_seed_localization
+    real(r8) :: nocomp_patch_areas(0:numpft) ! vector of the total patch areas for each nocomp PFT
+
     ! If the dispersal kernel is not turned on, keep the dispersal fraction at zero
     site_disp_frac(:) = 0._r8
     if (hlm_seeddisp_cadence .ne. fates_dispersal_cadence_none) then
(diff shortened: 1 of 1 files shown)


glemieux:
We have a hardcoded option called `homogenize_seed_rain`, should we add a check to avoid this option if `hlm_use_nocomp .eq. itrue`?

https://github.com/NGEET/fates/blob/d121971dd66b88e1dcb4020cbc213ed1f19af8c1/biogeochem/EDPhysiologyMod.F90#L2135-L2140

ckoven:
> We have a hardcoded option called `homogenize_seed_pfts`; Should we add a check to avoid this option if `nocomp_seed_localization .and. hlm_use_nocomp .eq. itrue`?

Honestly we should probably just delete the `homogenize_seed_pfts` option, it is a very niche use-case that I don't think anyone has used for a while (and simple enough to revert if someone really wants to).

biogeochem/EDPhysiologyMod.F90
```diff
@@ -2152,9 +2161,22 @@ subroutine SeedUpdate( currentSite )
...
+                   else
+                      nocomp_seed_scaling = 0._r8
+                   endif
+                endif
+
                 ! Seed input from local sources (within site).  Note that a fraction of the
                 ! internal seed rain is sent out to neighboring gridcells.
-                litt%seed_in_local(pft) = litt%seed_in_local(pft) + site_seed_rain(pft)*(1.0_r8-site_disp_frac(pft))/area ![kg/m2/day]
+                litt%seed_in_local(pft) = litt%seed_in_local(pft) + nocomp_seed_scaling * &
+                     site_seed_rain(pft)*(1.0_r8-site_disp_frac(pft))/area ![kg/m2/day]
```
glemieux: Note to self: push change to clarify `site_seed_rain/area`

glemieux: Done

glemieux:
@ckoven I made modifications to the external seed rain check to be consistent with `nocomp_seed_localization` via affa704.  Let me know if this is acceptable or if they external shouldn't conform to the localization switch.

glemieux (approved):

glemieux:
Regression testing is underway on derecho.

glemieux:
Regression testing on `derecho` against `fates-sci.1.85.2_api.40.0.0-ctsm5.3.059` is complete.  Only nocomp tests are showing differences as expected.  Differences are show up in the seed relevant history output and downstream variable calculations as well.

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1371-fates`

Issue #1354, opened 2025-03-10
ELM-FATES CNP mode `mass balance error` due to `sci.1.80.0_api.37.0.0`


glemieux:
### Describe the issue

During testing for E3SM-Project/e3sm#7027 it was discovered that the `fates_eca` and `fates_rd` tests are failing `RUN`.  Spot testing using the latest e3sm master branch commit and working forwards through the fates tags resulted in discovering that the changes associated with `sci.1.80.0_api.37.0.0` trigger the failure mode.  

### Relevant log output

```shell
61:  mass balance error detected
 61:  element type (see PRTGenericMod.F90):            1
 61:  error fraction relative to biomass stock:   1.118038134966836E-005
 61:  absolut error (flux in - change):   2.876421946683116E-002
 61:  call index:            1
 61:  Element index (PARTEH global):           1
 61:  net:    11.1017546883988
 61:  dstock:    11.0729904689319
 61:  seed_in:   0.000000000000000E+000
 61:  net_root_uptake:   0.000000000000000E+000
 61:  gpp_acc:    16.7568276755855
 61:  flux_generic_in:   0.000000000000000E+000
 61:  wood_product_harvest:   0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000
 61:  wood_product_landusechange:   0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000  0.000000000000000E+000
 61:   0.000000000000000E+000  0.000000000000000E+000
 61:  error from patch resizing:   0.000000000000000E+000
 61:  burn_flux_to_atm:   0.000000000000000E+000
 61:  seed_out:   0.000000000000000E+000
 61:  flux_generic_out:   0.000000000000000E+000
[...]

PR #1372, opened 2025-03-31
unit fix on excess respiration mass balance checking

rgknox:
### Description:

Excess respiration is activated when a plant is nutrient limited but continues to accumulate carbon.  When stores are full, the plant will respire excess carbon.  For book-keeping purposes, we make sure that carbon fluxes match the change in the carbon stocks.  This excess respiration term had a unit error when calculating these balance check fluxes. There was no error in how this term actually affected NPP, which drives plant allocations, just the checking process.  The unit error was fixed and descriptive text was updated.

Also, it was noticed that unit conversions of fluxes that are per year and per day, used a global variable "hlm_days_per_year". This is an integer, but was being used in multiply and divides on real numbers. Compilers should be converting this integer to a float for us, but it is better to pre-convert it to reduce warning messages and we shouldn't assume that the compilers will fix things for us.  It is better practice to specify type conversions ourselves, so I applied the conversions.

Fixes: #1354 

To verify that this fix I ran the following two ELM land tests:

>SMS_D_Ld20.f45_f45.IELMFATES.pm-cpu_gnu.elm-fates_rd
>SMS_Ld60.f45_f45.IELMFATES.pm-cpu_gnu.elm-fates_eca

Note that the second test is typically a 20 day test, but I wanted to run it longer to give it more time to trigger the error.  Both tests now pass.

### Collaborators:
@glemieux @ckoven 

This should fix failing tests in ELM.
It should not affect answers. If it does, that means the integer conversion was long overdue. :/

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -302,8 +302,8 @@ module FatesHistoryInterfaceMod
   integer :: ih_seed_bank_si            ! carbon only
   integer :: ih_seeds_in_si             ! carbon only
   integer :: ih_seeds_in_local_si       ! carbon only
-  integer :: ih_ungerm_seed_bank_si        ! carbon only
-  integer :: ih_seedling_pool_si    ! carbon only
+  integer :: ih_ungerm_seed_bank_si     ! carbon only
+  integer :: ih_seedling_pool_si        ! carbon only
   integer :: ih_ba_weighted_height_si
   integer :: ih_ca_weighted_height_si
   integer :: ih_seeds_in_local_elem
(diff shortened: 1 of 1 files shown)


glemieux (commented):
Do we need to update the calculation here in history as well?
https://github.com/NGEET/fates/blob/626f5e2950061ecf8ab063db8fafb85a741f23a5/main/FatesHistoryInterfaceMod.F90#L2839-L2841

rgknox:
tests pass on derecho

PR #1376, opened 2025-04-01
fix to trivial radiation transmission

rgknox:
### Description:

FATES tells the host model the fraction of transmitted radiation through the vegetation. This allows the host model to perform calculations on soil and vegetation heating. This transmission is defined by three terms, the 1) direct radiation flux at the canopy bottom, the 2) diffuse radiation flux scattered from a direct radiation source and 3) the diffuse radiation flux scattered from a diffuse source.  We assume trivial values for the case when there is no light. In the vast majority of cases, these trivial values are unused, yet some driver datasets (such as NIWO) have non-zero downwelling solar flux at negative zenith angles. In these rare cases, the host model needs these transmissions and canopy albedos, even if fates does not perform the solve.

This PR fixes a bug to that trivial flux assumption, by realizing that the total transmission from the first two sources (from direct radiation) cannot exceed 1 (that would be creating energy).

Fixes: https://github.com/ESCOMP/CTSM/issues/3038

### Collaborators:

@samsrabin detected the error and consulted

This should enable the NIWO NEON tests to run without FAIL, no meaningful changes to answers though

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/main/FatesInterfaceMod.F90
+++ b/main/FatesInterfaceMod.F90
@@ -373,13 +373,13 @@ contains
     fates%bc_out(s)%rssun_pa(:)     = 0.0_r8
     fates%bc_out(s)%rssha_pa(:)     = 0.0_r8
     
-    fates%bc_out(s)%albd_parb(:,:) = 0.0_r8   ! zero albedo, soil absorbs all rad
-    fates%bc_out(s)%albi_parb(:,:) = 0.0_r8   ! zero albedo, soil absorbs all rad
+    fates%bc_out(s)%albd_parb(:,:) = 1.0_r8   ! zero albedo, soil absorbs all rad
+    fates%bc_out(s)%albi_parb(:,:) = 1.0_r8   ! zero albedo, soil absorbs all rad
     fates%bc_out(s)%fabd_parb(:,:) = 0.0_r8   ! no rad absorbed by veg
     fates%bc_out(s)%fabi_parb(:,:) = 0.0_r8   ! no rad absorbed by veg
-    fates%bc_out(s)%ftdd_parb(:,:) = 1.0_r8   ! rad flux to soil at bottom of veg is 100%
+    fates%bc_out(s)%ftdd_parb(:,:) = 0.0_r8   ! rad flux to soil at bottom of veg is 100%
     fates%bc_out(s)%ftid_parb(:,:) = 0.0_r8   ! rad flux to soil at bottom of veg is 100%
-    fates%bc_out(s)%ftii_parb(:,:) = 1.0_r8   ! rad flux to soil at bottom of veg is 100%
+    fates%bc_out(s)%ftii_parb(:,:) = 0.0_r8   ! rad flux to soil at bottom of veg is 100%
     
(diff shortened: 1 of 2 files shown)


glemieux (approved):

glemieux:
FYI, I ran into a similar run failure to #3038 when testing e3sm-p[roject/e3sm#6918.  Applying this fix allowed my test case running at BCI on `lobata` to run successfully.

mvdebolskiy:
We've found another potential bug (NorESMhub/CAM#206):
In FatesSunShadeFracs:
https://github.com/NGEET/fates/blob/7d372b07f39c8280b26918c7eafcc82354efe5c2/radiation/FatesRadiationDriveMod.F90#L320C1-L323C23
You can notice that in FatesNormanRadiation:
https://github.com/NGEET/fates/blob/7d372b07f39c8280b26918c7eafcc82354efe5c2/radiation/FatesNormanRadMod.F90#L480
And in the Restart Interface:
https://github.com/NGEET/fates/blob/7d372b07f39c8280b26918c7eafcc82354efe5c2/main/FatesRestartInterfaceMod.F90#L3878C11-L3878C45
However, in the radiation driver itself:
https://github.com/NGEET/fates/blob/7d372b07f39c8280b26918c7eafcc82354efe5c2/radiation/FatesRadiationDriveMod.F90#L129C14-L129C70

Which means with Norman, fates will reset the rad_error, for the next timestep in CLM, but TwoStream does not do anything with it. Since wrap_sunfrac is actually called regardless of `doalb` flag, there is a possibility for ERS test to fail depending on the number of timesteps that are used for it (if the `doalb` is not `.true.` for every timestep).

glemieux:
Regression testing against `fates-sci.1.82.6_api.39.0.0-ctsm5.3.038` is underway on `derecho`.

glemieux:
The `NEON` files are all now passing, fixing escomp/ctsm#3038 as expected.  All DIFFs are showing in the `cpl` files for all tests with RMS values for the following variables:

```
RMS lndImp_Sl_anidf
RMS lndImp_Sl_anidr
RMS lndImp_Sl_avsdf
RMS lndImp_Sl_avsdr
```

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_pr1376-fates`

rgknox:
@mvdebolskiy, I think the solution here is to 1) zero the patch%rad_error on every timestep  on every model step prior to the normalized radiation call (regardless of doalb, will have to look to find a suitable location),  2) check to make sure that it is in the restart correctly (I assume it is but lets see), 3) make sure it is initialized as zero for a cold start and 4) remove the line where we set it to the ignore flag during the normalized radiation call.  Thoughts?

mvdebolskiy:
@rgknox this should work fine. Either way if it's consistent everywhere.

glemieux:
In discussion with @rgknox, we've decided to integrate this PR and Ryan will create a new PR to for the plan presented above: https://github.com/NGEET/fates/pull/1376#issuecomment-2822747010

PR #1407, opened 2025-05-12
Add some biophysical variables by land-use type

### LINKED ISSUE #1404: Biophysical variables by land use type?
We don't currently output any of the land-use variables resolved by land use type, we just have a few and they are mainly carbon cycle variables.  I suspect that including a small set of the biophysical ones would be of interest to include when land use change is active, i.e something like: veg temperature, 2m air temperature, sensible heat flux, latent heat flux, shortwave absorbed and net longwave. 

One issue is how to actually do this.  The relevant data is all outside of FATES, but the classic way of outputting sub-gridscale data via the ELM and CLM vector output format probably wouldn't work, because the patch land use labels aren't constant in time, so it wouldn't be straightforward to map from the patch vector to a (lat x lon x landuse) array in postprocessing of history files. 

The simplest way to do this I think would be to pass the relevant variables as a set of patch-resolved `bc_in` vectors  and then let FATES put them on the `fates_levlanduse` dimension.  It would have to be at high frequency, since these are necessarily all at that timestep.  

A more complicated solution might be to be to pass the land-use labels to ELM and CLM as `bc_out` vectors, and then do the remapping onto a land-use dimension outside of FATES. That might be more extensible, but I'm not sure it would be worth the additional complexity required to add that.  I suspect that the computational cost of each would be similar, but not really sure.

A further complication on any of this is the bare-ground stuff. Some fraction of bare-ground fluxes and temperatures would need to be averaged in with the vegetated patches of each land use type to do it properly.

ckoven:
In order to look at biophysical effects of land use, it's useful to have a small number of relevant variables indexed by land use type.  This PR allows that, so that you can look at: vegetation temperature, surface air temperature, and a basic set of energy fluxes (SW, LW, SH, LH) separately for each land use type. I.e., they are output on the `fates_levlanduse` history dimension for each gridcell.

This requires HLM-side logic. I have a commit on E3SM that handles this at https://github.com/ckoven/E3SM/commit/5fd137b0c9a43858efd925efc6c82dc698ec47be (PR forthcoming)

Currently these are all output when the hi frequency history flag is set to 1; in principle this should actually be when the flag is set to 2, so that should probably change before bringing this PR in. (for the purpose of the experiments I am currently running, I want these and only these subgridscale high-frequency variables though.)

Also, the logic that averages in bare-ground-patch properties will only work for nocomp configurations.  Further work is needed to be correct for full-FATES configurations as well, but I got lost in how to index a given site's bare-ground patch when nocomp isn't on, so I need to come back to that.

Also, the patch-weighting logic for these variables is slightly different than for other patch-level variables (e.g., the `fates_levage' -indexed ones). This is for two reasons: (1) we need to average in some of the bare ground patch properties to weight things correctly, because the FATES and HLM patch areas mean different things, and (2) the land use areas are less generally dynamic than the age binned logic, and so there is less chance of weirdness when the averaging denominator changes abruptly.  So the upshot is that less postprocessing is needed for these (i.e they are simpler to work with), but also that they might not sum perfectly to the total site values because of changing denominators over time.

### Description:
fixes #1404.

### Collaborators:
discussed with @glemieux, @rgknox, @aswann 

This only adds new diagnostic variables, so should be bit for bit otherwise.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -5017,7 +5017,7 @@ contains
     if(hlm_hist_level_hifrq>0) then
        call update_history_hifrq_sitelevel(this,nc,nsites,sites,bc_in,bc_out,dt_tstep)
        if (hlm_use_luh .eq. itrue) then
-          call update_history_hifrq_landuse(this,nc,nsites,sites,bc_in)
+          call update_history_hifrq_landuse(this,nc,nsites,sites,bc_in,dt_tstep)
        end if
        if(hlm_hist_level_hifrq>1) then
           call update_history_hifrq_subsite(this,nc,nsites,sites,bc_in,bc_out,dt_tstep)
@@ -5252,7 +5252,7 @@ contains
 
   ! ===============================================================================================
 
-  subroutine update_history_hifrq_landuse(this,nc,nsites,sites,bc_in)
+  subroutine update_history_hifrq_landuse(this,nc,nsites,sites,bc_in,dt_tstep)
 
     !
     ! Arguments
@@ -5261,6 +5261,7 @@ contains
     integer                 , intent(in)            :: nsites
     type(ed_site_type)      , intent(inout), target :: sites(nsites)
     type(bc_in_type)        , intent(in)            :: bc_in(nsites)
+    real(r8)                , intent(in)            :: dt_tstep
 
     ! Locals
     integer  :: s        ! The local site index
@@ -5272,8 +5273,11 @@ contains
     logical  :: foundbaregroundpatch
 
     type(fates_patch_type),pointer  :: cpatch
+    type(fates_cohort_type),pointer :: ccohort
+    real(r8) :: dt_tstep_inv          ! Time step in frequency units (/s)
 
     associate( hio_tveg_si_landuse    => this%hvars(ih_tveg_si_landuse)%r82d,&
+         hio_gpp_si_landuse           => this%hvars(ih_gpp_si_landuse)%r82d, &
          hio_tsa_si_landuse           => this%hvars(ih_tsa_si_landuse)%r82d,&
          hio_sw_abs_si_landuse        => this%hvars(ih_sw_abs_si_landuse)%r82d,&
          hio_lw_net_si_landuse        => this%hvars(ih_lw_net_si_landuse)%r82d,&
@@ -5283,6 +5287,8 @@ contains
       do_sites: do s = 1,nsites
 
          io_si  = sites(s)%h_gid
+
+         dt_tstep_inv = 1.0_r8/dt_tstep
 
          ! biophysical properties that are indexed by land use
          landuse_statevector(:) = sites(s)%get_current_landuse_statevector() * AREA
@@ -5376,6 +5382,22 @@ contains
 
          end do
 
+
+         cpatch => sites(s)%oldest_patch
+         do while(associated(cpatch))
+            ccohort => cpatch%shortest
+            do while(associated(ccohort))
+               if ( .not. ccohort%isnew ) then
+                  if (cpatch%land_use_label .gt. nocomp_bareground_land) then
+                     hio_gpp_si_landuse(io_si,cpatch%land_use_label) = hio_gpp_si_landuse(io_si,cpatch%land_use_label) &
+                          + ccohort%gpp_tstep * ccohort%n * dt_tstep_inv
+                  end if
+               end if
+               ccohort => ccohort%taller
+            end do
+            cpatch => cpatch%younger
+         end do
+
       end do do_sites
 
     end associate
@@ -5442,7 +5464,6 @@ contains
          hio_froot_mr_understory_si_scls     => this%hvars(ih_froot_mr_understory_si_scls)%r82d, &
          hio_resp_g_understory_si_scls       => this%hvars(ih_resp_g_understory_si_scls)%r82d, &
          hio_resp_m_understory_si_scls       => this%hvars(ih_resp_m_understory_si_scls)%r82d, &
-         hio_gpp_si_landuse                  => this%hvars(ih_gpp_si_landuse)%r82d, &
          hio_parsun_z_si_cnlf                => this%hvars(ih_parsun_z_si_cnlf)%r82d, &
          hio_parsha_z_si_cnlf                => this%hvars(ih_parsha_z_si_cnlf)%r82d, &
          hio_ts_net_uptake_si_cnlf           => this%hvars(ih_ts_net_uptake_si_cnlf)%r82d, &
@@ -5530,11 +5551,6 @@ contains
                     ! (kgC/m2/s) = (kgC/plant/s) * (plant/m2)
                     hio_ar_frootm_si_scpf(io_si,scpf) = hio_ar_frootm_si_scpf(io_si,scpf) + &
                          ccohort%froot_mr * n_perm2
-
-                    if (cpatch%land_use_label .gt. nocomp_bareground_land) then
-                       hio_gpp_si_landuse(io_si,cpatch%land_use_label) = hio_gpp_si_landuse(io_si,cpatch%land_use_label) &
-                            + ccohort%gpp_tstep * ccohort%n * dt_tstep_inv
-                    end if
 
                     ! accumulate fluxes on canopy- and understory- separated fluxes
                     if (ccohort%canopy_layer .eq. 1) then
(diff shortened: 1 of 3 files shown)


main/FatesHistoryInterfaceMod.F90
```diff
@@ -5240,6 +5252,160 @@ end subroutine update_history_hifrq_sitelevel
...
+         cpatch => sites(s)%oldest_patch
+         do while(associated(cpatch))
+            if (cpatch%total_canopy_area .gt. rsnbl_math_prec) then
+               ! for TVEG, since it is only defined on vegetated area of vegetated patches, normalize by the total vegetated area
+               hio_tveg_si_landuse(io_si,cpatch%land_use_label) = hio_tveg_si_landuse(io_si,cpatch%land_use_label) + &
+                    bc_in(s)%t_veg_pa(cpatch%patchno) * cpatch%total_canopy_area/canopy_area_bylanduse(cpatch%land_use_label)
+
+               ! for the rest of these, first weight by the vegetated area of each patch over the total patch area for each land use type
+               hio_tsa_si_landuse(io_si,cpatch%land_use_label) = hio_tsa_si_landuse(io_si,cpatch%land_use_label) + &
+                    bc_in(s)%t2m_pa(cpatch%patchno) * cpatch%total_canopy_area/landuse_statevector(cpatch%land_use_label)
```
samsrabin, line 5316: For normalization factors like `cpatch%total_canopy_area/landuse_statevector(cpatch%land_use_label)` that are repeated a bunch, it'd be good to create a variable for them with a useful name. E.g., what I did in `update_history_dyn_subsite_ageclass()` with [`patch_area_div_site_area`](https://github.com/NGEET/fates/blob/fe782ba4f851f8c8bfac8a89607e7c05bd397553/main/FatesHistoryInterfaceMod.F90#L4751).

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5240,6 +5252,160 @@ end subroutine update_history_hifrq_sitelevel
...
+               hio_shflux_si_landuse(io_si,cpatch%land_use_label) = hio_shflux_si_landuse(io_si,cpatch%land_use_label) + &
+                    bc_in(s)%shflux_pa(cpatch%patchno) * cpatch%total_canopy_area/landuse_statevector(cpatch%land_use_label)
+
+               hio_lhflux_si_landuse(io_si,cpatch%land_use_label) = hio_lhflux_si_landuse(io_si,cpatch%land_use_label) + &
+                    bc_in(s)%lhflux_pa(cpatch%patchno) * cpatch%total_canopy_area/landuse_statevector(cpatch%land_use_label)
+            endif
+            cpatch => cpatch%younger
+         end do
+
+         ! for all the land-use indexed variables, excpet for TVEG, also add in the component for the unvegetated area of each land use
```
samsrabin, line 5333: "excpet" typo.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5240,6 +5252,160 @@ end subroutine update_history_hifrq_sitelevel
...
+    real(r8)                , intent(in)            :: dt_tstep
+
+    ! Locals
+    integer  :: s        ! The local site index
+    integer  :: io_si     ! The site index of the IO array
+
+    real(r8) :: landuse_statevector(n_landuse_cats)
+    real(r8) :: canopy_area_bylanduse(n_landuse_cats)
+    integer  :: i_lu
+    logical  :: foundbaregroundpatch
```
samsrabin, line 5273: `foundbaregroundpatch` appears to be set but unused.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5240,6 +5252,160 @@ end subroutine update_history_hifrq_sitelevel
...
+         end do
+
+         ! for all the land-use indexed variables, excpet for TVEG, also add in the component for the unvegetated area of each land use
+         foundbaregroundpatch = .false.
+         cpatch => sites(s)%oldest_patch
+         do while(associated(cpatch))
+            if (cpatch%land_use_label .eq. nocomp_bareground_land .and. .not. foundbaregroundpatch) then
+               foundbaregroundpatch = .true.
+               do i_lu = 1, n_landuse_cats
+                  if ( landuse_statevector(i_lu) .gt. rsnbl_math_prec ) then
```
samsrabin, line 5340: To reduce indentation and thus readability, this `if ( landuse_statevector(i_lu) .gt. rsnbl_math_prec ) then ...` could be replaced by cycling on its inverse—`if ( landuse_statevector(i_lu) .le. rsnbl_math_prec ) then cycle`.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5240,6 +5252,160 @@ end subroutine update_history_hifrq_sitelevel
...
+            ! for tveg, ignore if there is no vegetation present on any patches of a given land use type
+            if ( canopy_area_bylanduse(i_lu) .le. rsnbl_math_prec ) then
+               hio_tveg_si_landuse(io_si,i_lu) = hlm_hio_ignore_val
+            end if
+
+         end do
+
+
+         cpatch => sites(s)%oldest_patch
+         do while(associated(cpatch))
```
samsrabin, line 5387: Why is this `do` loop separate from the others? Is it because GPP is the only one with the `ccohort%n * dt_tstep_inv` normalization factor? A comment would be useful to explain this.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5240,6 +5252,160 @@ end subroutine update_history_hifrq_sitelevel
...
+            ccohort => cpatch%shortest
+            do while(associated(ccohort))
+               if ( .not. ccohort%isnew ) then
+                  if (cpatch%land_use_label .gt. nocomp_bareground_land) then
```
samsrabin, line 5391: This `if` statement could be combined with the one above to reduce indentation.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -8743,6 +8903,51 @@ subroutine define_history_vars(this, initialize_variables)
...
+               ivar=ivar, initialize=initialize_variables, index = ih_shflux_si_landuse )
+
+          call this%set_history_var(vname='FATES_LHFLUX_LU', units='W m-2', &
+               long='fates latent heat flux by land use type', &
+               use_default='active', &
+               avgflag='A', vtype=site_landuse_r8, hlms='CLM:ALM', upfreq=group_hifr_simple, &
+               ivar=ivar, initialize=initialize_variables, index = ih_lhflux_si_landuse )
+
+          call this%set_history_var(vname='FATES_GPP_LU', units='kg m-2 s-1',        &
+               long='gross primary productivity by age bin in kg carbon per m2 per second', &
```
samsrabin, line 8945: "age bin" should be "land use type".

samsrabin (changes requested):
Looks good, Charlie! Just a few suggestions.

Also, a question: I'm having trouble understanding why and under what circumstances these should fail to sum to the site value. Could you explain that in a bit more detail?

glemieux:
@ckoven where is this on the priority queue for the CLM6 code freeze?

main/FatesHistoryInterfaceMod.F90
ckoven: It is used within the `if` clause on line https://github.com/NGEET/fates/pull/1407/files#diff-03feca0f1df8d3791c0d8f7b6055a0368d5241716d19d6a97c3d227fd6c5ad23R5337

main/FatesHistoryInterfaceMod.F90
ckoven: That's interesting -- I guess for me I prefer the extra indentation as it is easier to see the code block for which the `if` statement applies. for some reason I find `cycle` statements harder to parse visually.

main/FatesHistoryInterfaceMod.F90
ckoven: I made it separate because it is the only loop in this subroutine that loops over both patches and cohorts. adding comment to that effect.

glemieux:
@samsrabin would you confirm that your suggestions have been resolved and approve this if so?

main/FatesHistoryInterfaceMod.F90
samsrabin: Sure, different strokes for different folks!

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5358,6 +5370,162 @@ end subroutine update_history_hifrq_sitelevel
 
   ! ===============================================================================================
 
+
+  subroutine update_history_hifrq_landuse(this,nc,nsites,sites,bc_in,dt_tstep)
```
glemieux, line 5374: Linking @rgknox comment here [update_history_hifrq_landuse](https://github.com/NGEET/fates/pull/1529#discussion_r2783419381) due to closing #1529

main/FatesHistoryInterfaceMod.F90
```diff
@@ -5358,6 +5370,162 @@ end subroutine update_history_hifrq_sitelevel
...
+            if ( cpatch%land_use_label .ne. nocomp_bareground_land ) then
+               canopy_area_bylanduse(cpatch%land_use_label) = canopy_area_bylanduse(cpatch%land_use_label) + &
+                    cpatch%total_canopy_area
+            endif
+            cpatch => cpatch%younger
+         end do
+
+         cpatch => sites(s)%oldest_patch
+         do while(associated(cpatch))
+            if (cpatch%total_canopy_area .gt. rsnbl_math_prec) then
```
glemieux, line 5429: Copying @rgknox question here from https://github.com/NGEET/fates/pull/1529#discussion_r2783524535

Issue #1408, opened 2025-05-13
Expand rmean_type for other aggregation metrics


mpaiao:
### Task description

I am working on implementing a btran-based metric for drought deciduous phenology (main issue #1385), and I think for it to really work and avoid hydraulic failure mortality, we need to track the minimum daily btran instead of the daily average. I imagine at some point we may want a similar capability for maximum daily value, so I will likely implement it as well

This is how I plan to implement this in `FatesRunningMeanMod`:

1. Edit `rmean_type` class to add four more variables (`c_minimum`, `l_minimum`, `c_maximum`, `l_maximum`. 
2. Add functions `GetMin` and `GetMax` to retrieve the minimum and maximum.
3. Edit functions `InitRMean`, `UpdateRMean`, `FuseRMean` and `CopyFromDonor` so they handle the new variables too.
4. Existing flag `moving_ema_window` will be applied to the mean only, not the minima and maxima, which will always use the fixed windows. 

With this, it may make sense to rename the module, structure names and functions. I am thinking of `FatesRunningSummaryMod`, `rsummary_def_type` and `rsummary_type`, `InitRSummary`, etc. but I am open to other suggestions (including not changing it at all). Also, I suppose I should make this a separate pull request before bringing in the btran-based phenology, what do you think?

Issue #1385, opened 2025-04-09
Refactoring of cumulative thermal variables and soil moisture memory (phenology)


mpaiao:
### Task description

Through a few discussions with other FATES developers, several variables that are currently defined in sub-routine `phenology` would probably be better defined elsewhere, so they can be used by other parts of the code that are under implementation or planned. Examples include a crop model, crown fire, and mortality functions.

I started migrating these functions to a new module, and I will use this issue to discuss a few points that could use some feedback before I wrap up a pull request.

mpaiao:
I noticed that the cold deciduous phenology has one check that prevents leaves from flushing if the temperature never drops below the chilling temperature threshold (5°C, if using the default value of `fates_phen_chilltemp`). 

https://github.com/NGEET/fates/blob/7c45878697e92e5a333f783fdf08276a2f533772/biogeochem/EDPhysiologyMod.F90#L1116-L1133

The comment indicates that this is to avoid cold-deciduous plants to grow in the tropics. I can see how not having such check may allow cold-deciduous plants to grow in the tropics. But I am also wondering what happens in relatively mild subtropical/mid-latitude places, where some plants may be seasonally deciduous but the daily average temperatures may never or only rarely drop below the threshold.

mpaiao:
One suggestion for step 1 is to use the existing `FatesRunningMeanMod` instead of creating a new module. For now I am coding it in a new module, but it should be very trivial to move these new sub-routines to `FatesRunningMeanMod`. 

If you have a strong preference for one way or the other, please use the reaction button, let's say thumbs up for making a new module, thumbs down for not creating a new module and using `FatesRunningMeanMod` instead.

rosiealice:
Is it maybe slightly neater to put then in the existing routine so that there aren't loads of places where one might search for these things? I probably haven't thought enough about this to have a considered opinion but generelle it's great that you are doing this.

rosiealice:
Is it maybe slightly neater to put then in the existing routine so that there aren't loads of places where one might search for these things? I probably haven't thought enough about this to have a considered opinion but generelle it's great that you are doing this.

rosiealice:
Also, on this general theme, in my global simulations i am seeing near universal death of the drought deciduous plants from hydraulic failure in the first couple of months. I suspect that your Btran fix might well alter this behaviour, but just wanted to report that this is the present state of things...

mpaiao:
[...]

PR #1409, opened 2025-05-14
Add summary metrics for btran (and expand capability of running averages)

mpaiao:
### Description:

This pull request calculates summary variables for transpiration wetness factor (btran), which will be eventually used by drought deciduous as part of the updates to the phenology code (#1385). This builds on the running average structures that @rgknox implemented in #724, but now can be used to report the minimum and maximum values for variables using fixed moving average windows (as opposed to exponential moving averages). This is also a follow-up on issue #1408.

The rationale for this is that the mean daily btran may not be the best metric for tracking stress, especially if it varies markedly during the day (which I imagine is the case when FATES-Hydro is on). With this approach, we can use the minimum daily values, and this should provide a more robust way to compare with the values likely to trigger hydraulic failure mortality. Minimum and maximum daily values may be useful for other variables, applications or diagnostics too, and the implemented code can be readily used for other variables.

I did not implement the feature in exponential moving averages because I don't think it would make sense, but happy to modify it if anyone thinks otherwise or has some ideas on how to do it neatly. 

In the process of implementing the code, I saw one minor, likely benign bug in `FatesRestartInterfaceMod.F90` for variable `'fates_disturbance_rates'`, which was using the sub-routine for running averages instead of the regular one. I edited that, and it could impact the results. If this was my misunderstanding, I am happy to revert it. Otherwise, this should be bit-for-bit except for the new output variables.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


main/FatesRestartInterfaceMod.F90
```diff
@@ -1564,48 +1569,49 @@ subroutine define_restart_vars(this, initialize_variables)
          units='kg/m2/yr', flushval = flushzero, &
          hlms='CLM:ALM', initialize=initialize_variables, ivar=ivar, index = ir_emanpp_si)
     
-   call this%DefineRMeanRestartVar(vname='fates_tveg24patch',vtype=cohort_r8, &
+   call this%DefineRSummRestartVar(vname='fates_tveg24patch',vtype=cohort_r8, &
         long_name='24-hour patch veg temp', &
         units='K', initialize=initialize_variables,ivar=ivar, index = ir_tveg24_pa)
 
-   call this%DefineRMeanRestartVar(vname='fates_disturbance_rates',vtype=cohort_r8, &
+   call this%set_restart_var(vname='fates_disturbance_rates',vtype=cohort_r8, &
```
rgknox, line 1777: Interesting find @mpaiao . Yeah, looks like you found a bug

rgknox: As far as I can tell, this isn't breaking anything. All this is doing is inefficiently allocating two extra restart variables that aren't used (running mean restarts have three parts _cmean, _lmean _cindex).  The restart variable in this case will just use the "_cmean" portion. Either way, should be fixed.

rgknox:
Performed a re-review, and I think this PR is fairly straight forward. Most of the diffs are from re-naming existing subroutines.  The novel additions are focused around adding fixed window 24 hour averages, particularly btran for different pfts. The new fixed window subroutines added tracking of minimum and maximum values which is great. I don't really have any blockers for approval. I did question if we needed to add the minimum and maximum values to the restart file, but unfortunately those values can not be determined from any of the other memory variables, so they are required.

rgknox (approved):

mpaiao:
Thanks for reviewing it, @rgknox!

rgknox (approved):

rgknox:
Tests pass, made a couple of subtle bug fixes:
1) the way we define the indices for the registered values involved in the running sums, needed a fix to when it returned a dummy and when it returned the actual
2) needed a little more space in the restart vectors for SP runs to accomodate the btran running sums

PR #1424, opened 2025-06-13
promotion/demotion refactor and nclmax = 3

rgknox:
### Description:

This set of changes refactors promotion and demotion. The previous code was cumbersome and unclear, and had issues with achieving canopy cloures with more than 2 layers.  The new code re-uses the same routine for promotion and demotion, and does the calculations in a clearer more straightforward manor.  This also breaks up the termination call, where a new level is added "level 3" which is termination morality for plants in canopy layers greater than nclmax. This call is isolated to only be used after the promotion/demotion sequence has completed.  The refactor also introduces a new data type located on the patch structure that is a vector over max cohorts. The new data type has a cohort pointer and a real value used for weighting and determining demotion/promotion areas. This structure allows us to filter a list of cohorts and re-use that list without re-applying filtering logic, which makes the code more clear.

This would potentially replace #1315

I **do** expect non-b4b changes to results. It is possible that differences are not trivial, some of the probabilistic promotion/demotion code was quite complicated, so subtle changes associated with the new code could manifest.  But it should not affect SP because SP does not use promotion and demotion.
cc @adrifoster 

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


biogeochem/FatesPatchMod.F90
```diff
@@ -1141,14 +1164,16 @@ end subroutine CountCohorts
...
+    subroutine SortCohorts(this,check_order)
       !
       ! DESCRIPTION: sort cohorts in patch's linked list
       ! uses insertion sort to build a new list
       !
     
       ! ARGUMENTS:
       class(fates_patch_type), intent(inout), target :: this ! patch
+
+      logical,intent(in) :: check_order
```
glemieux: @rgknox, since this appears to be `.false.` in the majority of call cases, how about making this an optional argument that defaults to `.false.`?  I made a PR to this affect if that sounds good to you: https://github.com/rgknox/fates/pull/14

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -178,14 +200,23 @@ subroutine canopy_structure( currentSite , bc_in )
     !
     do while (associated(currentPatch)) ! Patch loop
 
+       ! Make sure we are sorted
+       ! call currentPatch%SortCohorts(check_order=.true.)
```
glemieux: @rgknox do you want to leave this in here as a `debug` option?

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -349,905 +377,315 @@ subroutine canopy_structure( currentSite , bc_in )
...
-  subroutine DemoteFromLayer(currentSite,currentPatch,i_lyr,bc_in)
-
-    use EDParamsMod, only : ED_val_comp_excln
+    ! --------------------------------------------------------------
+    ! This routine will:
+    ! 1) Identify the list of cohorts that are in the appropriate
+    !    layer for promotion or demotion into the adjacent
+    !    layer
+    ! 2) Calculate the combined crown area of those cohorts
+    !    that will be transferre to the adjacent layer
```
glemieux:
```suggestion
    !    that will be transferred to the adjacent layer
```

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -278,40 +302,44 @@ subroutine canopy_structure( currentSite , bc_in )
...
-                write(fates_log(),*) 'd2bl2: ',prt_params%allom_d2bl2(ipft)
-                write(fates_log(),*) 'd2bl_ediff: ',prt_params%allom_blca_expnt_diff(ipft)
-                write(fates_log(),*) 'd2ca_min: ',prt_params%allom_d2ca_coefficient_min(ipft)
-                write(fates_log(),*) 'd2ca_max: ',prt_params%allom_d2ca_coefficient_max(ipft)
-                currentCohort => currentCohort%shorter
+             do i_lyr = 1,z
+                write(fates_log(),*) '-----------------------------------------'
+                write(fates_log(),*) 'layer: ',i_lyr,' area: ',arealayer(i_lyr)
+                write(fates_log(),*) 'abs error (layer-patch): ',(arealayer(i_lyr)- &
+                     (1._r8-imperfect_fraction)*currentPatch%area)
```
glemieux: Did you want to make this the absolute error?
```suggestion
                write(fates_log(),*) 'abs error (layer-patch): ',abs(arealayer(i_lyr)- &
                     (1._r8-imperfect_fraction)*currentPatch%area)
```

glemieux (changes requested):
Review part 1/2 suggesting some fixes.  I still need to finish reviewing the updated promotion/demotion subroutine.

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -349,905 +377,315 @@ subroutine canopy_structure( currentSite , bc_in )
...
+      ! We update the target area to be no more than the
+      ! area of the layer (can't take more than there is..)
+      promdem_area = min(target_area,group_area)
+
+      
+      ! Store the number of cohorts in the layer
+      ! and zero out the array of area transfers
+      n_layer = ic
+      do ic = 1,n_layer
+         layer_co(ic)%pd_area = 0._r8
```
glemieux, line 470: Suggestion: zero out all indices of the vector at once, i.e.:
```suggestion
         layer_co(:)%pd_area = 0._r8
```

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -349,905 +377,315 @@ subroutine canopy_structure( currentSite , bc_in )
...
+         end do
+
+         if(excess_area>nearzero)then
+            do ic = 1,n_layer
+               cohort => layer_co(ic)%p
+               ! look at just the cohorts that still have space to give
+               ! remove from them the same fraction of their remaining space
+               if (abs(layer_co(ic)%pd_area-cohort%c_area) > nearzero) then
+                  layer_co(ic)%pd_area = layer_co(ic)%pd_area + &
+                       (excess_area/remainder_area) * &
```
glemieux: I assume this is unlikely, but is it possible that `remainder_area` could stay zero here?  For example, could `attempt_area` always be greater that `c_area` in line 503 above?

Maybe if there is only one cohort in the target layer and if `promdem = c_area` and if you were trying to promote then you could end up with that case, but is that even likely/possible?

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -349,905 +377,315 @@ subroutine canopy_structure( currentSite , bc_in )
...
+
+         cohort => layer_co(ic)%p
+
+         ! If the dem/prom area is the same area as the
+         !    cohort itself, move the whole thing
+         ! If the dem/prom area is less than the cohort area
+         !    and not trivialy small (larger than precision
+         !    check), then split it and move part of it
+         ! If the dem/prom area is less than zero or larger than
+         !    the cohort area within precision checks then FAIL
```
glemieux: Missing failure block?

glemieux (commented):
Review part 2/2.  A few questions and one suggestion.

biogeochem/EDCanopyStructureMod.F90
rgknox: Will add a check for area greater than the cohorts existing area, or less than zero.

biogeochem/FatesPatchMod.F90
rgknox: Nice, looks good to me, I'll merge in

biogeochem/EDCanopyStructureMod.F90
rgknox: yes, thats a better way to do it, I'll apply

biogeochem/EDCanopyStructureMod.F90
rgknox: ahh, no, I should leave it as the bias. will fix

biogeochem/EDCanopyStructureMod.F90
rgknox: hmm, this might not be allowed? Its a vector of data structures with a primitive in it's members and not a vector of primitives.

biogeochem/EDCanopyStructureMod.F90
rgknox: Yeah, I see what you mean... I think your right, we need to protect this situation..

biogeochem/EDCanopyStructureMod.F90
rgknox: done

biogeochem/EDCanopyStructureMod.F90
rgknox: done

biogeochem/EDCanopyStructureMod.F90
rgknox: after discussions with @glemieux and team, we believe vectorized assignment here could probably work, but we are not sure for all compilers, and this is only a minor potential efficiency gain

biogeochem/EDCanopyStructureMod.F90
rgknox: The failure check was provided in the checks before this logical block, so I updated text

PR #1427, opened 2025-06-18
Updates to hydraulic failure mortality

mpaiao:
### Description:

This is a bit tangential, but in line with the refactoring of the deciduous phenology (Issue #1385).

FATES currently has two parameters for defining hydraulic failure thresholds. When FATES-Hydro is enabled, we use `fates_mort_hf_flc_threshold`, which is based on the fraction of loss of conductivity (`flc`). When FATES-Hydro is disabled, we set `fates_mort_hf_sm_threshold`, which is a bit of a misnomer, because the threshold is based on `btran`, not soil moisture.

This pull request replaces these two parameters with a single one, `fates_mort_hydrfailure_threshold`, which is based on `btran` or the fraction of maximum conductivity (`fmc`), which is simply `fmc = 1 - flc`. The idea behind this switch is because `btran` is numerically equivalent to `fmc` when FATES-Hydro is enabled, so we can define the thresholds through a single parameter. 

For FATES-Hydro simulations, this only affects how we set the threshold, because now the lower the value, the more drought resistant the plant is going to be. Otherwise, the calculation of hydraulic failure mortality is logically the same. Results will probably not be bit-for-bit, though, because the mathematical formulation was adjusted to use `fmc` instead of `flc`. 

For runs without FATES-Hydro, the proposed pull request would change how mortality is calculated. Before we were applying a step function: if `btran` dropped below `fates_mort_hf_sm_threshold`, mortality would immediately become `fates_mort_scalar_hydrfailure`. In this pull request, it follows the same formulation as FATES-Hydro: hydraulic failure mortality increases linearly from 0 when `btran = fates_mort_hydrfailure_threshold` to `fates_mort_scalar_hydrfailure` when `btran = 0`. Evidently, this is a different formulation, so results will be definitely different when FATES-Hydro is disabled.

Merging this pull request will require a change in the technical documentation. The existing technical documentation is not up to date in any case: it explains how current FATES calculates hydraulic failure mortality when FATES-Hydro is disabled (but not when FATES-Hydro is enabled), and implies that the threshold (`fates_mort_hf_sm_threshold`) is not an adjustable parameter.

### Collaborators:

Discussed this at the developer meeting with @glemieux @ckoven @JessicaNeedham @adrifoster @samsrabin 

The results will not be bit-for-bit. Changes should be just due to numerical precision when FATES-Hydro is on, but should be rather different when FATES-Hydro is off. 

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/parameter_files/fates_params_default.cdl
+++ b/parameter_files/fates_params_default.cdl
@@ -444,9 +444,9 @@ variables:
 	double fates_mort_freezetol(fates_pft) ;
 		fates_mort_freezetol:units = "degrees C" ;
 		fates_mort_freezetol:long_name = "minimum temperature tolerance" ;
-	double fates_mort_hydrfail_threshold(fates_pft) ;
-		fates_mort_hydrfail_threshold:units = "fraction" ;
-		fates_mort_hydrfail_threshold:long_name = "threshold (btran units) below which hydraulic failure mortality begins. When running FATES-Hydro, this is numerically equivalent to fraction of total conductivity (or one minus fraction of loss of conductivity)." ;
+	double fates_mort_hydrfailure_threshold(fates_pft) ;
+		fates_mort_hydrfailure_threshold:units = "fraction" ;
+		fates_mort_hydrfailure_threshold:long_name = "threshold (btran units) below which hydraulic failure mortality begins. When running FATES-Hydro, this is numerically equivalent to fraction of maximum conductivity (or one minus fraction of loss of conductivity)." ;
 	double fates_mort_ip_age_senescence(fates_pft) ;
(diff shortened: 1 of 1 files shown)


rgknox:
cc @rosiealice , these changes would impact nocomp/fixebiogeo results (or  anything non sp for that matter)

rgknox:
Cross referencing work done on noresm-fates by @rosiealice : https://github.com/NorESMhub/fates/pull/28
Intention here is to consolidate, unify efforts.

PR #1426, opened 2025-06-18
Unit/functional test improvements to enable CTSM SystemTests

samsrabin:
### Description:
- Gets actual path to default config file to enable running from anywhere.
- Adds an error check that would have helped in my troubleshooting.
- Adds optional `--config-file` argument.
- Adds netCDF file required for unit testing.

Previously filed as adrifoster/fates#4.
CTSM PR here: ESCOMP/CTSM#3251

**Waiting on NGEET GitHub org to get Git LFS working.**

### Collaborators:
- @adrifoster 

### Expectation of Answer Changes:
None

### Checklist

*If this is your first time contributing, please read the [**CONTRIBUTING**](https://github.com/NGEET/fates/blob/main/CONTRIBUTING.md) document.*

All checklist items must be checked to enable merging this pull request:

*Contributor*
- [x] The in-code documentation has been updated with descriptive comments
- [x] The documentation has been assessed to determine if updates are necessary

*Integrator*
- [ ] FATES PASS/FAIL regression tests were run
- [ ] Evaluation of test results for answer changes was performed and results provided

### Test Results:

*CTSM (or) E3SM (specify which) test hash-tag:*

*CTSM (or) E3SM (specify which) baseline hash-tag:*

*FATES baseline hash-tag:*

*Test Output:*

PR #1432, opened 2025-06-30
Diagnostically track forest edge

samsrabin:
### Description:
This PR adds the ability to track how much of a site's forest area is in each "bin" of distance from a forest edge. This will eventually be used for the escaped fire work (forest closer to edge will have a higher fire danger index), but for now it's purely diagnostic. I will also hide all this behind a feature flag, so ~~none of the new variables will be allocated and~~ no calculations will happen unless the user explicitly requests it. **2025-09-10: I ended up running into problems with not allocating the variables, so I reverted that bit in 4d3b7d4.**

Includes changes from #1426; will look bigger than it really is until that is merged.

Corresponding CTSM PR is ESCOMP/CTSM#3340.
#### New parameter file parameters
The first step in this work was empirical analysis to develop a function for assigning forest to different edge distance bins (including what set of bins to use). I made a Python project to explore the possibilities and published it [here](https://github.com/samsrabin/Lin_edgeareas/tree/v0.1), with `lin_edgeareas_explore.py` being the script to run if you want to generate the parameter sets and figures yourself. If you look through that script you can see various commented-out options I went through during the process; the v0.1.1 tag shows the exact setup needed to reproduce my final parameter set. (I haven't tested the different options with v0.1.1, so some or all of them may be broken.) Briefly, given land cover and forest edge bin area from Lin's remote sensing work, it fits (for each bin) the fraction of the gridcell's forest area in the bin (Y variable) to the total forested fraction of the gridcell. It tries the following fits, choosing the best for each bin based on lowest AIC value (**bold** indicates fit types that actually get chosen in v0.1.1):
- **Gaussian**
- Skewed Gaussian
- **Lognormal**
- Logistic
- Exponential
- Linear
- **Quadratic**

It then saves a .cdl file skeleton with the resulting parameters, in addition to making a couple of figures illustrating the results.

This PR adds to the FATES parameter file:
- Variable `fates_forest_tree_fraction_threshold`: Tree fraction above which patch is considered "forest." Set to 0.5; **need to check whether this makes sense based on Lin's landcover data**.
- Dimension `fates_edgeforest_bins`: Number of edge forest bins, including "deep forest." All the following parameter file variables have this dimension.
- Variable `fates_edgeforest_bin_edges`: Boundaries of forest edge bins (for each bin, include value closest to zero): 0, 30, 60, 120, 150, and 300.
- Variables `fates_edgeforest_gaussian_*` (3): Parameters describing fits for bins with Gaussian fits.
- Variables `fates_edgeforest_lognormal_*` (3): Parameters describing fits for bins with lognormal fits.
- Variables `fates_edgeforest_quadratic_*` (3): Parameters describing fits for bins with quadratic fits.

Each edge forest bin has only one fit type, so the .cdl file has `_` for that bin in parameters of the other types.

#### New code
This PR adds three modules to the FATES code. The first, `main/FatesEcotypesMod.F90`, is used to determine whether a patch is "forest" or not. I named it `FatesEcotypesMod` instead of something about forest because it could be extended in the future for other land cover classes. The `is_patch_forest()` function checks whether a patch is forest based on the `fates_forest_tree_fraction_threshold` parameter (see above) as well as an optional grass biomass threshold (above which a patch can't be forest). The latter is not currently used for anything except experimental outputs, as there is no corresponding variable on the parameter file.

The other two have to do with edge forest specifically. The edge forest parameters are read in `main/FatesEdgeForestParamsMod.F90` to avoid junking up `EdParamsMod`. Those parameters are then used in `main/FatesEdgeForestMod.F90`, the main purpose of which is `get_fraction_of_edgeforest_in_each_bin()`. That subroutine loops through a site's patches from nearest to farthest from edge, assigning their area to the various edge forest bins. It is called via the public subroutine `calculate_edgeforest_area()` in `CalculateTreeGrassAreaSite()`(moved to `FatesEdgeForestMod` from `main/EdTypesMod.F90`, I think because of circular dependencies). I also added a call of it to `canopy_summarization()` (`biogeochem/EDCanopyStructureMod.F90`), but that's not actually called anywhere.

#### New outputs
- `FATES_AREA_PLANTS_AP` and `FATES_AREA_TREES_AP`: Age-structured versions of existing variables; these might be useful for experimentation as this work continues.
- `FATES_IS_FOREST` and `FATES_IS_FOREST_AP`: Fraction of the site (or each age bin in the site) that is forest according to the `fates_forest_tree_fraction_threshold` parameter (see above).
- `FATES_IS_FOREST_PCTX`: Fraction of the site that is forest according to a tree cover threshold of X% instead of the `fates_forest_tree_fraction_threshold` parameter.
- `FATES_IS_FOREST_PCTX_0GRASS`: Fraction of the site that is forest according to a tree cover threshold of X% and a grass biomass maximum of zero (although as noted above I currently have the grass biomass bit backwards).
- `FATES_FOREST_AREA_EB`: Fraction of the site's forest area that is in each edge bin. Has new `fates_levedge` dimension.

#### New namelist parameters

If the HLM's `use_edge_forest` parameter is `.false.`:
- Arrays are allocated but only for one bin (modeled after `hlm_use_tree_damage`).
- Edge forest area is not calculated.
- `FATES_FOREST_AREA_EB` output can be saved but will just have the initialization value.

### Checking results

Bins are labeled by their minimum distance to edge (m). Adjusted/normalized Y values show the results after adjusting the raw fit values so that the sum of all curves at any X value is equal to 1.

Expectation from my Python notebook based on Lin's work (`lin_edgeareas_explore.py` [here](https://github.com/samsrabin/Lin_edgeareas/tree/v0.1.1)):
screenshot_4115

✅ Functional testing:
edge_forest_plot_frac_in_every_bin 1

✅ Results at end of a four-year test (comparing to dashed lines from above):
screenshot_4117

### Collaborators:
@mpaiao, @ckoven, @menglinmet 

None

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/testing/run_unit_tests.py
+++ b/testing/run_unit_tests.py
@@ -18,7 +18,6 @@ This script builds and runs FATES units tests.
 
 """
 import os
-from shutil import copyfile
 import argparse
 
 from build_fortran_tests import build_tests
@@ -27,9 +26,7 @@ from utils import config_to_dict, parse_test_list
 
 add_cime_lib_to_path()
 
-from CIME.utils import (
-    run_cmd_no_fail,
-)  # pylint: disable=wrong-import-position,import-error,wrong-import-order
+from CIME.utils import run_cmd_no_fail  # pylint: disable=wrong-import-position,import-error,wrong-import-order
 
 # constants for this script
 _FILE_DIR = os.path.dirname(os.path.abspath(__file__))
@@ -129,7 +126,6 @@ def run_unit_tests(clean, verbose_make, build_dir, make_j, test_dict):
 
         test_dir = os.path.join(build_dir_path, _TEST_SUB_DIR, attributes["test_dir"])
         ctest_command = ["ctest", "--output-on-failure"]
-
         output = run_cmd_no_fail(
             " ".join(ctest_command), from_dir=test_dir, combine_output=True
         )
@@ -146,11 +142,7 @@ def main():
     test_dict = parse_test_list(full_test_dict, args.test_list)
 
     run_unit_tests(
-        args.clean,
-        args.verbose_make,
-        args.build_dir,
-        args.make_j,
-        test_dict,
+        args.clean, args.verbose_make, args.build_dir, args.make_j, test_dict
     )
 
 
--- a/testing/run_functional_tests.py
+++ b/testing/run_functional_tests.py
@@ -34,8 +34,7 @@ import matplotlib.pyplot as plt
 from build_fortran_tests import build_tests, build_exists
 from functional_class_with_drivers import FunctionalTestWithDrivers
 from path_utils import add_cime_lib_to_path
-from utils import copy_file, config_to_dict, parse_test_list
-from utils import check_param_file, create_param_file, _DEFAULT_CDL_PATH
+from utils import copy_file, create_nc_from_cdl, config_to_dict, parse_test_list
 
 # load the functional test classes
 from load_functional_tests import *
@@ -47,6 +46,14 @@ from CIME.utils import run_cmd
 # constants for this script
 _FILE_DIR = os.path.dirname(__file__)
 _DEFAULT_CONFIG_FILE = os.path.join(_FILE_DIR, "functional_tests.cfg")
+_DEFAULT_CDL_PATH = os.path.abspath(
+    os.path.join(
+        _FILE_DIR,
+        os.pardir,
+        "parameter_files",
+        "fates_params_default.cdl",
+    )
+)
 _CMAKE_BASE_DIR = os.path.join(_FILE_DIR, os.pardir)
 _TEST_SUB_DIR = "testing"
 
@@ -196,6 +203,25 @@ def check_arg_validity(args):
         raise RuntimeError(f"config 'file' is a directory: '{args.config_file}'")
 
 
+def check_param_file(param_file):
+    """Checks to see if param_file exists and is of the correct form (.nc or .cdl)
+
+    Args:
+        param_file (str): path to parameter file
+
+    Raises:
+        argparse.ArgumentError: Parameter file is not of the correct form (.nc or .cdl)
+        argparse.ArgumentError: Can't find parameter file
+    """
+    file_suffix = os.path.basename(param_file).split(".")[-1]
+    if not file_suffix in ["cdl", "nc"]:
+        raise argparse.ArgumentError(
+            None, "Must supply parameter file with .cdl or .nc ending."
+        )
+    if not os.path.isfile(param_file):
+        raise FileNotFoundError(param_file)
+
+
 def check_build_dir(build_dir, test_dict):
     """Checks to see if all required build directories and executables are present
 
@@ -332,6 +358,37 @@ def make_plotdirs(run_dir, test_dict):
         sub_dir = os.path.join(plot_dir, test)
         if not os.path.isdir(sub_dir):
             os.mkdir(sub_dir)
+
+
+def create_param_file(param_file, run_dir):
+    """Creates and/or move the default or input parameter file to the run directory
+    Creates a netcdf file from a cdl file if a cdl file is supplied
+
+    Args:
+        param_file (str): path to parmaeter file
+        run_dir (str): full path to run directory
+
+    Raises:
+        RuntimeError: Supplied parameter file is not netcdf (.cd) or cdl (.cdl)
+
+    Returns:
+        str: full path to new parameter file name/location
+    """
+    if param_file is None:
+        print("Using default parameter file.")
+        param_file = _DEFAULT_CDL_PATH
+        param_file_update = create_nc_from_cdl(param_file, run_dir)
+    else:
+        print(f"Using parameter file {param_file}.")
+        file_suffix = os.path.basename(param_file).split(".")[-1]
+        if file_suffix == "cdl":
+            param_file_update = create_nc_from_cdl(param_file, run_dir)
+        elif file_suffix == "nc":
+            param_file_update = copy_file(param_file, run_dir)
+        else:
+            raise RuntimeError("Must supply parameter file with .cdl or .nc ending.")
+
+    return param_file_update
 
 
 def run_fortran_exectuables(build_dir, test_dir, test_exe, run_dir, args):
--- a/testing/utils.py
+++ b/testing/utils.py
@@ -14,17 +14,6 @@ from CIME.utils import (
     run_cmd_no_fail,
 )  # pylint: disable=wrong-import-position,import-error,wrong-import-order
 
-# constants for this script
-_FILE_DIR = os.path.dirname(__file__)
-_DEFAULT_CDL_PATH = os.path.abspath(
-    os.path.join(
-        _FILE_DIR,
-        os.pardir,
-        "parameter_files",
-        "fates_params_default.cdl",
-    )
-)
-
 
 def round_up(num: float, decimals: int = 0) -> float:
     """Rounds a number up
@@ -214,55 +203,3 @@ def str_to_list(val: str) -> list:
         return []
     res = val.strip("][").split(",")
     return [n.strip() for n in res]
-
-
-def check_param_file(param_file):
-    """Checks to see if param_file exists and is of the correct form (.nc or .cdl)
-
-    Args:
-        param_file (str): path to parameter file
-
-    Raises:
-        argparse.ArgumentError: Parameter file is not of the correct form (.nc or .cdl)
-        argparse.ArgumentError: Can't find parameter file
-    """
-    file_suffix = os.path.basename(param_file).split(".")[-1]
-    if not file_suffix in ["cdl", "nc"]:
-        raise argparse.ArgumentError(
-            None, "Must supply parameter file with .cdl or .nc ending."
-        )
-    if not os.path.isfile(param_file):
-        raise FileNotFoundError(param_file)
-
-
-def create_param_file(param_file, run_dir):
-    """Creates and/or move the default or input parameter file to the run directory
-    Creates a netcdf file from a cdl file if a cdl file is supplied
-
-    Args:
-        param_file (str): path to parmaeter file
-        run_dir (str): full path to run directory
-
-    Raises:
-        RuntimeError: Supplied parameter file is not netcdf (.cd) or cdl (.cdl)
-
-    Returns:
-        str: full path to new parameter file name/location
-    """
-    if param_file is None:
-        print("Using default parameter file.")
-        param_file = _DEFAULT_CDL_PATH
-        param_file_update = create_nc_from_cdl(param_file, run_dir)
-    else:
-        print(f"Using parameter file {param_file}.")
-        file_suffix = os.path.basename(param_file).split(".")[-1]
-        if file_suffix == "cdl":
-            param_file_update = create_nc_from_cdl(param_file, run_dir)
-        elif file_suffix == "nc":
-            param_file_update = copy_file(param_file, run_dir)
-        else:
-            raise RuntimeError("Must supply parameter file with .cdl or .nc ending.")
-
-    print(f"Parameter file saved to {os.path.abspath(param_file_update)}")
-
-    return param_file_update


rosiealice:
Nice idea @samsrabin!  This will have lots of potential uses I'm sure...

main/FatesEcotypesMod.F90
```diff
@@ -0,0 +1,78 @@
...
+  ! RETURN VALUE
+  logical :: does_patch_have_forest_tcthresh
+  !
+  ! LOCAL VARIABLES
+  real(r8) :: tree_fraction = 0._r8
+
+  if (patchptr%area > 0._r8) then
+      tree_fraction = patchptr%total_tree_area / patchptr%area
+  else
+      tree_fraction = 0._r8
```
samsrabin, line 34: This `tree_fraction = 0._r8` is unnecessary because it's initialized to 0 when it's defined.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -2683,6 +2736,31 @@ subroutine update_history_dyn_sitelevel(this,nc,nsites,sites,bc_in)
...
+            hio_is_forest_pct10_0grass_si(io_si) = hio_is_forest_pct10_0grass_si(io_si) + cpatch%area * AREA_INV * &
+               merge(1._r8, 0._r8, is_patch_forest(cpatch, 0.1_r8, grass_biomass_threshold=0._r8))
+            hio_is_forest_pct25_0grass_si(io_si) = hio_is_forest_pct25_0grass_si(io_si) + cpatch%area * AREA_INV * &
+               merge(1._r8, 0._r8, is_patch_forest(cpatch, 0.25_r8, grass_biomass_threshold=0._r8))
+            hio_is_forest_pct50_0grass_si(io_si) = hio_is_forest_pct50_0grass_si(io_si) + cpatch%area * AREA_INV * &
+               merge(1._r8, 0._r8, is_patch_forest(cpatch, 0.5_r8, grass_biomass_threshold=0._r8))
+            hio_is_forest_pct75_0grass_si(io_si) = hio_is_forest_pct75_0grass_si(io_si) + cpatch%area * AREA_INV * &
+               merge(1._r8, 0._r8, is_patch_forest(cpatch, 0.75_r8, grass_biomass_threshold=0._r8))
+            hio_is_forest_pct90_0grass_si(io_si) = hio_is_forest_pct90_0grass_si(io_si) + cpatch%area * AREA_INV * &
+               merge(1._r8, 0._r8, is_patch_forest(cpatch, 0.9_r8, grass_biomass_threshold=0._r8))
```
samsrabin, line 2762: Replace `merge(1._r8, 0._r8, is_patch_forest(cpatch, 0.9_r8, grass_biomass_threshold=0._r8))` with a new `logical_to_integer` function.

main/FatesInterfaceMod.F90
```diff
@@ -1676,15 +1699,16 @@ subroutine set_fates_ctrlparms(tag,ival,rval,cval)
...
-            if((hlm_use_tree_damage .eq. itrue) .and. &
+         else if ((hlm_use_tree_damage .eq. itrue) .and. &
                  (hlm_parteh_mode .eq. prt_cnp_flex_allom_hyp))then
-               write(fates_log(),*) 'FATES tree damage (use_fates_tree_damage = .true.) is not'
-               write(fates_log(),*) '(yet) compatible with CNP allocation (fates_parteh_mode = 2)'
-               call endrun(msg=errMsg(sourcefile, __LINE__))
+            write(fates_log(),*) 'FATES tree damage (use_fates_tree_damage = .true.) is not'
+            write(fates_log(),*) '(yet) compatible with CNP allocation (fates_parteh_mode = 2)'
+            call endrun(msg=errMsg(sourcefile, __LINE__))
          end if
```
samsrabin, line 1715: I rearranged the above `if` block because the indentation was confusing.

testing/functional_testing/edge_forest/FatesTestEdgeForest.F90
```diff
@@ -0,0 +1,267 @@
...
+  call WriteVar(ncid, fracforestID, frac_forest(:))
+  call WriteVar(ncid, gaussianID, frac_in_bin_gaussian(:))
+  call WriteVar(ncid, lognormalID, frac_in_bin_lognormal(:))
+  call WriteVar(ncid, quadraticID, frac_in_bin_quadratic(:))
+  call WriteVar(ncid, everybinID, frac_in_every_bin(:,:))
+
+  ! close the file
+  call CloseNCFile(ncid)
+
+end subroutine WriteEdgeForestData
```
samsrabin, line 267: Add blank line at EOF.

main/EDTypesMod.F90
```diff
@@ -703,39 +702,6 @@ end subroutine ZeroMassBalFlux
    
   ! =====================================================================================
 
-  subroutine CalculateTreeGrassAreaSite(csite, tree_fraction, grass_fraction, bare_fraction)
```
samsrabin, line 740: Moved to FatesEdgeForestMod, I think because of circular dependencies.

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,653 @@
...
+  public :: gffeb_norm_denominator
+  public :: gffeb_norm
+  public :: gffeb_quadratic
+  public :: assign_patch_to_bins
+
+contains
+
+  ! =====================================================================================
+
+  subroutine CalculateTreeGrassAreaSite(csite, tree_fraction, grass_fraction, bare_fraction)
```
samsrabin, line 34: Moved from `EdTypesMod`, I think because of circular dependencies.

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,653 @@
...
+      end if
+      currentPatch => currentPatch%younger
+    end do
+
+    ! if cover > 1.0, grasses are under the trees
+    grass_fraction = min(grass_fraction, 1.0_r8 - tree_fraction)
+    bare_fraction = 1.0_r8 - tree_fraction - grass_fraction
+
+    ! Must come after patch loop with is_patch_forest() call
+    call calculate_edgeforest_area(csite)
```
samsrabin: This call of `calculate_edgeforest_area()` should be the only difference from the original `CalculateTreeGrassAreaSite()` implementation in `EdTypesMod`.

testing/functional_testing/edge_forest/edge_forest_test.py
```diff
@@ -0,0 +1,173 @@
+"""
+Concrete class for running the allometry functional tests for FATES.
```
samsrabin: Change "allometry" to "edge forest."

main/EDInitMod.F90
```diff
@@ -58,6 +58,7 @@ module EDInitMod
   use FatesInterfaceTypesMod         , only : hlm_use_inventory_init
   use FatesInterfaceTypesMod         , only : hlm_use_fixed_biogeog
   use FatesInterfaceTypesMod         , only : hlm_use_tree_damage
+  use FatesInterfaceTypesMod         , only : hlm_use_edge_forest
```
rgknox: can this use be removed?

samsrabin: Good catch!

main/FatesEcotypesMod.F90
```diff
@@ -0,0 +1,81 @@
+module FatesEcotypesMod
```
rgknox, line 1: Do you foresee this module being used to classify patches as more than forest? Or is it going to stay as a forest versus non-forest classification?  I ask because this module is named and isolated to be generalized for other ecotypes in the future possibly.

samsrabin: I'm not planning any such work, but I figured it'd be smart to name it this way just in case.

testing/run_unit_tests.py
```diff
@@ -96,12 +100,26 @@ def commandline_args():
         "for all tests. If not supplied, will run all tests.",
     )
 
+    parser.add_argument(
```
adrifoster: Generally it is considered bad practice to have unit tests do I/O. Is this absolutely necessary? For other unit testing where we have needed to use parameters I have just used hard-coded values. 

I am pretty reticent to start using I/O in our unit tests, but happy to discuss.

samsrabin: Good point! I can work on that.

main/FatesEcotypesMod.F90
```diff
@@ -0,0 +1,81 @@
...
+  real(r8), intent(in) :: forest_tree_fraction_threshold ! Tree fraction above which a patch is "forest"
+  real(r8), intent(in), optional :: grass_biomass_threshold ! Live grass biomass (kgC/m2) above which a patch is considered to "have grass"
+  !
+  ! RETURN VALUE
+  logical :: is_patch_forest
+
+  is_patch_forest = does_patch_have_forest_tcthresh(patchptr, forest_tree_fraction_threshold)
+  if (is_patch_forest .and. present(grass_biomass_threshold)) then
+     is_patch_forest = .not. does_patch_have_grass_bmthresh(patchptr, grass_biomass_threshold)
+  end if
```
ckoven: Would be useful to explain the logic here around the use of the grass biomass threshold for forest definition. Is there a reference to cite for this definition of forest?

ckoven: Though I am also noticing that the `grass_biomass_threshold` isn't actually used in the current logic, other than in some diagnostic history variables and unit tests.

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,653 @@
...
+    ! LOCALS:
+    type(fates_patch_type), pointer :: currentPatch ! patch object
+
+    tree_fraction = 0.0_r8
+    grass_fraction = 0.0_r8
+
+    currentPatch => csite%oldest_patch
+    do while(associated(currentPatch))
+      if (currentPatch%nocomp_pft_label /= nocomp_bareground) then
+        call currentPatch%UpdateTreeGrassArea()
```
ckoven, line 57: The function `currentPatch%UpdateTreeGrassArea()` is outside the sope of this PR, but I just realized that the variable `currentPatch%total_tree_area` is being defined in two subtly different ways in different parts of the code. I don't understand why it should be that way, so unless someone else does, we should align the two definitions.

In [currentPatch%UpdateTreeGrassArea()](https://github.com/NGEET/fates/blob/main/biogeochem/FatesPatchMod.F90#L809-L820), the `total_tree_area` and `total_grass_area` are calculated for both canopy and understory plants, but then clipped so that they cannot exceed the patch area.  

Whereas in [canopy_summarization](https://github.com/NGEET/fates/blob/main/biogeochem/EDCanopyStructureMod.F90#L860-L865), `total_tree_area`is calculated only for canopy trees (and thus cannot exceed patch area, unless something is wrong).

If PPA logic is strictly rank-ordered and trees are always taller than grasses, then this should not make much of a difference. It seems to me like we should base all of these definitions only on canopy plants, so I'd advocate changing the logic in currentPatch%UpdateTreeGrassArea() to align with the current logic in canopy_summarization. But curious to understand the difference.

main/FatesEcotypesMod.F90
```diff
@@ -0,0 +1,81 @@
...
+  use FatesPatchMod, only : fates_patch_type
+
+  implicit none
+  private  ! By default everything is private
+
+  ! Make public necessary subroutines and functions
+  public :: is_patch_forest
+  ! For unit testing
+  public :: does_patch_have_forest_tcthresh
+  public :: does_patch_have_grass_bmthresh
```
ckoven: I think we had agreed to use camel case for procedure names?

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,653 @@
...
+  public :: CalculateTreeGrassAreaSite
+  public :: calculate_edgeforest_area
+  ! Public for unit testing
+  public :: indexx
+  public :: get_fraction_of_edgeforest_in_each_bin
+  public :: gffeb_norm_numerator
+  public :: gffeb_norm_denominator
+  public :: gffeb_norm
+  public :: gffeb_quadratic
+  public :: assign_patch_to_bins
```
ckoven: camel case again?

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,653 @@
...
+          area_forest_patches = area_forest_patches + currentPatch%area
+       end if
+
+       currentPatch => currentPatch%older
+    enddo
+
+  end subroutine get_number_of_forest_patches
+
+
+  function get_number_of_patches(site) result(n_patches)
```
ckoven: I'm slightly shocked that we don't already have a procedure for this, I suggest putting it somewhere more generic than this module?

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,653 @@
...
+       currentPatch => currentPatch%older
+    enddo
+
+  end function get_number_of_patches
+
+
+  subroutine rank_forest_edge_proximity(site, indices, index_forestpatches_to_allpatches)
+    ! DESCRIPTION:
+    ! Rank forest patches by their proximity to edge.
+    !
```
ckoven: I think it'd be useful to note here that patch age is being used as a proxy for edge distance in the rank ordering currently.

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,653 @@
...
+    end if
+
+    ! Allocate arrays
+    allocate(array(1:n_forest_patches))
+
+    ! Fill arrays
+    f = 0
+    p = 0
+    index_forestpatches_to_allpatches(:) = 0
+    currentPatch => site%oldest_patch
```
ckoven, line 134: there should be probably be a pair of warnings both here and in `assign_patches_to_bins ` that the order of patch looping (oldest to youngest) needs to match in the two cases.

ckoven (approved):
@samsrabin this looks great to me, thanks for doing it! I have a couple comments, all minor stuff.

main/EDInitMod.F90
samsrabin: Resolved with 233e52505.

main/FatesEcotypesMod.F90
samsrabin: Resolved with 7ffd6f0e5 (assuming you meant PascalCase).

main/FatesEcotypesMod.F90
samsrabin: Yeah, grass biomass was just an idea I had for another constraint. If you still want me to explain the logic, let me know; otherwise please resolve this comment.

main/FatesEdgeForestMod.F90
samsrabin: Resolved with 7ffd6f0e5.

main/FatesEdgeForestMod.F90
samsrabin: Issue opened at #1481, so I'll resolve this comment.

main/FatesEdgeForestMod.F90
samsrabin: Moved to `ed_site_type` in 4ac9ac1bdef.

main/FatesEdgeForestMod.F90
samsrabin: Done with a8c0c358c.

main/FatesEdgeForestMod.F90
samsrabin: I don't think that's true, actually. The ordering of the loop in `AssignPatchesToBins()` shouldn't matter, because each forest patch should have a unique value in `index_forestpatches_to_allpatches` (as tested in `indexx_lowTie()`, `indexx_highTie()`, and `indexx_all_equal()`.

testing/functional_testing/edge_forest/edge_forest_test.py
samsrabin: Resolved with 1476eb3ba.

testing/run_unit_tests.py
samsrabin: Resolved with 131d93a999.

main/EDParamsMod.F90
```diff
@@ -350,6 +354,21 @@ subroutine FatesParamsInit()
 
   end subroutine FatesParamsInit
 
+  !-----------------------------------------------------------------------
+
+  subroutine FatesParamsInitForFactory()
+    ! Initialize some parameters that are needed for unit-testing factories
+
+    allocate(ED_val_history_ageclass_bin_edges(7))
```
rgknox, line 362: @samsrabin these are also still allocated here: https://github.com/samsrabin/fates/blob/edge-area-202410/main/EDParamsMod.F90#L119

Do we need to remove those statements?

parameter_files/fates_params_default.cdl
```diff
@@ -258,6 +259,36 @@ variables:
 	double fates_dev_arbitrary_pft(fates_pft) ;
 		fates_dev_arbitrary_pft:units = "unknown" ;
 		fates_dev_arbitrary_pft:long_name = "Unassociated pft dimensioned free parameter that developers can use for testing arbitrary new hypotheses" ;
+	double fates_edgeforest_bin_edges(fates_edgeforest_bins) ;
```
rgknox, line 262: group name-spacing (edgeforest), nice

rgknox: extra points for inserting it in alphabetical order!

main/FatesEcotypesMod.F90
rgknox: ok, sounds good

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,630 @@
...
+    integer, intent(out) :: n_forest_patches
+    real(r8), intent(out) :: area_forest_patches
+    real(r8), intent(out) :: area_site
+    !
+    ! LOCAL VARIABLES:
+    type(fates_patch_type), pointer :: currentPatch
+
+    n_forest_patches = 0
+    area_forest_patches = 0._r8
+    area_site = 0._r8
```
rgknox, line 90: @samsrabin , since there is no filtering of the the patch loop (ie it traverses all patches from youngest to oldest), and area_site is not inside any logic filters, shouldn't area_site always equal 10k (ie "AREA")?  Was this not expected or was this a check?

main/EDParamsMod.F90
samsrabin: No, the `FatesParamsInitForFactory()` subroutine is a special one that's only called during unit testing, when the other allocation code is skipped.

biogeochem/FatesPatchMod.F90
```diff
@@ -453,6 +461,10 @@ subroutine NanValues(this)
       this%ncl_p                        = fates_unset_int
       this%land_use_label               = fates_unset_int
       this%age_since_anthro_disturbance = nan
+
+      ! FOREST INFO
+      this%is_forest                       = .false.
```
mpaiao, line 466: I haven't seen many structure variables in FATES declared as `logical`. But I actually much prefer it, I think it would make the code a bit neater than using `if (var == itrue)` or `if(var == ifalse)` constructions...

main/EDParamsMod.F90
```diff
@@ -350,6 +354,21 @@ subroutine FatesParamsInit()
...
+    ED_val_history_ageclass_bin_edges = [0, 1, 2, 5, 10, 20, 50]
+
+    allocate(ED_val_history_sizeclass_bin_edges(13))
+    ED_val_history_sizeclass_bin_edges = [0, 5, 10, 15, 20, 30, 40, 50, 60, 70, 80, 90, 100]
+
+    allocate(ED_val_history_coageclass_bin_edges(13))
+    ED_val_history_coageclass_bin_edges = [0, 5]
```
mpaiao, line 369: I guess either the following:
```suggestion
    allocate(ED_val_history_coageclass_bin_edges(2))
    ED_val_history_coageclass_bin_edges = [0, 5]
```
or define all the 13 elements?

main/FatesEcotypesMod.F90
```diff
@@ -0,0 +1,81 @@
...
+  !
+  ! ARGUMENTS:
+  type(fates_patch_type), intent(in), pointer :: patchptr  ! pointer to patch object
+  real(r8), intent(in) :: grass_biomass_threshold ! Live grass biomass (kgC/m2) above which a patch is considered to "have grass"
+  !
+  ! RETURN VALUE
+  logical :: DoesPatchHaveForest_GrassBiomass
+
+  DoesPatchHaveForest_GrassBiomass = patchptr%livegrass > grass_biomass_threshold
+
```
mpaiao, line 57: I personally prefer to have the commands inside a function to be indented, I think it makes it easier to read.

main/FatesEcotypesMod.F90
```diff
@@ -0,0 +1,81 @@
...
+  real(r8), intent(in), optional :: grass_biomass_threshold ! Live grass biomass (kgC/m2) above which a patch is considered to "have grass"
+  !
+  ! RETURN VALUE
+  logical :: IsPatchForest
+
+  IsPatchForest = DoesPatchHaveForest_TreeCover(patchptr, forest_tree_fraction_threshold)
+  if (IsPatchForest .and. present(grass_biomass_threshold)) then
+     IsPatchForest = .not. DoesPatchHaveForest_GrassBiomass(patchptr, grass_biomass_threshold)
+  end if
+
```
mpaiao, line 77: Same thing.

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,630 @@
...
+    end do patchloop
+
+    ! Get indices of sorted forest patches
+    call indexx(array, indices)
+
+    ! Clean up
+    deallocate(array)
+  end subroutine RankForestEdgeProximity
+
+  function GetFracEdgeForestInEachBin_norm_numerator(x_in, A, mu, sigma, lognorm)
```
mpaiao, line 159: I think the current code is fine, but I wondered why there was a need for one function for the numerator and another one for the denominator...

main/FatesEdgeForestMod.F90
```diff
@@ -0,0 +1,630 @@
...
+          write(fates_log(),*) "ERROR: bin fractions don't sum to 1;    actual minus expected = ",err_chk
+          call endrun(msg=errMsg(__FILE__, __LINE__))
+       end if
+    end if
+
+
+  end subroutine GetFracEdgeForestInEachBin
+
+
+  subroutine AssignPatchToBins(fraction_forest_in_each_bin, area_forest_patches, patch_area, nlevedgeforest, tol, sum_forest_bins_so_far_m2, area_in_edgeforest_bins_m2)
```
mpaiao, line 329: I see the differences between subroutines `AssignPatchToBins` and `AssignPatchesToBins`, but it is a bit confusing. Perhaps rename one of them to something a bit more unique, for example, rename the latter to `DistributePatchesAcrossBins` or something like this?

main/FatesHistoryInterfaceMod.F90
```diff
@@ -4866,6 +4956,10 @@ subroutine update_history_dyn_subsite_ageclass(this,nc,nsites,sites)
                   + cpatch%zstar * patch_area_div_site_area
           end if
 
+          ! whether patch is forest according to FATES parameter file threshold
+          hio_is_forest_si_age(io_si,cpatch%age_class) = hio_is_forest_si_age(io_si,cpatch%age_class) + &
+               merge(1._r8, 0._r8, cpatch%is_forest) * patch_area_div_site_area
```
mpaiao, line 4961: Perhaps use the new function `logical_to_real` you created in here?

main/FatesUtilsMod.F90
```diff
@@ -299,5 +302,24 @@ subroutine QuadraticRootsSridharachary(a,b,c,root1,root2,err)
...
+
+     real(r8), intent(in) :: param
+
+     logical :: is_param_set
+
+     is_param_set = .not. isnan(param)
+     if (is_param_set) then
+        is_param_set = param < fates_check_param_set
+     end if
+  end function is_param_set
```
mpaiao, line 316: Nice to have this as a function!

PR #1477, opened 2025-09-30
Refactor boundary condition interface structure 

glemieux:
Initial refactor of the FATES interface design

### Description:

This pull request refactors the HLM-FATES interface design as an initial step towards enabling "dynamic multi-column" FATES runs for NGEE-Arctic cross cut task 4.  In this regard, this redesign is intended to address design assumptions in which all FATES patches on a given site are  associated with a single column.  Given that this redesign will touch on a large amount of code (on both sides), this PR is intended as a minimum-viable case that introduces the new design setup and only refactors a subset of boundary condition variables to prove the concept. 

Given the scope of the undertaking, the refactor is also used as an opportunity to introduce a standard method of  passing variable data across the HLM-FATES interface and introducing a common vocabulary for associated variables.  The intent is to minimize the amount of necessary shared code between FATES and the HLM. 

This pull request must be coordinated with E3SM-Project/e3sm#7482 and ESCOMP/ctsm#3705.

For more details, wee NGEET/fates-users-guide#109 for updates to the developer's guide section detail the interface design document (_work in progress_)

### Collaborators: 
@rgknox @ckoven 

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


main/FatesInterfaceTypesMod.F90
```diff
@@ -834,13 +862,106 @@ module FatesInterfaceTypesMod
...
+
+    ! Is registry have a FATES patch that exists?
+    logical, private :: active
+
+    ! Container array of interface variables indexed by key
+    type(fates_interface_variable_type), allocatable :: hlm_vars(:) 
+    type(fates_interface_variable_type), allocatable :: fates_vars(:) 
+    
+    ! Array of keys associated with the interface variables
+    character(len=48), allocatable, private :: key(:)
```
rgknox, line 867: bump this up in case we want long keys in the future?

main/FatesInterfaceTypesMod.F90
```diff
@@ -834,13 +862,106 @@ module FatesInterfaceTypesMod
 
    end type bc_pconst_type
 
+   ! Base type to be extended for the API registry
+   type, public :: fates_interface_registry_type
+
+    ! Is registry have a FATES patch that exists?
+    logical, private :: active
+
+    ! Container array of interface variables indexed by key
```
rgknox, line 862: Do the indices of hlm_vars and fates_vars have correspondence with each other? If so, lets note that here in the description

main/FatesInterfaceTypesMod.F90
```diff
@@ -834,13 +862,106 @@ module FatesInterfaceTypesMod
...
+    
+    ! Array of keys associated with the interface variables
+    character(len=48), allocatable, private :: key(:)
+
+    ! Variable regsitry metadata
+    integer :: num_api_vars                       ! number of variables in the registry
+    integer :: num_api_vars_update_init_dims      ! number of variables dimensions needed during initialization
+    integer :: num_api_vars_update_init           ! number of variables that update only at initialization
+    integer :: num_api_vars_update_daily          ! number of variables that update daily
+    integer :: num_api_vars_update_timestep       ! number of variables that update on the model timestep
```
rgknox, line 874: We have variables that need to update at different times during the model time-step.  If these transfers need to happen at different times, how can we group them into one "timestep" group?

main/EDMainMod.F90
```diff
@@ -779,13 +779,17 @@ subroutine ed_integrate_state_variables(currentSite, bc_in, bc_out )
 
     currentPatch => currentSite%youngest_patch
     do while(associated(currentPatch))
+
+       if (currentPatch%nocomp_pft_label .ne. nocomp_bareground) then
```
rgknox: Was this needed for consistency, speed, removing spurious behavior?

main/FatesInterfaceMod.F90
```diff
@@ -2208,6 +2254,115 @@ end subroutine set_fates_ctrlparms
...
+      ! Set the active registry counter and filter index iterator to zero
+      this%num_active_patches = 0
+      i = 0
+
+      ! Loop over sites and patches to set active registries and iterate the counter
+      do s = 1, this%nsites
+         currentPatch => this%sites(s)%oldest_patch
+         do while (associated(currentPatch))
+
+            if (currentPatch%nocomp_pft_label .ne. nocomp_bareground) then
```
rgknox, line 2305: I think one of our plans was to get rid of these conditionals by implementing that refactor that removes bareground patches before this, right? Are we punting on that?

main/FatesInterfaceMod.F90
```diff
@@ -2676,7 +2831,372 @@ subroutine FatesReadParameters(param_reader)
...
+   ! Initialize the current gridcell index and the fates site counter
+   gc = fates_unset_int
+   s = 0
+   ifp = 0
+   
+   ! Iterate over the number of vegetated patches and determine 
+   do r = 1, this%npatches
+      
+      ! Get the gridcell index
+      g = this%registry(r)%GetGridcellIndex()
```
rgknox, line 2900: I think technically, we match our fates sites with land-units now right? ELM could have multiple topo-units, each with a natural vegetation land-unit that we would want to match with.  Matching at the LU level in CLM would also work.

main/FatesInterfaceVarTypeMod.F90
```diff
@@ -0,0 +1,605 @@
...
+      this%subgrid_type = subgrid_type
+      this%conversion_factor = conversion_factor
+
+    end subroutine RegisterInterfaceVariable_2d
+
+  ! ====================================================================================
+
+    subroutine SetLastState(this, last_state)
+    
+      ! This procedure sets the last state logical metadata for the calling interface variable
```
ckoven, line 394: I'm not following (yet) how or why the `is_last` logic is used in practice. maybe some further details on that would be helpful here.

main/FatesInterfaceTypesMod.F90
```diff
@@ -852,4 +973,1046 @@ subroutine ZeroBCOutCarbonFluxes(bc_out)
...
+          key == hlm_fates_litter_carbon_labile .or. &
+          key == hlm_fates_litter_carbon_lignin .or. &
+          key == hlm_fates_litter_nitrogen_cellulose .or. &
+          key == hlm_fates_litter_nitrogen_labile .or. &
+          key == hlm_fates_litter_nitrogen_lignin .or. &
+          key == hlm_fates_litter_phosphorus_cellulose .or. &
+          key == hlm_fates_litter_phosphorus_labile .or. &
+          key == hlm_fates_litter_phosphorus_lignin) then
+            this%num_api_vars_litter_flux= this%num_api_vars_litter_flux + 1
+      end if
```
ckoven, line 1336: I guess an alternative, which might be easier to maintain, would be to make an 'is_litter_flux' or something like that boolean flag for all interface variables rather than enumerating all the variables here?

main/FatesInterfaceTypesMod.F90
```diff
@@ -852,4 +973,1046 @@ subroutine ZeroBCOutCarbonFluxes(bc_out)
...
+          this%key(index) == hlm_fates_litter_phosphorus_cellulose .or. &
+          this%key(index) == hlm_fates_litter_carbon_labile .or. &
+          this%key(index) == hlm_fates_litter_nitrogen_labile .or. &
+          this%key(index) == hlm_fates_litter_phosphorus_labile .or. &
+          this%key(index) == hlm_fates_litter_carbon_lignin .or. &
+          this%key(index) == hlm_fates_litter_nitrogen_lignin .or. &
+          this%key(index) == hlm_fates_litter_phosphorus_lignin) then
+            count_litter_flux = count_litter_flux + 1
+            this%filter_litter_flux(count_litter_flux) = index
+      end if
```
ckoven, line 1606: similar comment to the above -- having a boolean flag attached to the variables might be easier to maintain than enumerating them here?

main/FatesInterfaceTypesMod.F90
```diff
@@ -2,11 +2,15 @@ module FatesInterfaceTypesMod
...
   use FatesConstantsMod   , only : itrue,ifalse
+  use FatesConstantsMod   , only : fates_unset_int
   use FatesGlobals        , only : fates_global_verbose
   use FatesGlobals        , only : fates_log
   use FatesGlobals        , only : endrun => fates_endrun
   use shr_log_mod         , only : errMsg => shr_log_errMsg
   use shr_infnan_mod      , only : nan => shr_infnan_nan, assignment(=)
+  use PRTGenericMod       , only : prt_cnp_flex_allom_hyp
+  use FatesInterfaceVariableTypeMod, only : fates_interface_variable_type
+  use FatesInterfaceParametersMod
```
ckoven, line 13: I think I agree that it makes sense to do an unlimited `use` statement rather than enumerating every possible interface variable name in the list of `only` variables, which might get quite long, but I guess the risk is that a reader of the code sees all the string indices in `DefineInterfaceRegistry` and it isn't obvious where those all come from. So I wonder if it would make sense to only have the open-ended `use` statement in that subroutine and restrict it to `use only` the integer parameters throughout the rest of the module?

ckoven: an alternative might be to give the string keys names something that is more obvious what they are, like `fates_interface_var_longname`?

main/FatesInterfaceMod.F90
```diff
@@ -2676,7 +2831,372 @@ subroutine FatesReadParameters(param_reader)
...
+! ======================================================================================
+
+subroutine InitializeBoundaryConditions(this, patches_per_site)
+
+   ! This procedure initializes the boundary conditions arrays with the maximum number 
+   ! of possible patches for each FATES site.  It then registers the boundary condition
+   ! variables for each patch.  Conversion factors are applied as necessary during 
+   ! registration, which will be utilized during the update API calls.
+   
+   use FatesInterfaceParametersMod
```
ckoven, line 2952: I like how the open-ended `use` here is limited to the subroutine

ckoven (commented):
Thanks @glemieux, this all looks good to me, only minor comments on first reading. Will continue to think some more about how this all works.

main/EDMainMod.F90
glemieux: This was for spurious behavior, but might actually be defunct code that should be removed per recent development.  I'll double check this.

main/FatesInterfaceMod.F90
glemieux: For now.  We need this in place until we remove the code that instantiates a fates bareground patch, which is a more significant refactor.  That said, we'll need to deal with it during the future multi-column fates pull request.

glemieux: I'll add a comment above the check noting this.

glemieux, line 2305:
```suggestion
            ! This check is necessary until FATES no longer instantiates a bareground patch
            if (currentPatch%nocomp_pft_label .ne. nocomp_bareground) then
```

rgknox:
@glemieux , can you confirm this is intended B4B?

glemieux:
> @glemieux , can you confirm this is intended B4B?

That's correct @rgknox.  It's currently near-B4B due to changes to math order of operations changes resulting in very small DIFFs in only a subset of variables.  I've added the near-B4B label.

PR #1493, opened 2025-10-20
JSON based parameters

rgknox:
### Description:

This set of changes migrates FATES away from using a netcdf based parameter file.  The file format is changed to JSON (or XML depending on discussion). A fortran JSON parsing utility is provided (not external module), which utilizes a character-by-character file scan and indexing functions. Python utilities are planned to be converted from netcdf to json based as well.

A new utility is available, that converts pre-existing netcdf parameter datasets to the json (or xml) format.

Synchronized with: CTSM [3570](https://github.com/ESCOMP/CTSM/pull/3570)

### Collaborators: 

@glemieux 

### To Do:

- [x] Write python conversion utility that transforms pre-existing cdl files to json/xml
- [x] Generate new default file
- [x] Write Fortran JSON/XML parser (JSON parser feature complete)
- [x] Write code that transfers new parameter data-structure into FATES primitive parameter memory
- [x] Refactor: tools/modify_fates_paramfile.py
  - Include new commands appended to history
  - Include a change by name feature 
- [x] Refactor: tools/FatesPFTIndexSwapper.py
  - Include a change by name feature
- [x] Refactor: tools/ncvarsort.py
  - Change name to sort_parameters.py
- [x] Remove temporary file datatypes.txt
- [x] Refactor: tools/BatchPatchParams.py
- [x] Remove: tools/UpdateParamAPI.py
- [x] Add parameter destroy method in the fortran code (ie deallocation)
- [x] Convert existing unit tests to use JSON
- [x] Updated documentation: https://github.com/NGEET/fates-users-guide/pull/104, https://fates-users-guide--104.org.readthedocs.build/en/104/user/Parameter-File.html
- [x] HLM PR:
   - Update default file path to be fates/parameters/fates_parameter_default.* instead of the binary
   - Update the HLM fortran code that calls FATES parameter reads
   - Update scripting in cime_config/testdefs/testmods_dirs/clm
   - Update at least 1 test configuration to create/use a modified version of the new file type in CASEDIR or scratch

This should not have any meaningful changes, but reading in the parameters differently could have very small, inconsequential changes to answers (barely outside of roundoff diffs).

### Checklist

*Contributor*
- [x] Python manipulation tools completed

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -76,15 +76,14 @@ module EDCanopyStructureMod
   character(len=255) :: smsg ! Message string for deallocation errors
 
   ! Precision targets for demotion and promotion
-  ! We have two:
-  ! "pa_area_target_precision" is the required precision at the patch level,
-  !    we keep shuffling and splitting cohorts until each layer is within this precision
   ! "co_area_target_precision" is the required precision at the cohort level,
   !    essentially it is the minimum amount of change required to not ignore
-  !    a partial promotion or demotion
+  !    a partial promotion or demotion. This number was chosen because
+  !    math precision is around 15 digits in fortran r8s, with numbers that
+  !    can get to magnitude e4, this gives us about two orders of magitude in math
+  !    precision (ie e15-(e4+e9)=e2) in significant digits to match this absolute precision
   
-  real(r8), parameter :: pa_area_target_precision = 1.0E-11_r8
-  real(r8), parameter :: co_area_target_precision = 1.0E-12_r8 
+  real(r8), parameter :: co_area_target_precision = 1.0E-9_r8 
 
   integer, parameter :: demotion_phase  = 1
   integer, parameter :: promotion_phase = 2
@@ -94,8 +93,6 @@ module EDCanopyStructureMod
 
   real(r8), parameter :: area_check_precision  = 1.0E-7_r8     ! Area conservation checks must
   ! be within this absolute tolerance
-  real(r8), parameter :: area_check_rel_precision = 1.0E-4_r8  ! Area conservation checks must
-  ! be within this relative tolerance
 
   real(r8), parameter :: similar_height_tol = 1.0E-3_r8    ! I think trees that differ by 1mm
   ! can be roughly considered the same right?
@@ -169,7 +166,7 @@ contains
     integer  :: i_lyr                  ! current layer index
     integer  :: z                      ! Current number of canopy layers. (1= canopy, 2 = understorey)
     integer  :: ipft
-    real(r8) :: arealayer(nclmax+5)    ! Amount of plant area currently in each canopy layer
+    real(r8) :: arealayer              ! Amount of plant area currently in each canopy layer
     integer  :: patch_area_counter     ! count iterations used to solve canopy areas
     logical  :: area_not_balanced      ! logical controlling if the patch layer areas
     real(r8) :: target_area            ! Canopy area that is either in excess/defiency
@@ -182,7 +179,7 @@ contains
     ! try to re-balance 3 times.  If that doesn't give layer areas
     ! within tolerance of canopy area, there is something wrong
 
-    integer, parameter  :: max_patch_iterations = 10
+    integer, parameter  :: max_patch_iterations = nclmax + 7
 
 
     !----------------------------------------------------------------------
@@ -216,6 +213,7 @@ contains
        ! canopy layer has a special bounds check
        currentCohort => currentPatch%tallest
        do while (associated(currentCohort))
+          currentCohort%canopy_layer_yesterday = currentCohort%canopy_layer
           if( currentCohort%canopy_layer < 1 ) then
              write(fates_log(),*) 'lat:',currentSite%lat
              write(fates_log(),*) 'lon:',currentSite%lon
@@ -243,11 +241,13 @@ contains
           z = NumCanopyLayers(currentPatch)
 
           do i_lyr = 1,z ! Loop around the currently occupied canopy layers.
-             call CanopyLayerArea(currentPatch,currentSite%spread,i_lyr,arealayer(i_lyr))
-             target_area = max(0._r8,arealayer(i_lyr) - (1._r8-imperfect_fraction)*currentPatch%area)
+             call CanopyLayerArea(currentPatch,currentSite%spread,i_lyr,arealayer)
+             target_area = max(0._r8,arealayer - (1._r8-imperfect_fraction)*currentPatch%area)
              call PromoteOrDemote(currentSite, currentPatch, i_lyr, demotion_phase, target_area)
           end do
 
+          ! Terminate only for type 1 (near zero number density)
+          call terminate_cohorts(currentSite, currentPatch,1,23,bc_in)
           call fuse_cohorts(currentSite, currentPatch, bc_in)
 
           ! ---------------------------------------------------------------------------------------
@@ -261,11 +261,13 @@ contains
           ! We only promote if we have at least two layers
           if (z>1) then
              do i_lyr=2,z
-                call CanopyLayerArea(currentPatch,currentSite%spread,i_lyr-1,arealayer(i_lyr-1))
-                target_area = max(0._r8,(1._r8-imperfect_fraction)*currentPatch%area - arealayer(i_lyr-1))
+                call CanopyLayerArea(currentPatch,currentSite%spread,i_lyr-1,arealayer)
+                target_area = max(0._r8,(1._r8-imperfect_fraction)*currentPatch%area - arealayer)
                 call PromoteOrDemote(currentSite, currentPatch, i_lyr, promotion_phase, target_area)
              end do
 
+             ! Terminate only for type 1 (near zero number density)
+             call terminate_cohorts(currentSite, currentPatch,1,24,bc_in)
              call fuse_cohorts(currentSite, currentPatch, bc_in)
 
           end if
@@ -281,14 +283,14 @@ contains
 
           z = NumCanopyLayers(currentPatch)
           area_not_balanced = .false.
-          do i_lyr = 1,z
-             call CanopyLayerArea(currentPatch,currentSite%spread,i_lyr,arealayer(i_lyr))
+          do i_lyr = 1,min(z,nclmax)
+             call CanopyLayerArea(currentPatch,currentSite%spread,i_lyr,arealayer)
              if(i_lyr < z)then
-                if (abs(arealayer(i_lyr)-(1._r8-imperfect_fraction)*currentPatch%area) > area_check_precision) then
+                if (abs(arealayer-(1._r8-imperfect_fraction)*currentPatch%area) > area_check_precision) then
                    area_not_balanced = .true.
                 end if
              else
-                if ((arealayer(i_lyr)-(1._r8-imperfect_fraction)*currentPatch%area) > area_check_precision) then
+                if ((arealayer-(1._r8-imperfect_fraction)*currentPatch%area) > area_check_precision) then
                    area_not_balanced = .true.
                 end if
              end if
@@ -308,8 +310,9 @@ contains
              write(fates_log(),*) 'spread:',currentSite%spread
              do i_lyr = 1,z
                 write(fates_log(),*) '-----------------------------------------'
-                write(fates_log(),*) 'layer: ',i_lyr,' area: ',arealayer(i_lyr)
-                write(fates_log(),*) 'bias [m2] (layer-patch): ',(arealayer(i_lyr)- &
+                call CanopyLayerArea(currentPatch,currentSite%spread,i_lyr,arealayer)
+                write(fates_log(),*) 'layer: ',i_lyr,' area: ',arealayer
+                write(fates_log(),*) 'bias [m2] (layer-patch): ',(arealayer - &
                      (1._r8-imperfect_fraction)*currentPatch%area)
                 currentCohort => currentPatch%tallest
                 do while (associated(currentCohort))
@@ -590,22 +593,10 @@ contains
          end do
       end if comp_excl_type
 
-      ! Check to make sure the changes are within bounds
-      do ic = 1,n_layer
-         cohort => layer_co(ic)%p
-         if( ((layer_co(ic)%pd_area - cohort%c_area) > co_area_target_precision ) .or. &
-              (layer_co(ic)%pd_area < 0._r8) ) then
-            write(fates_log(),*) 'negative,or more area than the cohort has is being promoted/demoted'
-            write(fates_log(),*) 'change: ',layer_co(ic)%pd_area
-            write(fates_log(),*) 'existing area:',cohort%c_area
-            write(fates_log(),*) 'excess: ',layer_co(ic)%pd_area - cohort%c_area
-            call endrun(msg=errMsg(sourcefile, __LINE__))
-         end if
-      end do
-
       ! Part 3:
       ! Apply the area changes by splitting the cohort and re-assigning
       ! either all or part of it to a new layer
+      ! Check to make sure the changes are within bounds
 
       ic_loop0: do ic = 1,n_layer
 
@@ -617,25 +608,30 @@ contains
          !    and not trivialy small (larger than precision
          !    check), then split it and move part of it
          ! If the dem/prom area is less than zero or larger than
-         !    the cohort area within precision checks then
-         !    we would have failed in the previous checks
+         !    the cohort area within precision checks then fail
+         
+         
+         whole_or_part: if( ((layer_co(ic)%pd_area - cohort%c_area) > co_area_target_precision ) .or. &
+              (layer_co(ic)%pd_area < 0._r8) ) then
+            write(fates_log(),*) 'negative,or more area than the cohort has is being promoted/demoted'
+            write(fates_log(),*) 'change: ',layer_co(ic)%pd_area
+            write(fates_log(),*) 'existing area:',cohort%c_area
+            write(fates_log(),*) 'excess: ',layer_co(ic)%pd_area - cohort%c_area
+            call endrun(msg=errMsg(sourcefile, __LINE__))
 
-         whole_or_part: if ( abs(layer_co(ic)%pd_area - cohort%c_area) < &
-              co_area_target_precision ) then
+         
+         elseif ( abs(layer_co(ic)%pd_area - cohort%c_area) < co_area_target_precision ) then
 
             ! Whole cohort promotion/demotion
             cohort%canopy_layer = cohort%canopy_layer + ilyr_change
-
-         elseif( (layer_co(ic)%pd_area < cohort%c_area) .and. &
-                 (layer_co(ic)%pd_area > 0 ) ) then
+            
+         elseif( layer_co(ic)%pd_area > 0._r8 ) then
 
             ! Partial cohort promotion/demotion
-
             ! Make a copy of the current cohort.  The copy and the original
             ! conserve total number density.  The copy
             ! remains in the upper-story.  The original is the one
             ! demoted to the understory
-
 
             allocate(copyc)
 
@@ -660,7 +656,7 @@ contains
             call copyc%InitPRTBoundaryConditions()
 
             remainder_area = cohort%c_area - layer_co(ic)%pd_area
-            copyc%n = cohort%n*remainder_area/cohort%c_area
+            copyc%n = cohort%n*min(1._r8,max(0._r8,remainder_area/cohort%c_area))
             cohort%n = cohort%n - copyc%n
 
             ! The copied cohort is the part that remains in-layer
@@ -692,24 +688,26 @@ contains
          ! Part 4:
          ! keep track of number and biomass promoted/demoted
 
-         leaf_c   = cohort%prt%GetState(leaf_organ,carbon12_element)
-         store_c  = cohort%prt%GetState(store_organ,carbon12_element)
-         fnrt_c   = cohort%prt%GetState(fnrt_organ,carbon12_element)
-         sapw_c   = cohort%prt%GetState(sapw_organ,carbon12_element)
-         struct_c = cohort%prt%GetState(struct_organ,carbon12_element)
-
-         if(phase==demotion_phase) then
-            site%demotion_rate(cohort%size_class) = &
-                 site%demotion_rate(cohort%size_class) + cohort%n
-            site%demotion_carbonflux = site%demotion_carbonflux + &
-                 (leaf_c + store_c + fnrt_c + sapw_c + struct_c) * cohort%n
-         else
-            site%promotion_rate(cohort%size_class) = &
-                 site%promotion_rate(cohort%size_class) + cohort%n
-            site%promotion_carbonflux = site%promotion_carbonflux + &
-                 (leaf_c + store_c + fnrt_c + sapw_c + struct_c) * cohort%n
+         if( layer_co(ic)%pd_area > 0._r8 ) then
+            leaf_c   = cohort%prt%GetState(leaf_organ,carbon12_element)
+            store_c  = cohort%prt%GetState(store_organ,carbon12_element)
+            fnrt_c   = cohort%prt%GetState(fnrt_organ,carbon12_element)
+            sapw_c   = cohort%prt%GetState(sapw_organ,carbon12_element)
+            struct_c = cohort%prt%GetState(struct_organ,carbon12_element)
+            
+            if(phase==demotion_phase) then
+               site%demotion_rate(cohort%size_class) = &
+                    site%demotion_rate(cohort%size_class) + cohort%n
+               site%demotion_carbonflux = site%demotion_carbonflux + &
+                    (leaf_c + store_c + fnrt_c + sapw_c + struct_c) * cohort%n
+            else
+               site%promotion_rate(cohort%size_class) = &
+                    site%promotion_rate(cohort%size_class) + cohort%n
+               site%promotion_carbonflux = site%promotion_carbonflux + &
+                    (leaf_c + store_c + fnrt_c + sapw_c + struct_c) * cohort%n
+            end if
          end if
-
+            
       end do ic_loop0
 
     end associate
--- a/biogeochem/EDPatchDynamicsMod.F90
+++ b/biogeochem/EDPatchDynamicsMod.F90
@@ -3,7 +3,7 @@ module EDPatchDynamicsMod
   ! Controls formation, creation, fusing and termination of patch level processes. 
   ! ============================================================================
   use FatesGlobals         , only : fates_log
-  use FatesGlobals         , only : FatesWarn,N2S,A2S
+  use FatesGlobals         , only : FatesWarn,N2S,A2S,I2S
   use FatesInterfaceTypesMod, only : hlm_freq_day
   use FatesInterfaceTypesMod, only : hlm_current_tod
   use EDPftvarcon          , only : EDPftvarcon_inst
@@ -3381,6 +3381,7 @@ contains
     real(r8) areatot ! variable for checking whether the total patch area is wrong.
     real(r8) :: state_vector_driver(n_landuse_cats)  ! [m2/m2]
     real(r8) :: state_vector_internal(n_landuse_cats)  ! [m2/m2]
+    character(len=1024) :: warn_msg   ! for defining a warning message
     !---------------------------------------------------------------------
 
     ! Initialize the count cycles
@@ -3411,8 +3412,12 @@ contains
 
              if ( .not. gotfused ) then
                 !! somehow didn't find a patch to fuse with.
-                write(fates_log(),*) 'Warning. small nocomp patch wasnt able to find another patch to fuse with.', &
-                     currentPatch%nocomp_pft_label, currentPatch%land_use_label, currentPatch%area
+                warn_msg = 'small nocomp patch wasnt able to find '// &
+                           'another patch to fuse with. '// &
+                           'nocomp pft: '//trim(I2S(currentPatch%nocomp_pft_label))// &
+                           'lu label: '//trim(I2S(currentPatch%land_use_label))// &
+                           'area: '//trim(N2S(currentPatch%area))
+                call FatesWarn(warn_msg,index=5)
              endif
 
           else nocomp_if
--- a/biogeochem/FatesAllometryMod.F90
+++ b/biogeochem/FatesAllometryMod.F90
@@ -748,7 +748,8 @@ contains
           tree_lai = (log(exp(-1.0_r8 * kn * canopy_lai_above) - &
                kn * slat * leafc_per_unitarea) + &
                (kn * canopy_lai_above)) / (-1.0_r8 * kn)
-
+          ! precision errors in the above when leafc_per_unit_area is tiny can make treelai negative
+          tree_lai = max(0.0_r8,tree_lai)
           ! If leafc_per_unitarea becomes too large, tree_lai becomes an imaginary number 
           ! (because the tree_lai equation requires us to take the natural log of something >0)
           ! Thus, we include the following error message in case leafc_per_unitarea becomes too large.
@@ -774,7 +775,8 @@ contains
                kn * slat * leafc_slamax) + &
                (kn * canopy_lai_above)) / (-1.0_r8 * kn)) + &
                (leafc_per_unitarea - leafc_slamax) * sla_max
-
+          ! precision errors in the above when leafc_per_unit_area is tiny can make treelai negative
+          tree_lai = max(0.0_r8,tree_lai)
           ! if leafc_slamax becomes too large, tree_lai_exp becomes an imaginary number 
           ! (because the tree_lai equation requires us to take the natural log of something >0)
           ! Thus, we include the following error message in case leafc_slamax becomes too large.
--- a/biogeochem/FatesCohortMod.F90
+++ b/biogeochem/FatesCohortMod.F90
@@ -262,7 +262,6 @@ module FatesCohortMod
     real(r8) :: dndt      ! time derivative of cohort size [n/year]
     real(r8) :: dhdt      ! time derivative of height [m/year]
     real(r8) :: ddbhdt    ! time derivative of dbh [cm/year]
-    real(r8) :: dbdeaddt  ! time derivative of dead biomass [kgC/year]
 
     !---------------------------------------------------------------------------
 
@@ -451,7 +450,6 @@ module FatesCohortMod
       this%dndt                    = nan 
       this%dhdt                    = nan 
       this%ddbhdt                  = nan
-      this%dbdeaddt                = nan 
    
       ! FIRE
       this%fraction_crown_burned   = nan 
@@ -794,7 +792,6 @@ module FatesCohortMod
       copyCohort%dndt                    = this%dndt
       copyCohort%dhdt                    = this%dhdt
       copyCohort%ddbhdt                  = this%ddbhdt
-      copyCohort%dbdeaddt                = this%dbdeaddt
 
       ! FIRE
       copyCohort%fraction_crown_burned   = this%fraction_crown_burned
@@ -1140,7 +1137,6 @@ module FatesCohortMod
       write(fates_log(),*) 'cohort%dndt                   = ', this%dndt
       write(fates_log(),*) 'cohort%dhdt                   = ', this%dhdt
       write(fates_log(),*) 'cohort%ddbhdt                 = ', this%ddbhdt
-      write(fates_log(),*) 'cohort%dbdeaddt               = ', this%dbdeaddt
       write(fates_log(),*) 'cohort%fraction_crown_burned  = ', this%fraction_crown_burned
       write(fates_log(),*) 'cohort%fire_mort              = ', this%fire_mort
       write(fates_log(),*) 'cohort%crownfire_mort         = ', this%crownfire_mort
--- a/main/ChecksBalancesMod.F90
+++ b/main/ChecksBalancesMod.F90
@@ -302,7 +302,7 @@ contains
 
             select case(element_list(el))
             case(carbon12_element)
-               net_uptake = diag%npp + site_mass%net_root_uptake*area_inv
+               net_uptake = (site_mass%gpp_acc - site_mass%aresp_acc +  site_mass%net_root_uptake)*area_inv
             case(nitrogen_element)
                net_uptake = site_mass%net_root_uptake*area_inv
             case(phosphorus_element)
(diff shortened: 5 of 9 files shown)


rgknox:
This is passing fates test suit on derecho, B4B with base

tools/datatypes.txt
```diff
@@ -0,0 +1,310 @@
+fates_history_ageclass_bin_edges 1
```
ekluzek: I'm wondering about this dataset. It's really long and detailed and I wonder if it has to correspond to code or data that's encapsulated elsewhere? If so I wonder if there's a way to generate it? It also might be data that doesn't change often so it might not be an issue.

It would also be helpful to allow for comments to give more of an explanation of what's in the file.

parameter_files/patch_default_bciopt224.json
```diff
@@ -0,0 +1,61 @@
...
+    "usage":"This file should be passed as input to the batch patch python script in fates, from the parameter_files directory:, like so: (cd parameter_files;python tools/batch_patch_params.py --fin patch_default_bciopt224.json)",
+    "base_file":"fates_params_default.json",
+    "new_file":"fates_params_opt224.json",
+    "pft_trim_list": [1],
+    "parameters":{
+	"pft_parameters":{
+	    "1":{
```
ekluzek, line 9: I'm guessing this file must be just a small part of what will become the full dataset? So this is a proof of concept to show the structure of the data?

main/FatesInterfaceMod.F90
```diff
@@ -785,16 +791,36 @@ subroutine SetFatesGlobalElements1(use_fates,surf_numpft,surf_numcft,param_reade
...
-         call FatesReadParameters(param_reader)
+         if ( hlm_masterproc == itrue ) then
+            write(fates_log(), *) 'Reading FATES parameters'
+         end if
+
+         ! We compare dataset values that have no-data against a parameter fates_check_param_set
+         ! If the value in the file is less than fates_check_param_set, then we assume
+         ! it is usable. So if the parser enounters no-data strings, such as null, nan or '_'
+         ! we set those values to fates_check_param_set plus a little extra to flag it is
+         ! invalid.
```
rgknox, line 812: Note to namespace these setter calls so people know on the calling side that they are for json file handling

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+          write(log_unit,*) 'fatal i/o error! code:', io_status
+          write(log_unit,*) 'msg: ',io_msg
+          call shr_sys_abort()
+       end if
+       
+       if (scan(filechar,'"')>0)then
+          inside_quotes = .not.inside_quotes
+       end if
+
+       if (scan('{}',filechar)>0 .and. inside_quotes) then
```
glemieux: I'm not familiar with `scan`: Why are the brackets the first argument for `scan` here as opposed to the other way around?  Would this also evaluate as `.true.` if there was a quote inside the brackets as well?

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    
+  end subroutine GetMetaString
+  
+  ! =====================================================================================
+
+  subroutine GetStringVec(string_in,string_vec_out,n_vec_out)
+
+    ! This routine parses a string, does some cleaning of extra charachters
+    ! and generally uses commas as separators to fill in a vector of
+    ! cleaned strings. It is assumed that the sting passed in contains
```
glemieux:
```suggestion
    ! cleaned strings. It is assumed that the string passed in contains
```

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    n_vec_out = l
+
+  end subroutine GetStringVec
+
+  ! =====================================================================================
+  
+  subroutine PopString(apstring,newchar)
+    
+    character(len=*) :: apstring
+    character(len=1) :: newchar
```
glemieux:
```suggestion
    character(len=*), intent(inout) :: apstring
    character(len=1), intent(in) :: newchar
```

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+       
+    end do do_str_scan
+
+    n_vec_out = l
+
+  end subroutine GetStringVec
+
+  ! =====================================================================================
+  
+  subroutine PopString(apstring,newchar)
```
glemieux: What do you think about changing this to be called `ShiftStringAndPushChar` or something like that?  If I'm interpreting it correctly it's shifting the string characters to the left and adding on the new character (dropping the first).  I usually think of a `pop` as removing an element and then returning it.

parameter_files/patch_default_bciopt224.json
glemieux: @ekluzek This is a patch file that is provided as an input to `batch_patch_params.py` to modify the default json parameter file.  I think the intended workflow would be for the user to run `batch_patch_params` with this to create the new json `paramfile` that would be referenced for the user's specific case.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+
+    ! -----------------------------------------------------------------------------------
+    ! Step 3: Parse the dimension data string for dimension data
+    !         Each dimension should have a colon ":".
+    !         Also, scan for the "scalar" parameter
+    !         !!! Just count for now
+    ! -----------------------------------------------------------------------------------
+
+    found_scalar = .false.
+    if( index(trim(dimdata_str(1:dimdata_len)),'"scalar"')>0 ) then
```
glemieux: For my own knowledge, when would scalar be included in the `dimensions` object?  Would it ever not be value `1`?

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    return
+  end subroutine ReadCharVar
+
+  ! =====================================================================================
+  
+  subroutine GetParameters(file_unit,string_scr,pstruct)
+
+    integer,intent(in)                  :: file_unit
+    character(len=max_ll), dimension(*) :: string_scr ! Internal scratch space
+    type(params_type)                   :: pstruct
```
glemieux:
```suggestion
    character(len=max_ll), dimension(*), intent(inout) :: string_scr ! Internal scratch space
    type(params_type), intent(inout)                   :: pstruct
```

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    !
+    ! All parameters must be scalar or associated with a known dimension.
+    !
+    ! Each section is started with an open bracket '{' and closed with a closed bracket '}'
+    ! For each section here we end with a call
+    ! to rewind and go back to the top of the file
+    
+    character(len=*),intent(in) :: filename
+    integer,intent(in)          :: file_unit
+    type(params_type)           :: pstruct
```
glemieux:
```suggestion
    type(params_type), intent(inout)           :: pstruct
```

parameter_files/patch_default_bciopt224.json
ekluzek: Ahhh, I see that makes sense. So this is more or less an example file to show the workflow and how batch_patch_params.py works. Although then where is the default parameter file in JSON format? And if it's just an example file, there should be some comments that explain that in the file.

main/FatesInterfaceMod.F90
```diff
@@ -785,16 +791,36 @@ subroutine SetFatesGlobalElements1(use_fates,surf_numpft,surf_numcft,param_reade
       logical,                    intent(in) :: use_fates    ! Is fates turned on?
       integer,                    intent(in) :: surf_numpft  ! Number of PFTs in surface dataset
       integer,                    intent(in) :: surf_numcft  ! Number of CFTs in surface dataset
-      class(fates_param_reader_type), intent(in) :: param_reader ! HLM-provided param file reader
+      character(len=*),           intent(in) :: paramfile        ! Full path to the fates parameter file
+      integer,                    intent(in) :: paramfile_unit   ! File unit reserved for the FATES
```
ekluzek: Remove this as an input argument, and make it just local to inside of the JSON file read. It doesn't need to be passed to any other level.

main/FatesInterfaceMod.F90
```diff
@@ -785,16 +791,36 @@ subroutine SetFatesGlobalElements1(use_fates,surf_numpft,surf_numcft,param_reade
...
+         end if
+
+         ! We compare dataset values that have no-data against a parameter fates_check_param_set
+         ! If the value in the file is less than fates_check_param_set, then we assume
+         ! it is usable. So if the parser enounters no-data strings, such as null, nan or '_'
+         ! we set those values to fates_check_param_set plus a little extra to flag it is
+         ! invalid.
+         call SetInvalid(fates_check_param_set+10._r8)
+         call SetLogInit(fates_log())
+         call ReadJSON(paramfile,paramfile_unit,pstruct)
```
ekluzek: Remove paramfile_unit from here.

main/FatesInterfaceMod.F90
```diff
@@ -2638,45 +2649,56 @@ subroutine DetermineGridCellNeighbors(neighbors,seeds,numg)
...
+   if(debug .and. masterproc) call pstruct%ReportAccessCounts()
 
-  call FatesReceiveParams(fates_params)
-  call SpitFireReceiveParams(fates_params)
-  call PRTReceiveParams(fates_params)
-  call LeafBiophysReceiveParams(fates_params)
-  call FatesSynchronizedParamsInst%ReceiveParams(fates_params)
+ end subroutine FatesTransferParameters
 
-  call fates_params%Destroy()
```
ekluzek, line 2677: Is there an equivalent of the Destroy and deallocate for the new method?

parameter_files/TestParsing/TestJSON.F90
```diff
@@ -0,0 +1,52 @@
+program TestJSON
```
ekluzek, line 1: I'm wondering why this doesn't use a unit testing framework?

ekluzek: Oh, it's probably because it does I/O which we try to avoid in unittesting. Although since this is native Fortran it might be OK to use I/O for these kind of text files.

Hmmm. It still might be easier to have tests declared in a test framework.

parameter_files/TestParsing/WrapJSonTestMod.F90
```diff
@@ -0,0 +1,8 @@
+module shr_sys_mod
+
+contains
+
+  subroutine shr_sys_abort()
+    stop
```
ekluzek, line 6: With a stop, you won't be able to test error handling. This is where a unittest framework can really help.

parameter_files/patch_default_bciopt224.json
ekluzek: To answer one of my own questions. It's above in

[parameter_files/fates_params_default.json](https://github.com/NGEET/fates/pull/1493/files#diff-b45118b5b47b7b8c7144a135f97cb943c3760bb39522e7a04638c8c0bf454ceb)

which I skipped looking at because it's appropriately large, and it didn't render it in github.

That file looks how I would expect it to.

parameter_files/fates_params_default.json
```diff
@@ -0,0 +1,2190 @@
...
+      "long_name": "Name of the litter classes, for variables associated with dimension fates_litterclass",
+      "units": "unitless - string",
+      "data": ["twig", "small branch", "large branch", "trunk", "dead leaves", "live grass"]
+    },
+    "fates_alloc_organ_priority": {
+      "dtype": "integer",
+      "dims": ["fates_plant_organs", "fates_pft"],
+      "long_name": "Priority level for allocation, 1  replaces turnover from storage, 2  same priority as storage use/replacement, 3  ascending in order of least importance",
+      "units": "index",
+      "data": [[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],[2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2],[3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3],[4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4]]
```
ekluzek, line 108: @mvdebolskiy recommended using 2D format for 2D fields, so something like...

```suggestion
      "data": [
                      [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
                      [2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2],
                      [3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3],
                      [4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4]
                   ]
```

Which does make it nicely readable by humans

parameter_files/fates_params_default.json
```diff
@@ -0,0 +1,2190 @@
+{
+  "attributes": {"history": "05/11/25, First instantation, copied from: ../parameter_files/fates_params_default.cdl. "},
+  "dimensions": {
```
ekluzek, line 3: It does seem like this could use some more meta-data. A general description of the file contents. I thought this might've been on the FATES parameter file cdl, but it looks like it isn't. And the CTSM parameter file doesn't have much of anything either. Since, it's not a NetCDF file the CF convention doesn't directly apply. I am really glad you have history as that's tends to be the most useful thing from the CF convention.

A couple other possible header attributes to use from the CF Convention:

title
comment
institution
references

Some others that might be useful to think about:

user (user that created the file)
hostname (hostname of the machine the file was created on, this could be an issue to replace exactly because of possible FP rounding)
file_version? or minimum_fates_version? (something to identify what FATES version the file will work with)

The last one is useful to give a graceful error message that a given file is older and needs to be updated. And can help you figure that out from the get go. In general this will happen when it for example doesn't find a needed parameter, but it can help to have it explicit.

That also makes me think that the process to update a JSON parameter file from an older version to the latest is something to work out. I think it means using the old file as input to modify_fates_paramfile.py and then have it create a new one. Does that sound about right? Actually that could also be a good test to do -- run modify_fates_paramfile.py on the default JSON file and make sure the output is identical.

main/JSONParameterUtilsMod.F90
ekluzek: It's good to always have an intent in arguments to show you thought about it. It does look like this needs to be both IN and OUT, if I'm reading it right. Since, it's a read method, making it a OUT could be a good thing to try for.

main/JSONParameterUtilsMod.F90
ekluzek: Here's info from intel on scan. 

https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-0/scan-function.html

ekluzek: It does look like it's backwards from the above fortran example, but possibly it reverses it if that makes more sense?

But, maybe using keywords would be a good idea for clarity anyway, such as:
```suggestion
       if (scan(set='{}',string=filechar)>0 .and. inside_quotes) then
```

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+     character(len=max_sl), allocatable :: dim_names(:) ! These are the indices of the dimensions
+                                                        ! associated with this variable
+     integer               :: ndims                     ! Number of dimensions,
+                                                        ! same as size (dim_names)
+     integer                         :: access_count    ! This is used to count how many
+                                                        ! times this parameter has been read
+                                                        ! on initialization, expecte number
+                                                        ! of times is 1, helps for debugging
+
+     ! --- Data holding structures. Each parameter will only use ONE of these ----
```
ekluzek, line 145: FYI. We talked a little bit about different approaches to this. There are some worthwhile advantages to a different programming model.

Two that we talked about:

1. OO Design
2. *.in file format


**OO Design**

This would have a base abstract class, and then the different type options in the specific extensions of it. This does help with:
- Reducing memory needed
- Making you only see the data type for one thing at a time
- Fewer lines of code for the top level
- Easier to expand
- Easier to remove unneeded options

***.in file format:**

This is a format available from CIME (so all HLM's) that allows types and dimensionality to be abstracted better and to create all the different instantiations needed for them. For that matter it really works with making the OO design a little easier by making some of the code for you.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    ! 2) parameters
+    !
+    ! All parameters must be scalar or associated with a known dimension.
+    !
+    ! Each section is started with an open bracket '{' and closed with a closed bracket '}'
+    ! For each section here we end with a call
+    ! to rewind and go back to the top of the file
+    
+    character(len=*),intent(in) :: filename
+    integer,intent(in)          :: file_unit
```
ekluzek: Make file_unit local to this subroutine.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    ! scratch space for storing data as strings
+    ! before its copied into data structures
+    character(len=max_ll), dimension(max_scr) :: string_scr 
+    
+    ! Flush the scratch string
+    call ClearStringScratch(string_scr,max_scr)
+
+    ! Open the file, it is assumed that an open file unit has been identified
+    ! and passed in as an argument
+    open(unit=file_unit, file=filename, status='old', &
```
ekluzek: Use the modern Fortran unit handling syntax.

```suggestion
    open(unewnit=file_unit, file=filename, status='old', &
```

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+        action='READ', iostat=io_status)
+    
+    if (io_status /= 0) THEN
+       write(log_unit,*) 'ERROR: Could not open parameter file: ', trim(filename)
+       write(log_unit,*) 'IOSTAT value: ', io_status
+       call shr_sys_abort()  ! Terminate program gracefully if file cannot be opened
```
ekluzek, line 224: Why is shr_sys_abort used here in place of fates_endrun?

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+
+    ! Use IACHAR to convert the character to its integer
+    ! code. Capital letters have a different code than lower
+    ! case.
+    
+    do i = 1,len(input_string)
+
+       char_code = iachar(input_string(i:i))
+        
+       ! Check if the character is an uppercase letter (A-Z)
```
ekluzek, line 1100: This does rely on ASCII, which if it can get the ASCII representation will function. But, won't if not. I think iachar returns 0 when it's not in ASCII. So you should add an error when it's zero. So this implicitly adds a requirement that input JSON files are ASCII representable, which is probably fine. But, something we should increase visibility for.

I did find an example function that doesn't require ASCII at the end of this stackloverflow answer.

https://stackoverflow.com/questions/10759375/how-can-i-write-a-to-upper-or-to-lower-function-in-f90

But, it still might be OK to insist that characters in our JSON files are ASCII representable. Something to think about.

ekluzek (approved):
I do have a list of suggestions, and questions but I don't want to hold this up by needing to get my approval again. I also think this is large enough that it would be good to bring in, without requiring it get larger. So approving.

Thanks for all the work here @rgknox I do really like how the JSON files are so much easier to work with than CDL and map so well into python dictionaries. It also removes some of the HLM back and forth regarding NetCDF parameter files, so this moves a good chunk of code the HLM's to FATES, which is better code reusse.

JSON files as a recognized standard format also have more tools to work with them and do syntax checking and linting. So this is a nice modernization improvement. And should make it easier for users to work with model.

parameter_files/fates_params_default.json
```diff
@@ -0,0 +1,2190 @@
...
+    "fates_hlm_pftno": 14,
+    "fates_hydr_organs": 4,
+    "fates_landuseclass": 5,
+    "fates_leafage_class": 1,
+    "fates_litterclass": 6,
+    "fates_pft": 14,
+    "fates_plant_organs": 4
+  },
+  "parameters": {
+    "fates_history_ageclass_bin_edges": {
```
ekluzek, line 19: I was thinking about this and I think it would help to have one additional level in here. And that's a groupname level. So there would be something like this:

``` json
  "parameters": {
    "fates_history": {
        "comment": { "FATES History parameters"},
        "fates_history_ageclass_bin_edges": {
...
        }
        "fates_generic":{
              "comment": { "Generic FATES parameters that are always used"},
              "fates_pftname": {
...
               }
        }
        "fates_alloc":{
              // This list of parameters is used in the allocate part of FATES code
              "comment":{ "FATES parameters having to do with allocation"},
              "fates_alloc_organ_id": {
...
               }
        // FATES parameters used in FatesAllometryMod.F90
        "fates_allom":{
              "comment":{ "FATES parameters having to do with Allometry"},
              "fates_allom_agb1": {
...
               }
        }
        // FATES parameters ONLY used when fates_hydro is turned on
        "fates_hydro": {
            "comment":{ "FATES parameters for the fates_hydro parameterization"},
            "fates_hydro_organ_name": {
...
            }
        // FATES parameters used by the spitfire model in the fates/fire directory
        "fates_spitfire": {
            "comment":{ "parameters used by the spitfire model, and only used when it's active"},
            "fates_fire_alpha_SH": {
...
            }
        }
    }
```
}

ekluzek: As I think about it, the above structure helps to match the parameters to what is happening in the code. And would be useful for users to see to figure out how to know what parameters to modify for a given thing they want to do. This also makes parameterizations that are new to be added in as a group and ones that are deprecated to be easily removed. It also helps with figuring out what version of FATES a file will work with -- as you can see the group level and see why it only works with FATES version X > Y.

ekluzek: In the above I also demonstrate using C++ style comments which I think would be a useful add for FATES. C++ style comments aren't allowed in the standard, but are for the MS extension of it.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+                                                        ! of times is 1, helps for debugging
+
+     ! --- Data holding structures. Each parameter will only use ONE of these ----
+     real(r8)                        :: r_data_scalar
+     real(r8), allocatable           :: r_data_1d(:)
+     real(r8), allocatable           :: r_data_2d(:,:)
+     integer                         :: i_data_scalar
+     integer,  allocatable           :: i_data_1d(:)
+     integer,  allocatable           :: i_data_2d(:,:)
+     character(len=128)              :: c_data
```
ekluzek: In the last weeks CTSM SE meeting it was talked about that the 128 character length should be larger. And likely you should use for example the share code shr_kind_CL. This is available in share code and each HLM has it. And actually shr_kind_CL is already elsewhere in FATES.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    if(.not.found_scalar)then
+       pstruct%dimensions(n_dims+1)%name = 'scalar'
+       pstruct%dimensions(n_dims+1)%size = 1
+    end if
+    
+    
+    return
+  end subroutine GetDimensions
+
+  ! =====================================================================================
```
ekluzek, line 555: I recommend adding a subroutine to ignore and remove comments that can be used here and in other places, so that comments get removed from the final group_str.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+  integer, parameter :: max_ll = 2048          ! Maximum allowable line-length
+  integer, parameter :: max_sl = 512           ! Maximum allowable symbol-length
+  integer, parameter :: max_ul = 128           ! Maximum characters for units
+  integer, parameter :: max_scr = 1000         ! Maximum number of strings that
+                                               ! need to be remembered at once. This
+                                               ! needs to be larger than the
+                                               ! total number of vector/array members
+                                               ! in any single variable/parameter
+  
+  integer, parameter :: vardata_max_len = 4096 ! Maximum number of characters that would
```
ekluzek, line 80: use shr_kind_CXX

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+  integer, parameter :: i_1d_type     = 5
+  integer, parameter :: i_2d_type     = 6
+  integer, parameter :: c_solo_type   = 7
+  integer, parameter :: c_1d_type     = 8
+  integer, parameter :: c_2d_type     = 9
+
+  logical, parameter :: debug = .false.
+
+  integer, parameter :: max_ll = 2048          ! Maximum allowable line-length
+  integer, parameter :: max_sl = 512           ! Maximum allowable symbol-length
```
ekluzek, line 72: USe shr_kind_CX

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+  integer, parameter :: max_ul = 128           ! Maximum characters for units
```
ekluzek, line 73: Seems like this should be shr_kind_CL

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+          write(log_unit,*) 'fatal i/o error! code:', io_status
+          write(log_unit,*) 'msg: ',io_msg
+          call shr_sys_abort()
+       end if
+
+       i=i+1
+       
+       if(i<=max_sl)then
+          group_str(i:i) = filechar
+       else
```
ekluzek: What happens if a string is longer than max_sl? Here and in other places with similar things going on? Rather than terminating at this point, you should probably report that the string is too long and abort. That might be in PopString, but I wasn't sure.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    character(len=1)   :: filechar
+    logical            :: inside_quotes
+    integer            :: io_status
+    character(len=256) :: io_msg
+    
+    inside_quotes = .false.
+    
+    do_readfile: do 
+       
+       read(unit=file_unit,fmt='(A1)',ADVANCE='NO', iostat=io_status, iomsg=io_msg) filechar
```
ekluzek, line 312: It looks like this structure is repeated at least 4X, so a good point to refactor into a little helper function/subroutine.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    end if
+    
+    
+    return
+  end subroutine GetDimensions
+
+  ! =====================================================================================
+
+  subroutine ReadCharVar(file_unit,phase,var_num,string_scr,found_vartag,pstruct)
+
```
ekluzek, line 558: This subroutine is really long and complicated, so could use to be broken up into smaller pieces.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+       dimsizes(i) = pstruct%GetDimSizeFromName(trim(param%dim_names(i)))
+    end do
+    call ClearStringScratch(string_scr,n_vec_out)
+
+    
+    call GetMetaString(vardata_str,'"dtype"',beg_id,end_id)
+    call GetStringVec(vardata_str(beg_id:end_id),string_scr,n_vec_out)
+    data_str = trim(string_scr(1))
+    param%dtype = -999
+    if(trim(CleanSymbol(data_str))=='float') then
```
ekluzek, line 608: The list of if's for data type, would be better to be inside a a helper function to hide the complexity having to do with type.

main/JSONParameterUtilsMod.F90
```diff
@@ -0,0 +1,1304 @@
...
+    
+    call GetMetaString(vardata_str,'"units"',beg_id,end_id)
+    call GetStringVec(vardata_str(beg_id:end_id),string_scr,n_vec_out)
+    data_str = trim(string_scr(1))
+    param%units = trim(CleanSymbol(data_str))
+    call ClearStringScratch(string_scr,n_vec_out)
+    
+    call GetMetaString(vardata_str,'"data"',beg_id,end_id)
+    call GetStringVec(vardata_str(beg_id:end_id),string_scr,n_vec_out)
+    select case(param%dtype)
```
ekluzek, line 669: Another good place to move the case statement into a helper subroutine.

main/JSONParameterUtilsMod.F90
ekluzek: From looking at PopString it looks like the way it might fail is that it wouldn't find the ending close of the string or bracket. But, this would be hard for a user to decipher because they would have it right in the file. So I think more thinking about error handling for strings that exceed max_sl should be done.

tools/datatypes.txt
rgknox: This dataset is temporary and will be removed, I just used it as a notepad to write down the datatypes

main/FatesInterfaceMod.F90
rgknox: done

main/JSONParameterUtilsMod.F90
rgknox: This is looking to see if the second argument, which is a single character can be found in any of the first argument's characters.

main/JSONParameterUtilsMod.F90
rgknox: Yes, this is exactly what it is doing, shifting the string, adding the newly read character to the end, and dropping the first. I'm always happy to have more descriptive names

main/JSONParameterUtilsMod.F90
rgknox: This gives the option if people want to have scalar defined in the dimensions.  But I suppose that by giving it a dimension, people might assume its an array...

main/FatesInterfaceMod.F90
rgknox: Done

main/FatesInterfaceMod.F90
rgknox: done

main/FatesInterfaceMod.F90
rgknox: No, but we could add one.

main/JSONParameterUtilsMod.F90
rgknox: Easier for simplified unit testing.

parameter_files/fates_params_default.json
rgknox: I like the idea of a group.  @glemieux and I talked a little about having groups.  We agreed that it would be something to roll out in a future version.  But it would be nice to at least have the structure now.  We could have a group called "all" to hold all of the current parameters or something similar.  It wouldn't be a trivial change to the structure of the code and the derived type though...

rgknox: Also, I like the idea of sticking to  "comment" : "this is a comment".  This adheres more strictly to the json standard, works just fine, and would make the code a little simpler. How does that sound @ekluzek ?

main/JSONParameterUtilsMod.F90
rgknox: we've never used shr_kind_CL in fates.  I'll take a look at how it's used in CTSM

rgknox: so shr_kind_CL is just an integer, 256.  I like the idea of specifying this size locally, that way we can change based off of this data structure's specific needs if we like

main/JSONParameterUtilsMod.F90
rgknox: Again, I want this locally defined so that we can adjust these as need be.  Extra long string should not be a universal parameter imho anyway, it should depend on the use case.

main/JSONParameterUtilsMod.F90
rgknox: see above

parameter_files/fates_params_default.json
glemieux: > I like the idea of a group. @glemieux and I talked a little about having groups. We agreed that it would be something to roll out in a future version. But it would be nice to at least have the structure now. We could have a group called "all" to hold all of the current parameters or something similar. It wouldn't be a trivial change to the structure of the code and the derived type though...

I agree this would be really nice to have now, but I'm leaning towards updating this in a future V2.  I feel like we'd have to change the parser code a non-trivial amount and I feel like getting this in soon is preferrable.

ekluzek: > Also, I like the idea of sticking to "comment" : "this is a comment". This adheres more strictly to the json standard, works just fine, and would make the code a little simpler. How does that sound @ekluzek ?

I can go for using "comment" for comments, rather than C style commenting. I think I prefer having both personally. But, I'm fine with using comment as a proper JSON style token for things like this.

ekluzek: > > I like the idea of a group. @glemieux and I talked a little about having groups. We agreed that it would be something to roll out in a future version. But it would be nice to at least have the structure now. We could have a group called "all" to hold all of the current parameters or something similar. It wouldn't be a trivial change to the structure of the code and the derived type though...
> 
> I agree this would be really nice to have now, but I'm leaning towards updating this in a future V2. I feel like we'd have to change the parser code a non-trivial amount and I feel like getting this in soon is preferable.

Hmmm. I was hoping it wouldn't change the parser much to add that extra level. Definitely if it does, don't include it now. I might not be appreciating how hard that would be to add. I was thinking there would just need to be a subroutine or two to handle that layer. But, maybe that makes the finding the end bracket bit more difficult?

main/JSONParameterUtilsMod.F90
ekluzek: Yeah, there that much special about shr_kind_CL, it's just managed by CIME. And being managed by CIME brings both pro's and con's. I do think it's important to have a parameter that is used -- but I don't care that much if you use the CIME version or a local version. There is some value in using something general that's already provided for FATES and all HLM's.

It is used in FATES, but only two places:

https://github.com/search?q=repo%3ANGEET%2Ffates+shr_kind_CL&type=code

main/JSONParameterUtilsMod.F90
ekluzek: OK, sounds good.

main/JSONParameterUtilsMod.F90
ekluzek: Removing this request.

main/FatesInterfaceMod.F90
rgknox: destroy method has been added

parameter_files/TestParsing/TestJSON.F90
rgknox: @adrifoster 's unit testing framework has been updated to use the JSON format. In following PRs we will add tests that target JSON usage.  This test simply provides the parser with the filename and calls the parser, which is the first step in most of @adrifoster 's unit tests anyway.

parameter_files/TestParsing/WrapJSonTestMod.F90
rgknox: yup, now that this is hooked into @adrifoster 's unit testing we will migrate away from this simplified test and use actual sys aborts, not a dummy wrapper

main/JSONParameterUtilsMod.F90
rgknox: major refactor done to have subroutines used for repeated procedures.

main/JSONParameterUtilsMod.F90
rgknox: I started to add this capability, then switched over to refactor the code so that it used subroutines to get the next tag and get the next object.  It should not be difficult to add this feature in with the refactored code. I think I just need to add in a logical return on FindNextTag() to report if it found "comment". I'll make an issue

rgknox: see #1514

main/JSONParameterUtilsMod.F90
rgknox: Refactor reduced the scope of this routine considerably.

main/JSONParameterUtilsMod.F90
rgknox: I think it should be fine here. The lowercase function is only used to check for values of "null" or "nan", also by forcing these strings with uppercase to lowercase.  If a character is encountered outside the ascii set, it simply won't be changed, since it shouldn't return an iachar() between A and Z.  We also use iachar() to find and remove whitespace, I suppose we could find ways to expand to whitespace codes outside of the ascii set.

main/JSONParameterUtilsMod.F90
rgknox: This is deprecated now, we don't define the scalar in the dimension set anymore, its kind of meaningless

main/JSONParameterUtilsMod.F90
rgknox: popstring() was dropped in the refactor

parameter_files/TestParsing/TestJSON.F90
ekluzek: OK, if I understand the plan it's that this will be refactored to use the unit testing framework, in a follow on PR. So I'll close this conversation.

The modularization refactor you did goes really far into making it easier to unit test the JSON parser.

main/JSONParameterUtilsMod.F90
ekluzek: This conversation was just about talking about alternative paths. I don't think there's an action item for this PR here. I just wanted it there as something to think about.

So I'll close this for the purposes of this PR.

main/JSONParameterUtilsMod.F90
ekluzek: This will be something that comes out of adding unit-testing.

parameter_files/fates_params_default.json
ekluzek: We resolved the comment bit and that's captured in #1514. I'm going to add mention of the group idea, and make a separate issue for it.

ekluzek: OK this is all captured under #1514 now.

parameter_files/fates_params_default.json
ekluzek: Mentioned in #1514

main/JSONParameterUtilsMod.F90
ekluzek: Yes, thank you! That made a huge difference. I added a note to #1514 that more could be done. But, the refactor really helped.

main/JSONParameterUtilsMod.F90
ekluzek: I added a mention of this in #1514

parameter_files/fates_params_default.json
ekluzek: This is talked about in #1514

PR #1495, opened 2025-10-27
Update secondary young frac calculation

### LINKED ISSUE #1478: Unexpected large jumps in harvest
### Describe the issue

While analyzing the Trendy runs from CLM-FATES I found some suspicious large jumps in the harvest fields. These are seen across experiments, not just when transient land-use is turned on (S3). Below is a screen shot of some selected points that exhibit this in S0 (reference case with constant climate, CO2 and harvest rates). 

Image

I noticed that FATES_DISTURBANCE_RATE_LOGGING shows a large spike at the same times (up to 5-6 m2/m2/yr in the monthly data, which match a ~50% drop in LAI and cVeg), but FATES_MORTALITY_LOGGING_SZ does not snow similar spikes in any of the size classes (here for 3 sites in Norway that shows this behavior during the same decade in S3):

Image

I cannot see anything in the LU forcing file that should give this spike at these times. 

@ckoven, do you see similar behavior in your ELM-FATES runs? And do you have any ideas where this might come from?

### Relevant log output

```shell

```

### FATES tag

sci.1.85.1_api.40.0.0_noresm_v3

### Host land model tag

ctsm5.3.045_noresm_v10

### Machine

other HLM supported machine

### Other supported machine name

betzy

### Additional context

_No response_

JessicaNeedham:
### Description:
This change fixes #1478 .  Since get_harvest_rate_area is using patch age_since_anthro_disturbance to determine if patches are over the threshold to count as secondary old, the function get_secondary_young_frac also needs to use age_since_anthro_disturbance and not patch age, otherwise the secondary young frac can be 1, but some patches are over the threshold to count as secondary old, triggering 100% harvest. 

### Collaborators:
@kjetilaas @rosiealice @ckoven @sshu88 

This will be answer changing if land use is on. It will reduce large spikes in harvest. 

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:
Tested on the NorESM-FATES side and prevents the case where secondary young frac = 1, but the patch is over the age threshold.

### DIFF
--- a/main/EDTypesMod.F90
+++ b/main/EDTypesMod.F90
@@ -848,7 +848,7 @@ contains
      currentPatch => this%oldest_patch
      do while (associated(currentPatch))
         if (currentPatch%land_use_label .eq. secondaryland) then
-           if ( currentPatch%age .ge. secondary_age_threshold ) then
+           if ( currentPatch%age_since_anthro_disturbance .ge. secondary_age_threshold ) then
               secondary_old_area = secondary_old_area + currentPatch%area
            else
               secondary_young_area = secondary_young_area + currentPatch%area


glemieux:
Regression testing underway on `derecho`.

PR #1497, opened 2025-10-27
Rothermel model bug fix

### LINKED ISSUE #1494: Correct Rothermel ROS model for SPITFIRE
### Task description

As already pointed out by [Oberhagemann et al. 2025](https://gmd.copernicus.org/articles/18/2021/2025/) that the implementation of Rothermel ROS model in SPITFIRE is wrong due to:
1) the weighthing factor for calculating average fuel load, SA:V, and bulk density is based on fuel load contribution by each fuel size class (namely, the fractional fuel load); however, Rothermel model uses the fractional total surface area contributed by fuel classes. The difference between fuel-load vs fuel SA:V weighted fuel characteristics is that: the former will aways be weighted toward large/heavy fuel size (e.g. basically all woody fuels) and the latter will always be weighted toward finer fuels with large SA:V (e.g., leaf litter, live grass and maybe also 1h fuel). This creates a huge difference between the averaged SA:V of a heterogenous fuel bed, especially when the fuel bed has considerable amount of woody fuels. See figures below for the comparison. 

**Note: fuel model 161 and 149 have woody fuels with 161 having a much higher fractional fuel load in 1h-100h fuel class.**
Fig1. Average SA:V calculated using Rothermel model

Image

Fig2. Average SA:V calculated using SPITFIRE model 

Image

2) Rothermel model treats live and deal fuel as distinct fuel category by calculating reaction intensity, mean fuel moisture content, fuel moisture of extinction separately for live vs dead fuels. To calculate ROS, reaction intensity of dead and live fuels are summed later on. However, SPITFIRE calculate ROS by ignoring the difference between dead and live fuels and mix the two into one average fuel using the fuel load weighting factor, and from there evrything is calculated based on this average fuel trait (e.g., averaged fuel moisture, averaged fuel moisture of extinction, and reaction intensity based on the averaged FMC and MEF). 

We originally thought that this might not be a big problem we have to fix now (e.g., overestimated or underestimated ROS under certain wind speed according to Oberhagemann et al. 2025). But during the process of validating crown fire model, I noticed that calculated crowning index using EQ. 20 in Scott and Reinhard 2001 is always smaller than the actual crowning index (wind speed at which an active crown fire will happen). This is due to the wrong calculation of average fuel bed SA:V in SPITFIRE model while EQ. 20 is based on the correct Rothermel model. This mismatch create a weird pattern where we see that a passive crown fire can occur after wind speed passes the torching index, however, we did not see a gradual transition to active crown fire, rather a transition from passive crown fire to surface fire and then to active crown fire. This is largely due to the fact that when wind speed in in-between the calculated, false crowning index and the true crowning index, neither the passive crown nor the active crown fire condition is met, thus the model determines there is no crown fire and switches back to surface fire. 

Fig. 3: crown fraction burned by crown fire. 0<CFB<1 indicates a passive crown fire; CFB=1 indicates a active crown fire; 0 means surface fire. 

Image

I'm currently working on a branch that try to correct Rothermel model in FATES-SPITFIRE. There is another issue I have discovered: all existing fuel models define surface fuel bed as anything, live or dead, inbetween the ground surface and ~6ft above the ground, so tree saplings are also considered surface fuels. But FATES only consider live grasses as live fuels, so in tree-dominant systems we have ZERO live fuels. This can create a problem for some of the Rothermel model calculations when live fuel acts as the denominator in the equation. While it's not hard to redefine our surface fuel category, I also realized that we always need to calculte fuel fraction burnt in order to determine fire intensity, which is then used to determine fire impacts (e.g crown fraction burnt) for calculating post-fire cohort mortality. For cohorts that also act as surface fuels this becomes circular.....

Any comments are welcomed. And I will submit a PR later for fixing the Rothermel model

XiulinGao:
### Description:
This PR is to correct Rothermel surface fire model in FATES-SPITFIRE by fixing
1) the calculation of verage fuel bed SA:V, bulk density, and fuel load of a heterogenous fuel bed (e.g., a fuel bed that is made of different sizes and category of fuel) by switching to a SA:V based weighting factor

2) calculate reaction intensity, fuel load, fuel moisture, and fuel moisture of extinction for dead vs live fuel separately instead of calculating a single value based on the averaged fuel traits.

3) I also correct some minor unit conversion issue. For instance, the orginal EQ. for calculating max. reaction velocity in Rothermel is SAV^1.5*(495 + 0.0594*SAV^1.5)^-1, in which unit of SAV is ft2/ft3; as we are using cm2/cm3 for SAV, we need to convert cm2/cm3 to ft2/ft3 by multiplying 30.48 in order to use this EQ, which leads to 1.0_r8/(0.0591_r8 + 2.942_r8*(SAV**(-1.5_r8))) (need to do a simple transformation by dividing the orginal equation by SAV^1.5); however in SPITFIRE this is 1.0_r8/(0.0591_r8 + 2.926_r8*(SAV**(-1.5_r8))), a relatively large difference in the third constant. 

Note: if you see a large file diff that's because I used code formatting tool, which added a lot white space and I don't know how to revert that without quitting all the essential code change. So please view file change by selecting 'hide whitespace' in setting next to the 'conversation' dropdown. 

This is to fix #1494 

### Collaborators:
@rosiealice 

Yes, we expect to see changes in surface fuel characteristics and fire behavior when fire is turned on, which will also result in different vegetation response.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/fire/FatesFuelClassesMod.F90
+++ b/fire/FatesFuelClassesMod.F90
@@ -1,63 +1,63 @@
 module FatesFuelClassesMod
 
-  use FatesLitterMod, only : ncwd
-  
-  implicit none
-  private
+   use FatesLitterMod, only : ncwd
 
-  integer, parameter, public :: num_fuel_classes = 6 ! number of total fuel classes  
-    
-  type :: fuel_classes_type
-    ! There are six fuel classes:
-    ! 1) twigs, 2) small branches, 3) large branches 4) trunks
-    ! 5) dead leaves, 6) live grass
-    integer, private :: twigs_i = 1          ! array index for twigs pool
-    integer, private :: small_branches_i = 2 ! array index for small branches pool
-    integer, private :: large_branches_i = 3 ! array index for large branches pool
-    integer, private :: dead_leaves_i = 5    ! array index for dead leaves pool
-    integer, private :: live_grass_i = 6     ! array index for live grass pool
-    integer, private :: trunks_i = 4         ! array index for trunks pool
+   implicit none
+   private
 
-    contains 
+   integer, parameter, public :: num_fuel_classes = 6 ! number of total fuel classes
 
-    procedure :: twigs, small_branches, large_branches, trunks
-    procedure :: dead_leaves, live_grass
+   type :: fuel_classes_type
+      ! There are six fuel classes:
+      ! 1) twigs, 2) small branches, 3) large branches 4) trunks
+      ! 5) dead leaves, 6) live grass
+      integer, private :: twigs_i = 1          ! array index for twigs pool
+      integer, private :: small_branches_i = 2 ! array index for small branches pool
+      integer, private :: large_branches_i = 3 ! array index for large branches pool
+      integer, private :: dead_leaves_i = 5    ! array index for dead leaves pool
+      integer, private :: live_grass_i = 6     ! array index for live grass pool
+      integer, private :: trunks_i = 4         ! array index for trunks pool
 
-  end type fuel_classes_type
+   contains
 
-  ! actual type we can pass around 
-  type(fuel_classes_type), public :: fuel_classes
+      procedure :: twigs, small_branches, large_branches, trunks
+      procedure :: dead_leaves, live_grass
 
-  contains
+   end type fuel_classes_type
 
-  integer function twigs(this)
-    class(fuel_classes_type), intent(in) :: this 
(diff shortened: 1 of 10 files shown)


samsrabin:
Unsubscribing for now, but let me know when this is ready for review.

Issue #1508, opened 2025-11-17
Negative promotion/demotion area due to double float precision


jiamenglai:
### Describe the issue

Hello, I met an error of 'negative or more area than the cohort has is being promoted/demoted' in EDCanopyStructureMod, and I think it's due to double float precision. I am running ELM-FATES with ed_st3 on and with my inventory file. The trigger of this error is in the code lines below (Line 662 - 664):

remainder_area = cohort%c_area - layer_co(ic)%pd_area
copyc%n = cohort%n*remainder_area/cohort%c_area
cohort%n = cohort%n - copyc%n

When I output those values, I found that pd_area is about 4E-16 (positive), but the copyc%n is slightly larger than the original cohort%n, which is wrong mathematically. This gives a negative cohort%n after those code, and led to negative demotion area in the next demotion phase. 

Any suggestions to fix this bug? Thanks!

### Relevant log output

```shell

```

### FATES tag

sci.1.88.0_api.42.0.0

### Host land model tag

v3.1.0-alpha-4304-ge909ef9f6f

### Machine

perlmutter

mpaiao:
Thanks for making this issue @jiamenglai. I don't know if it is helpful, but maybe you could paste the extra output you added to figure out the problem in this issue too.

jiamenglai:
Thanks @mpaiao for the suggestion. I revised the code to output some variables:

remainder_area = cohort%c_area - layer_co(ic)%pd_area
! The copied cohort is the part that remains in-layer
copyc%n = cohort%n*remainder_area/cohort%c_area
            if (copyc%n > cohort%n) then
               write(fates_log(),*)'copyc%n > cohort%n'
               write(fates_log(),*)'cohort%n',cohort%n
               write(fates_log(),*)'copyc%n',copyc%n
               write(fates_log(),*)'pd_area',layer_co(ic)%pd_area
               write(fates_log(),*)'remainder_area',remainder_area
               write(fates_log(),*)'remainder_area/cohort%c_area',remainder_area/cohort%c_area
            endif
! The original cohort changes layers
cohort%n = cohort%n - copyc%n

The output is as below:
<img width="679" height="224" alt="Image" src="https://github.com/user-attachments/assets/44adf04d-f5c7-4544-a704-138c45bc5da0" />

jiamenglai:
> [@jiamenglai](https://github.com/jiamenglai) I submitted a potential fix here: [#1509](https://github.com/NGEET/fates/pull/1509)
> 
[...]

PR #1509, opened 2025-11-19
Protections on promotion/demotion for near zero sized cohorts

rgknox:
### Description:

During promotion and demotion, it is possible to have very small or even negative very small cohorts after the process of splitting.  These changes seek to 1) prevent negatively sized cohorts by forcing fractional promotion/demotion rates between 0 and 1 at the time of allocation and 2) force a termination step (level 1, which is cohort number only) following each promotion and demotion pass through.

In the past we had trouble with a termination step during the promotion/demotion sequence.  This is because termination would remove area from the layer while it is trying to form layers with specific target areas, thereby constantly messing with the convergence algorithm's attempts.  However, the code has since been refactored and so sparse cohort numbers should be very small, not marginally small. And, we are only imposing level 1 termination (number density only) which should have minimal effects on canopy area.

Also, we somehow dropped the updating of the previous day's canopy layer, this has been re-instated.

Fixes: #1508 

### Collaborators:
@jiamenglai  @walkeranthonyp 

This should have round-off level impacts on all cases where we perform canopy promotion/demotion/dynamics (ie everything but SP).

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -78,7 +78,10 @@ module EDCanopyStructureMod
   ! Precision targets for demotion and promotion
   ! "co_area_target_precision" is the required precision at the cohort level,
   !    essentially it is the minimum amount of change required to not ignore
-  !    a partial promotion or demotion
+  !    a partial promotion or demotion. This number was chosen because
+  !    math precision is around 15 digits in fortran r8s, with numbers that
+  !    can get to magnitude e4, this gives us about two orders of magitude in math
+  !    precision (ie e15-(e4+e9)=e2) in significant digits to match this absolute precision
   
   real(r8), parameter :: co_area_target_precision = 1.0E-9_r8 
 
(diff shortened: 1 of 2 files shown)


biogeochem/EDCanopyStructureMod.F90
mvdebolskiy: with `arealayer` being an array, if z>nclmax+5, you will get out of bounds. Since if area is `area_not_balanced` and you've demoted a cohort that was in layer `nclmax+5`. I am not sure how this is happening but with Rosie's parameter file for NorESM it does happen with ESCOM/CTSM:master:HEAD checkout. I've tried to change tolerances for fusing (since Rosie has them x3 of what the default on NGEET is) but that did not help. 

Further investigations showed that there is a single cohort in the lowest layer that keeps getting demoted, not being able to fuse with anyone or getting promoted back into it's original layer.

glemieux (approved):

PR #1528, opened 2026-01-30
direct usage of parameter datastructure - allometry proof of concept

rgknox:
### Description:

With the introduction of the parameter data structure, we can access parameters directly from that structure and bypass the middle-man, ie the allocated data primitive arrays that we fill from the parameter data structure.

Following initial review and discussion, the relevant data primitives will be removed.

This should not change any answers, B4B

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

TBD

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/FatesAllometryMod.F90
+++ b/biogeochem/FatesAllometryMod.F90
@@ -25,7 +25,6 @@
 ! bbgw:   biomass below ground woody tissues [kgC] 
 !         similar to AGBW, this essentially encompasses
 !         all non-fineroot belowground tissues
-! bfrmax: biomass in fine roots when "on allometry" [kgC]
 ! bsap: biomass in sapwood (above and below) [kgC]
 ! bdead: biomass (above and below) in the structural pool [kgC]
 !
@@ -83,7 +82,8 @@ module FatesAllometryMod
 
   ! If this is a unit-test, these globals will be provided by a wrapper
 
-  use PRTParametersMod, only  : prt_params
+  ! This contains the pstruct data structure, and anything pid_
+  use FatesParametersInterface
   use FatesConstantsMod, only : r8 => fates_r8
   use FatesConstantsMod, only : i4 => fates_int
   use FatesConstantsMod, only : g_per_kg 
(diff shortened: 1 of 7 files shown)


rgknox:
I have a prototype of an alternative approach, see this commit:  
https://github.com/rgknox/fates/commit/54da9688eff046f49a7a574f282c43838e4c8cba

In this approach, the idea more similar to our previous approach. Every parameter is associated with a uniquely named primitive data structure, with a few key differences.

1) These are pointers, but not allocated pointers, because each points to the datastructure generated in the JSON parser which stays allocated. (I think it is better to keep the original structure intact, rather than transfer and deallocate, because it contains all the metadata, dimension names, etc, which is useful to keep)

2) These pointers have a protected status (this will prevent the pointer's association from changing outside the module)

3) All fates parameters are defined in one data structure in main/FatesParametersInterfaceMod.F90 (this would be completed in following steps, ie consolidate all parameters in fire,edpftvar,edparams,rad, parteh,etc)

4) the new data structure can either be a stand-alone, or an extension of the data structure defined in main/JSON (two options provided)

rgknox:
Thanks for your thoughts @adrifoster , yeah, I'm curious what others think on this too.

I also like the idea of moving back to one data structure to hold all parameter constants. So if we do go with the plan of transferring to the holder data structure, and removing the json data structure, lets stick with one structure, instanced in FatesParametersInterfaceMod, which has almost no dependencies.

Something we could do to mitigate some of the issues described thus far:

If we went with maintaining just the "pstruct" data structure, the index that identifies a parameter could have the same exact syntax as the parameter name in the file:  for instance the parameter 'fates_leaf_vcmax25top', could use integer index fates_leaf_vcmax25top.  This would actually reduce the number of names used, as right now we call this %vcmax25top.  This would reduce ambiguity, we only need remember the name as it appears in the file, should be easily searchable in the code.

That being said, the part that I also don't like about exclusively using the pstruct% json data structure to access the parameters in the code, is that we will dealing with these different array types in our code: %r_data_1d, sometimes %i_data_1d, etc ... It doesn't come across as very clean, and people might not understand what is going on.   Maybe I'm coming around on this idea of using a holder structure and deallocating the pstruct json structure.

Issue #1492, opened 2025-10-17
Hydraulic failure mortality is 0.0 for the whole permafrost region


mvdebolskiy:
In here:
https://github.com/NGEET/fates/blob/3b58472a513e39fcb8d0dafc7bc0ae31b91f6e95/biogeochem/EDMortalityFunctionsMod.F90#L186

`minval()` is used over the array `bc_in%t_soisno_sl` which in CLM (and I assume ELM) has size (1:nlevsoi). This means, that the condition will not trigger until the __whole__ soil column is >-2.0 degrees C. This excludes most of the permafrost region, while (I assume) the plants only care about the soil layers where they actually have roots (so they _should_ be able to experience hydraulic failure mortality during the summer when active layer is deep enough).

PR #1533, opened 2026-02-10
Update hydraulic failure mortality for places with frozen soils

JessicaNeedham:
This PR address issue https://github.com/NGEET/fates/issues/1492.
Hydraulic failure mortality (hmort) occurs when btran is less than a soil moisture threshold. If soils are frozen then btran is zero. To prevent hmort of plants at high latitude during winter there is an if statement that prevents hmort being calculated if any soil layer is frozen. However, in regions with permafrost this essentially turns off hmort entirely, even in summer months when plants may actually be water stressed.

This PR adds additional logic so that we still calculate hmort if soil layers with 90% of the root biomass are not frozen.

### Collaborators:
@mvdebolskiy @rosiealice @kjetil

This will change hmort rates of plants at high latitudes.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:
Not tested

### DIFF
--- a/biogeochem/EDMortalityFunctionsMod.F90
+++ b/biogeochem/EDMortalityFunctionsMod.F90
@@ -71,7 +71,7 @@ contains
     use FatesInterfaceTypesMod, only : hlm_mort_cstarvation_model
     
     type (fates_cohort_type), intent(in) :: cohort_in
-    type (ed_site_type),intent(in)       :: site_in
+    type (ed_site_type),intent(inout)    :: site_in   ! We modify scratch space, hence in/out
     type (bc_in_type), intent(in)        :: bc_in
     real(r8), intent(in)                 :: btran_ft(maxpft)
     real(r8), intent(in)                 :: mean_temp
@@ -191,7 +191,7 @@ contains
           
           if ( (.not. is_decid_dormant) .and. &
                ( btran_ft(cohort_in%pft) <= hf_sm_threshold ) .and. &
-               ( ( minval(bc_in%t_soisno_sl(1:soil_root_i)) - tfrz ) > soil_tfrz_thresh ) ) then
+               ( ( minval(bc_in%t_soisno_sl(1:max_soil_ind)) - tfrz ) > soil_tfrz_thresh ) ) then
              hmort = EDPftvarcon_inst%mort_scalar_hydrfailure(cohort_in%pft)*((hf_sm_threshold- btran_ft(cohort_in%pft))/hf_sm_threshold)
           else
              hmort = 0.0_r8


JessicaNeedham:
For reference, here is global btran by season from a recent historic run. 

DJF: 
image

MAM: 
image

JJA: 
image


SON: 
image

JessicaNeedham:
With the condition that we calculate hydraulic failure when 50% of the soil layers in the active layer are thawed (~0.6 m with CLM soil layers) we see increased mortality - especially in NE China, SE Russia. Cold-deciduous broad leaf and needleleaf tree PFTs. 

image

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -181,9 +181,12 @@ subroutine mortality_rates( cohort_in,bc_in, btran_ft, mean_temp,             &
        else
           ! When FATES-Hydro is off, hydraulic failure mortality occurs only when btran
           ! falls below a threshold and plants have leaves.
+          n_layer_thawed = count((bc_in%t_soisno_sl - tfrz ) > soil_tfrz_thresh)
```
rgknox: @JessicaNeedham , could we also perform this logic based off of the fraction of non-frozen depth, or the fraction of non-frozen fine-root biomass?  That might buffer this calculation from artificial model constructs like how we decide on layering.

biogeochem/EDMortalityFunctionsMod.F90
```diff
@@ -181,9 +181,12 @@ subroutine mortality_rates( cohort_in,bc_in, btran_ft, mean_temp,             &
...
+          nlev_eff_soil   = max(bc_in%max_rooting_depth_index_col, 1)
+          
           if ( (.not. is_decid_dormant) .and. &
                ( btran_ft(cohort_in%pft) <= hf_sm_threshold ) .and. &
-               ( ( minval(bc_in%t_soisno_sl) - tfrz ) > soil_tfrz_thresh ) ) then
+               ( n_layer_thawed .ge. (0.5_r8 * nlev_eff_soil)  )) then
```
rgknox: Recommend setting 0.5 to a named constant like "hmort_thaw_frac_threshold" or whatever you deem appropriate

biogeochem/EDMortalityFunctionsMod.F90
JessicaNeedham: @rgknox yes it also occurred to me we could get quite different behaviour here if the HLM layering scheme changed. I might have missed something, but it looked to me like FATES doesn't actually see the thickness of each HLM soil layer, which makes it hard to do this by depth, rather than number of soil layers. Making it by root biomass works, and that was what I originally wrote. It is a bit clunkier but maybe safer (see this [first commit](https://github.com/NGEET/fates/pull/1533/changes/1aa88a6937aaa811ed79139dc7889285fcd3e5c0)).

PR #1540, opened 2026-03-05
Performance Optimizations

rgknox:
### Description:

This branch contains a litany of performance enhancements to FATES. I may try to break these changes into chunks, but it might not be worth it.  With a full FATES run at a single site with 9 patches and ~250 cohorts, speed-ups are about the following:

- 1) Everything in this PR: 2/3 to 1/2 total run-time as base **(1.5x+)**
- 2) Everything in 1) AND dual-loop* energy balance in the host: less than 1/2 of total run-time **(ie 2x+)**
- 3) Everything in 2) AND using 4 threads per site (host-side parallelism): **3x+**

*dual-loop energy balancing is a host-side change that will work with these changes. It reduces the total calls to photosynthesis to achieve temperature convergence in CanopyFluxesMod.

This PR is should be coupled with, or precede: https://github.com/E3SM-Project/E3SM/pull/8143

These changes can be described in several categories:

- Breaking up photosynthesis and canopy radiation drivers to have patch-specific driver subroutines, which accommodates patch-level parallelism and a refactored "double loop" land-energy balance calculation
- Utilizing automatic arrays (stack memory) versus dynamically allocated scratch arrays (heap) where appropriate
- Defining patch-level data-structures to help reduce memory usage and perform more efficient calculations: the list of unique pfts on each patch, number of unique pfts, and maximum number of veg-layers on each patch
- Using arrays of cohort data (as attached to the patch data structure) during high frequency operations, which are filled at the end of the dynamics timestep during summarization
- Created a pointer array that helps us quickly identify the fates linked-list patch associated with the host's patch index, which is synonymous with patch%patchno.  Ie, so we can do this:   patch => site%pa_vec(ifp)
- Removing unnecessary routines
- Moving routines to places in the call sequence where necessary but less frequently (e.g. organ respiration rates need not be inside the land energy balance iterator since they don't affect conductance).  Also, temperature affects on biophysical rates (i.e. vcmax, jmax and LMR) need only be updated once per patch-PFT during photosynthesis, and need not be updated for every unique leaf layer. 
- Converting routines to be classified as "elemental". Elemental's primary role is to enable calling routines to operate on a subroutine as either scalar or vector arguments. Its secondary role is that it tells compilers that they can safely in-line code from that routine into the calling routine.  This reduces computational overhead, sometimes greatly when it can leverage vectorization and SIMD stuff, but also to some degree when it can't.
- Pre-calculating as much math as possible in high-frequency routines.  For instance when applying temperature corrections to biophysical rates, the math is demanding (exponentials, power functions, divides, etc). So the functions were split into parts to perform as much math as possible before any math is applied that is dependent on vegetation temperature (which changes at the model timestep).
- Making more "use specific" variants of highest used functions, which allows us to make them faster.  For instance, many calls to quadratic smoothing have constants as the "a" term. In these scenarios, we don't need to perform "if" calls to make sure that a is positive and non-zero. Since quadratics are literally the most called function we have, its usefull to get rid of these logicals.
- Method of retrieving the values (ie the masses) of  C, N and P in plant organs that is faster and minimizes "cache misses".  

### Note:

This work made use of conversations with AI bots, such as gemini and claude (via Kiro). No suggestions were applied directly by the AIs.  Blocks of code and vignettes were copied over, but in small sections and were evaluated each in entirety.

### Collaborators:

@cdkoven @rosiealice @bishtgautam @peterdschwartz @mpaiao @glemieux @samsrabin 

This work used "do_b4b" logical flags.  When these flags are set to true, changes are expected at the roundoff level for all FATES configurations. When these flags are set to false, there are non round-off level changes that are still appropriate.  For instance, we don't need to update the growth and home temperatures for leaf temperature acclimation every 30 minutes, its excessive. This can be done once per day, since these are at least 30 day averages.  But changing this value will subtly change results. Thus the do_b4b flag.

*Note: Aside from mentioned above. I found FATES can potentially generate chaotic behavior.  For instance, when converting Q10 equations to use the exponential function instead of the power function, difference in the math outcome should be incredibly small (e-15). However these differences generated non-trivial differences in results.  I'm intending to investigate this more and compare differences from the math alternatives against perturbations of initial conditions.

### Checklist

*Contributor*

WIP, NOT YET

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*


peterdschwartz:
Have you experimented much with `do concurrent` constructs?  If you have a loop that doesn't require changes to convert, it may be worth it.  Compilers can take do concurrent and automatically parallelize and offload with openMP or openacc with certain compiler flags.  I realize that may not be important right now, but it would also serve as a compiler enforced contract for which loops are expected to be truly parallel for when that becomes import in the future.

peterdschwartz:
Also, not really an issue but might warrant a second thought:  `CoArrays` is the name of Fortran's native parallel programming data types.  So it could be potentially confusing down the road to have a variable with that name that isn't related actual coarrays.  

Again this isn't necessarily an issue - i just think it's worth pointing out

biogeophys/FatesPlantRespPhotosynthMod.F90
```diff
@@ -273,884 +273,769 @@ subroutine FatesPlantRespPhotosynthDrive (nsites, sites,bc_in,bc_out,dtime)
...
+                        laisun = (fsun*cohort_layer_elai(iv))
+                        laisha = ((1._r8 - fsun)*cohort_layer_elai(iv))
+                        if(fsun>nearzero) then
+                           par_per_sunla = (rd_abs_leaf*fsun + rb_abs_leaf)! / laisun
                         else
-                           crown_reduction = 0.0_r8
+                           par_per_sunla = 0._r8
                         end if
+                        par_per_shala = rd_abs_leaf*(1._r8-fsun) !/ laisha
+                        canopy_area = 1._r8 !patch%canopy_area_profile(cl,ft,iv)
```
rgknox, line 615: discussion with @adrifoster we determined that this comment behind the 1 is confusing and made us question if something was left undone. This text should be removed.

PR #1543, opened 2026-03-16
fungal wood respiration

rgknox:
### Description:

This set of changes introduces a new heterotrophic respiration loss applied to coarse woody debris.  This loss pathway is designed to mimic fungal wood decomposition, prior to entering the decomposition cascades in the host model.  This is applied to the fragmented loss fluxes.  A portion of the fragmented carbon loss from wood is sent to the hosts for further decomposition, and a portion is lost to the atmosphere now as this fungal respiration term.  

This change was motivate through discussions and brainstorming by the NGEET nutrient: including Anthony Walker, Charlie Koven, Xiaojuan Yang, Matthew Craig, Daniela Yaffar, Mingjie Shi, myself and others.

New FATES parameter: [fates_cwd_hrfrag_frac](https://github.com/rgknox/fates/blob/fungal-wood-respiration/parameter_files/fates_params_default.json#L1686-L1692)

This sets the fraction of carbon that is lost to fragmentation that ends up as fungal respiration flux.  Note that the higher the value of this term, the lower the C/N and C/P ratio (i.e. higher nutrient concentration) of the fragmentation flux into the host.

Control of total fragmentation rate remains unchanged, and is still based on a maximum rate set by the parameters:  [fates_frag_maxdecomp](https://github.com/rgknox/fates/blob/fungal-wood-respiration/parameter_files/fates_params_default.json#L1636C1-L1642C7)

This parameter is most useful in nutrient limited simulations. Moreover, we have found that this parameter is important to keeping C/N ratios of the total litter pool at lower and reasonable levels. 

### Collaborators:

Conceptual design: 
@walkeranthonyp, @ckoven, Xiaojuan Yang, @rgknox 

These changes should not impact C-only runs, aside from specific output variables related to litter composition. C-only dynamics should not be affected by this change.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -3295,12 +3295,12 @@ contains
     
     do c = 1,ncwd
 
-       decomp_loss = SF_val_max_decomp(c) * years_per_day * fragmentation_scaler(soil_layer_index)
+       decomp_rate = SF_val_max_decomp(c) * years_per_day * fragmentation_scaler(soil_layer_index)
        litt%ag_cwd_frag(c)   = litt%ag_cwd(c) * decomp_rate * (1._r8-cwd_hr_frag_frac_eff)
        litt%ag_cwd_funghr(c) = litt%ag_cwd(c) * decomp_rate * cwd_hr_frag_frac_eff
        
        do ilyr = 1,nlev_eff_decomp
-          decomp_loss = SF_val_max_decomp(c) * years_per_day * fragmentation_scaler(ilyr)
+          decomp_rate = SF_val_max_decomp(c) * years_per_day * fragmentation_scaler(ilyr)
           litt%bg_cwd_frag(c,ilyr)   = litt%bg_cwd(c,ilyr) * decomp_rate * (1._r8-cwd_hr_frag_frac_eff)
           litt%bg_cwd_funghr(c,ilyr) = litt%bg_cwd(c,ilyr) * decomp_rate * cwd_hr_frag_frac_eff
        enddo
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -20,6 +20,7 @@ module EDPhysiologyMod
   use FatesInterfaceTypesMod, only    : hlm_use_nocomp
   use EDParamsMod           , only    : crop_lu_pft_vector
   use EDParamsMod           , only    : GetNVegLayers
+  use EDParamsMod           , only    : cwd_hr_frag_frac
   use FatesInterfaceTypesMod, only    : hlm_use_tree_damage
   use FatesInterfaceTypesMod, only : hlm_use_ed_prescribed_phys
   use FatesConstantsMod, only    : r8 => fates_r8
@@ -3273,6 +3274,9 @@ contains
     integer :: ilyr                    ! Soil layer index
     integer :: dcmpy                   ! Decomposibility pool indexer
     integer :: soil_layer_index = 1    ! Soil layer index associated with above ground litter
+    real(r8) :: decomp_rate            ! per-mass per-day loss rate due to fragmentation AND fungal HR
+    real(r8) :: cwd_hr_frag_frac_eff   ! Effective (element specific) proportion of
+                                       ! losses due to HR
     !----------------------------------------------------------------------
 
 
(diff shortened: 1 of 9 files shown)


ckoven:
This looks awesome @rgknox, thanks!  Everything here looks great to me. The only comment I have is that we should probably pass this information on the fungal respiration flux through a bc_out field to the host model, so that it can be added into the `HR`, `NEE`, and `NBP` history variables as well. Not sure if that should come with this PR or a separate one though.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -3281,9 +3285,13 @@ subroutine CWDOut( litt, fragmentation_scaler, nlev_eff_decomp )
        litt%ag_cwd_frag(c)   = litt%ag_cwd(c) * SF_val_max_decomp(c) * &
              years_per_day * fragmentation_scaler(soil_layer_index)
 
+       litt%ag_cwd_funghr(c) = litt%ag_cwd(c) * fung_hetresp_baserate * years_per_day
```
walkeranthonyp, line 3288: This looks as if fungal HR is applied to the CWD pool rather than the CWD fragmentation flux. Discussed more in overall comments.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -3281,9 +3285,13 @@ subroutine CWDOut( litt, fragmentation_scaler, nlev_eff_decomp )
...
+       
        do ilyr = 1,nlev_eff_decomp
            litt%bg_cwd_frag(c,ilyr) = litt%bg_cwd(c,ilyr) * SF_val_max_decomp(c) * &
                 years_per_day * fragmentation_scaler(ilyr)
+
+           litt%bg_cwd_funghr(c,ilyr) = litt%bg_cwd(c,ilyr) * fung_hetresp_baserate * years_per_day
```
walkeranthonyp, line 3294: Same here

parameter_files/fates_params_default.json
```diff
@@ -1640,6 +1640,12 @@
       "units": "yr-1",
       "data": [0.52, 0.383, 0.383, 0.19, 1.0, 999.0]
     },
+      "fates_funghr_frac": {
+	  "dtype": "float",
+	  "long_name": "fraction of cwd losses due to fungal respiratoin and not fragmentation",
```
walkeranthonyp, line 1645: Minor spelling mistake in respiration

parameter_files/fates_params_default.json
```diff
@@ -1640,6 +1640,12 @@
...
+	  "units": "-",
+	  "data": [0.5]
```
walkeranthonyp, line 1647: Could make this ncwd long to allow the option to vary this by CWD pool

walkeranthonyp (approved):
This looks great Ryan, thanks for working on this! Overall the PR looks good to me. 

Only real comment is that it looks to me like you're applying the HR rate to the CWD pool rather than the fragmentation flux. Nothing wrong with that in itself but I  understood from the PR text that you were applying the HR rate to the fragmentation flux. 

Maybe I missed something and the rate is a applied to the fragmentatino flux. Bu in either case, I think this is an interesting point of discussion as each option (pool or flux application of HR rate) would have different impacts on model behaviour. Applied to the pool, HR will directly affect the C:N/P ratio of the pool and then the flux will be the C:N/P ratio of the pool. Applied to the flux, the C:N/P ratio of the pool will simply be a function of input amount and ratios, and the flux C:N/P ratio will be fung_hetresp_baserate * C:N/P.

Applied to the pool, the behaviour will be quite dynamic and C:N/P ratios of the flux will be more time-dependent and a result of the quantity of CWD litter inputs relative to the size of the pool and state of decomposition. Applied to the flux, the C:N/P of the flux will be more consistent through time and a function of the relative C:N/P ratios of the CWD inputs.        

Application to the pool may be more defensible if we assume the CWD is homogeneously colonised by fungi (which is probably a stretch). If we are going this route, a hr rate of 0.5 might be a bit high. 

Agree with @ckoven 's comment that we prob need to pass this HR flux over to the host model in order to enable accurate carbon accounting in the host. 

Also, could make the new parameter a vector ncwd long to allow future users to play with that.

parameter_files/fates_params_default.json
```diff
@@ -1682,6 +1683,13 @@
       "units": "unitless",
       "data": [0.8]
     },
+    "fates_cwd_hrfrag_frac": {
+	"dtype": "float",
+	"dims": ["scalar"],
+	"long_name": "fraction of cwd losses due to fungal heterotrophic respiration and not fragmentation",
+	"units": "-",
+	"data": [0.5]
```
rgknox, line 1691: I want to set this parameter to a default of zero before we integrate...

PR #1547, opened 2026-04-01
Update Grazing Logic (cherry-picked mirror of NorESMHub 55)

rgknox:
### Description:

This updates grazing logic.  This is a mirror of @rosiealice 's updates submitted to NorESMhub:

https://github.com/NorESMhub/fates/pull/55

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/FatesLandUseChangeMod.F90
+++ b/biogeochem/FatesLandUseChangeMod.F90
@@ -453,7 +453,7 @@ contains
     real(r8) :: crown_depth
     
     grazing_rate = landuse_grazing_rate(land_use_label) * EDPftvarcon_inst%landuse_grazing_palatability(ft)
-    if(npp_acc .le. 0.0_r8 .or. lai .lt. 1.0_r8) then
+    if(npp_acc .le. 0.0_r8 .or. lai .lt. 0.25_r8 .and. land_use_label .eq. cropland) then
        grazing_rate = 0.0_r8
     endif
     if ( grazing_rate .gt. 0._r8) then
--- a/biogeochem/FatesLandUseChangeMod.F90
+++ b/biogeochem/FatesLandUseChangeMod.F90
@@ -453,7 +453,7 @@ contains
     real(r8) :: crown_depth
     
     grazing_rate = landuse_grazing_rate(land_use_label) * EDPftvarcon_inst%landuse_grazing_palatability(ft)
-    if((npp_acc .le. 0.0_r8 .or. lai .lt. 0.25_r8) .and. land_use_label .eq. cropland) then
+    if(npp_acc .le. 0.0_r8 .or. (lai .lt. 0.25_r8 .and. land_use_label .eq. cropland)) then
 ! We allow grazing on cropland at present to loosely mimic the impact of human management on fluxes and LAI. This is a placeholder for a better future model which includes actual planting and harvest.  
 ! But if we allow grazing to occur at all times, then the crop plants eventually die in most grid cells. 
 ! Restricting grazing to when NPP is positive and LAI > a threshold prevents this in most cases. 
--- a/biogeochem/FatesLandUseChangeMod.F90
+++ b/biogeochem/FatesLandUseChangeMod.F90
@@ -453,7 +453,7 @@ contains
     real(r8) :: crown_depth
     
     grazing_rate = landuse_grazing_rate(land_use_label) * EDPftvarcon_inst%landuse_grazing_palatability(ft)
-    if(npp_acc .le. 0.0_r8 .or. lai .lt. 0.25_r8 .and. land_use_label .eq. cropland) then
+    if((npp_acc .le. 0.0_r8 .or. lai .lt. 0.25_r8) .and. land_use_label .eq. cropland) then
        grazing_rate = 0.0_r8
     endif
     if ( grazing_rate .gt. 0._r8) then
--- a/biogeochem/FatesLandUseChangeMod.F90
+++ b/biogeochem/FatesLandUseChangeMod.F90
@@ -469,7 +469,6 @@ contains
        call PRTHerbivoryLosses(prt, leaf_organ, grazing_rate)
 
     end if
-    write(*,*) 'grazing_rate',grazing_rate,lai,npp_acc
     
   end subroutine FatesGrazing
 
--- a/biogeochem/FatesLandUseChangeMod.F90
+++ b/biogeochem/FatesLandUseChangeMod.F90
@@ -429,7 +429,7 @@ contains
     
   !----------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -446,12 +446,16 @@ subroutine FatesGrazing(prt, ft, land_use_label, height)
...
     real(r8), intent(in) :: height
+    real(r8), intent(in) :: npp_acc
+    real(r8), intent(in) :: lai
 
     real(r8) :: grazing_rate    ! rate of grazing (or browsing) of leaf tissue [day -1]
     real(r8) :: crown_depth
     
     grazing_rate = landuse_grazing_rate(land_use_label) * EDPftvarcon_inst%landuse_grazing_palatability(ft)
-    
+    if(npp_acc .le. 0.0_r8 .or. lai .lt. 0.25_r8 .and. land_use_label .eq. cropland) then
```
ckoven: I can never remember the order of operations in boolean logic off the top of my head, so like to use parentheses to clarify the intended meaning. so I suggest doing that here.  

also: should we be having grazing on cropland in general, or are we generally applying it to just pasture and rangeland?

lastly, maybe add a comment explaining why this is needed / what behavior it prevents so that future us understands why we did this.

ckoven (approved):
Looks good, see comment. Thanks @rgknox, @rosiealice, and @JessicaNeedham for doing this.

biogeochem/FatesLandUseChangeMod.F90
rosiealice:
```suggestion
    if((npp_acc .le. 0.0_r8 .or. lai .lt. 0.25_r8) .and. land_use_label .eq. cropland) then
```

rosiealice: Agreed on the parentheses. Don't know why I didn't add those before!

biogeochem/FatesLandUseChangeMod.F90
```diff
@@ -446,12 +446,16 @@ subroutine FatesGrazing(prt, ft, land_use_label, height)
...
+       grazing_rate = 0.0_r8
```
rosiealice, line 483:
```suggestion
! We allow grazing on cropland at present to loosely mimic the impact of human management on fluxes and LAI. This is a placeholder for a better future model which includes actual planting and harvest.  
! But if we allow grazing to occur at all times, then the crop plants eventually die in most grid cells. 
! Restricting grazing to when NPP is positive and LAI > a threshold prevents this in most cases. 
! In principle this restriction should also apply to rangeland and pasture but we were unable to create a stable situation where LAI/GPP was realistic in all managed land with the same logic and so we only applied this change to crops. 
       grazing_rate = 0.0_r8
```

biogeochem/FatesLandUseChangeMod.F90
ckoven: I think this actually changes the logic? [Intel's table of the logical order of operations](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-2/summary-of-operator-precedence.html) lists `and` before `or`, so the old code would apply the npp filter everywhere but only the lai filter on croplands? Is that correct or am I misreading? If it is different, does that require pulling that commit back into the noresm PR also?

rosiealice:
Actually @glemieux @ckoven, @JessicaNeedham and I discussed this and I think we should keep the code working the same way as it is working in the first instance, prior to https://github.com/NGEET/fates/pull/1547/changes/47c0a8f16801fc9195c287b668b94aa5c791a8c6 
The reason is that any change at this point will require a re-tuning of the grazing rates, and I don't think the logic is more or less right/wrong either way. It makes sense that for all LUC types grazing doesn't happen if NPP is negative, and we are adding an additional case for crops where we don't allow them to be grazed to the ground. The only issue with this is that I have not yt figured out why crops need this change to avoid extinction and rangelands and pastures do not. I think it is to do with them being a distinct PFT and possibly having seed bank differences. But I never properly got to the bottom of it. But 1) they are much better in this implementation and 2) the rangelands and pastures were more productive than we would like if they were also subject to the LAI threshold. So i think in retrospect we should leave it like this until we have a chance to do further investigation.

biogeochem/FatesLandUseChangeMod.F90
glemieux: For posterity: https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vn74/index.html

TL;DR: precedence from first to last: arithmetic relation -> relational -> logical.  For logical the `.and.` evaluates before `.or.`.  With paranthesis for clarity, the above looks like:
`if(npp_acc .le. 0.0_r8 .or. (lai .lt. 0.25_r8 .and. land_use_label .eq. cropland)) then`

glemieux:
Per the software discussion today, @rosiealice confirmed we should revert back to the original behavior in (as of 783eae39397113970ed9bd1be2d6ac94e8885b88).  @ckoven confirmed he's good with moving forward with that decision.  @glemieux to update the paranthesis to reflect this decision.

biogeochem/FatesLandUseChangeMod.F90
glemieux: See https://github.com/NGEET/fates/pull/1547#issuecomment-4239057511.  Resolved with [41cd9e4](https://github.com/NGEET/fates/pull/1547/commits/41cd9e4bbe39d1d4a41114bcdf8cf087324d2f42)

glemieux (approved):

glemieux:
Regression testing on `derecho` is underway

glemieux:
Regression testing against `fates-sci.1.92.0_api.44.1.0-ctsm5.4.032` is complete ~and B4B for all expected tests~.

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_0414-105100de`

UPDATE (4/20/2026): I noticed weird run failures for `FatesColdHydro`.  I'm going to double check this.

glemieux:
> Regression testing against `fates-sci.1.92.0_api.44.1.0-ctsm5.4.032` is complete ~and B4B for all expected tests~.
> 
> Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_0414-105100de`
> 
> UPDATE (4/20/2026): I noticed weird run failures for `FatesColdHydro`. I'm going to double check this.

Upon re-review and retesting against new baseline `fates-sci.1.92.1_api.44.1.0-ctsm5.4.033` all expected runs are B4B.  The unexpected runs from last test were due to a glitch in the testing not flagging expected failures appropriately.

This is good to integrate.

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_0420-095914de`

PR #1550, opened 2026-04-03
Split initial density into to parameters with a switch on HLM side

mvdebolskiy:
### Description:
Instead of using negative numbers in `fates_recruit_init_density` for an option to coldstart nocomp with dbh instead of density. Add `fates_recruit_init_dbh` parameter that can only be used when a new switch `use_dbh_init` is on. That one has to be set on the hlm side.
This is needed to have NorESM calibrated parameters to work and be default while NOT braking any full-fates configurations. 

None, since switch is off by default.
If the switch is on, the model will break anyway with the current values for `init_seed`.

### Checklist

*Contributor*

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/main/EDPftvarcon.F90
+++ b/main/EDPftvarcon.F90
@@ -1260,19 +1260,6 @@ contains
         ! Check that in initial density is not equal to zero in a cold-start run
         !-----------------------------------------------------------------------------------
         
-        if ( hlm_use_inventory_init == ifalse .and. & 
-             abs( EDPftvarcon_inst%initd(ipft) ) < nearzero ) then
-          
-           write(fates_log(),*) ' In a cold start run initial density cannot be zero.'
-           write(fates_log(),*) ' For a bare ground run set to initial recruit density.'
-           write(fates_log(),*) ' If no-comp is on it is possible to initialize with larger  '
-           write(fates_log(),*) ' plants by setting use_fates_dbh_init to true'
-           write(fates_log(),*) ' so that initial dbh parameter is used instead of density. '
-           write(fates_log(),*) ' Aborting'
-           call endrun(msg=errMsg(sourcefile, __LINE__))
-           
-        end if
-        
         if ( EDPftvarcon_inst%initd(ipft) < -nearzero ) then
            write(fates_log(),*) ' Option to use negative values for initial density'
            write(fates_log(),*) ' has been depricated. Set use_fates_dbh_init to true'
@@ -1281,6 +1268,21 @@ contains
            write(fates_log(),*) ' Aborting'
            call endrun(msg=errMsg(sourcefile, __LINE__))
         end if
+
+        if ( hlm_use_inventory_init == ifalse .and. & 
+            EDPftvarcon_inst%initd(ipft) < nearzero ) then
+          
+           write(fates_log(),*) ' In a cold start run initial density cannot be zero'
+           write(fates_log(),*) ' without inventory initialization.'
+           write(fates_log(),*) ' If no-comp is on it is possible to initialize with larger  '
+           write(fates_log(),*) ' plants by setting use_fates_dbh_init to true'
+           write(fates_log(),*) ' so that initial dbh parameter is used instead of density. '
+           write(fates_log(),*) ' Aborting'
+           call endrun(msg=errMsg(sourcefile, __LINE__))
+           
+        end if
+        
+
 
 
         if (hlm_use_dbh_init .eq. itrue) then
--- a/main/EDPftvarcon.F90
+++ b/main/EDPftvarcon.F90
@@ -60,8 +60,10 @@ module EDPftvarcon
      real(r8), allocatable :: displar(:)             ! ratio of displacement height to canopy top height
      real(r8), allocatable :: bark_scaler(:)         ! scaler from dbh to bark thickness. For fire model.
      real(r8), allocatable :: crown_kill(:)          ! scaler on fire death. For fire model.
-     real(r8), allocatable :: initd(:)               ! initial seedling density [/m2] (positive values)
-                                                     ! or -dbh [cm] (negative values)
+     real(r8), allocatable :: initd(:)               ! initial seedling density [/m2] 
+     real(r8), allocatable :: initdbh(:)             ! initial seedling dbh [cm]
+                                                     ! alternative to initd for nocomp
+                                                     ! used only when use_dbh_init and use_nocomp are both true
      real(r8), allocatable :: init_seed(:)           ! Initial seed bank [kg/m2]
                                                      ! For SP: this is unused
                                                      ! For Nocomp: This only applies the seed from the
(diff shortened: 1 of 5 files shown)


mvdebolskiy:
CTSM-side PR:ESCOMP/CTSM#3910

parameter_files/fates_params_default.json
```diff
@@ -1272,16 +1265,23 @@
...
+      "units": "cm",
+      "data": [ 1, 1, 1, 1, 1, 1, 0.5, 0.5, 0.5, 0.5, 0.5, 0.1, 0.1, 0.1]
+    },
     "fates_recruit_init_seed": {
       "dtype": "float",
       "dims": ["fates_pft"],
       "long_name": "initial seed pool - per m2 of associated pft-patch for no-comp, per m2 of whole site for non no-comp",
       "units": "kg/m2",
-      "data": [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+      "data": [0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01]
```
glemieux: @mvdebolskiy per you conversation with Jessie, is my understanding that this is being changed to avoid a user having either `fates_recruit_init_seed` and `fates_recruit_init_density` be zero if `hlm_use_dbh_init` mode?

rgknox:
This actually brings up a broader discussion on how we maintain (paradoxically) more than 1 default file.  

We have utilities to handle this:

https://github.com/NGEET/fates/blob/main/tools/batch_patch_params.py

Note that this json file that works in conjunction with the previous script, applies a patch that turns on sized based initialization:

https://github.com/NGEET/fates/blob/main/parameter_files/patch_nocomp_noresm.json

rgknox:
We discussed this more at the CTSM dev meeting.  There were good arguments to support the idea to having dedicated parameters for DBH initialization and dedicated parameters for number initialization.  One argument was that more than one user thought it was more logical to have two parameters, and that using negative values "flipped their brain".  There was also an agument that the PPEs and parameter ranges are less confusing when they don't have two modes.  It may also simplify how we connect compsets to different parmaeter configurations since it doesn't have to modify the parameter file, and would just modify the NL variable.

It sounds like the consensus is to move ahead on this set of changes and get it in for the maintenance tag.  Lets see if anyone else chimes in here.

main/EDPftvarcon.F90
JessicaNeedham: Should we remove the abs here since we don't want initd to be negative or a tiny number? Then we could also remove the check below?

main/EDPftvarcon.F90
```diff
@@ -1265,25 +1266,39 @@ subroutine FatesCheckParams(is_master)
...
 
+
+        if (hlm_use_dbh_init .eq. itrue) then
+           if (EDPftvarcon_inst%initdbh(ipft) < nearzero) then
+              write(fates_log(),*) ' You are running in nocomp mode using initial dbh instead of density.'
+              write(fates_log(),*) ' In a cold start run initial dbh cannot be less or equal zero.'
+              write(fates_log(),*) ' Aborting'
+              call endrun(msg=errMsg(sourcefile, __LINE__))
+           endif
+           if (EDPftvarcon_inst%init_seed(ipft) < nearzero) then
```
JessicaNeedham, line 1302: Why do we need an extra check for init_seed being positive if we are in dbh initialisation mode? There's another check right below that should catch it.

JessicaNeedham (commented):
Thanks @mvdebolskiy ! These changes look good to me and I agree that having two parameters is less confusing than having one with different meanings. Only a minor question about some of the checks.

JessicaNeedham:
This will also need an update to FATES user guide. @rgknox and @glemieux if you ping me when this gets merged I can make those changes.

main/EDPftvarcon.F90
mvdebolskiy: You can not initiate the model with 0 seed and dbh. It will crash within first few days because of the canopy. Basically you are starting with fat grown tries with no carbon?

JessicaNeedham: That makes sense. I guess I was wondering if this is specific to dbh initialisation? I'm assuming the same applies if you have 0 seeds and 0 number density? So is the check here not sufficient? https://github.com/mvdebolskiy/fates/blob/a78990cfcd80856c12ec9a25d91f9df8bb409490/main/EDPftvarcon.F90#L1302

mvdebolskiy: With 0 seeds non-zero density is fine.

main/FatesInterfaceMod.F90
```diff
@@ -1683,6 +1683,14 @@ subroutine set_fates_ctrlparms(tag,ival,rval,cval)
             call endrun(msg=errMsg(sourcefile, __LINE__))
          end if
 
+         if (  .not.((hlm_use_dbh_init.eq.1).or.(hlm_use_dbh_init.eq.0))    ) then
```
glemieux: This should use the named variables for true/false integers:
```suggestion
         if (  .not.((hlm_use_dbh_init.eq.itrue).or.(hlm_use_dbh_init.eq.ifalse))    ) then

glemieux (changes requested):
Looks good!  One minor requested change and one question asking for clarification.

parameter_files/fates_params_default.json
glemieux: Per the ctsm-fates standup today, @mvdebolskiy is going to revert this change.  He will add a test in ctsm that updates this parameter on the fly.  See https://github.com/ESCOMP/CTSM/pull/3910#issuecomment-4372872257.

glemieux: @rgknox noted during the fates SE meeting today that we could change this to non-zero value, but we would like to see that done in a dedicated parameter file update, not in this PR.

glemieux (approved):

glemieux:
Starting `fates` regression tests on `derecho`.

mvdebolskiy:
One small thing I forgot, and am afk now. Can you swap dbh and density in the parameter file? It's alphabetical prder and b shold be earlier than e.

mvdebolskiy:
@glemieux disregard comment above, fixed it.

glemieux:
Regression testing on `derecho` against `fates-sci.1.92.4_api.45.0.0-ctsm5.4.037` is B4B for expected tests.  

Results: `/glade/u/home/glemieux/scratch/ctsm-tests/tests_0506-154524de`

Running tests on `izumi` now.

glemieux:
Regression testing on `izumi` against `fates-sci.1.92.4_api.45.0.0-ctsm5.4.037` is complete.  All expected tests are b4b.

Results: `/scratch/cluster/glemieux/ctsm-tests/tests_0508-110832iz`

PR #1568, opened 2026-05-11
Deactivate history fields in SP mode

mvdebolskiy:
Use a bool switch to deactivate default fields in SP-mode, so that on the HLM-side long hist_fexl* are not used.

### Description:

See above

### Collaborators:

@rgknox @glemieux 

For SP-test a few extra fields will be deactivated, since f.e. in CTSM tests,  the FATES-SP usermod has not been updated in a while.

### Description of generative AI usage (as necessary)

None.

### Checklist

*Contributor*
- [x] Describe use of generative AI (if necessary)

*Integrator*

### Documentation

### Test Results:

Only field list will change for SP tests with:
```
 Could not find match for file1 variable FATES_SEED_BANK_LUPF in file2
 Could not find match for file1 variable FATES_SEED_BANK_PF in file2
```
Tested with ctsm's `FatesColdSatPhen` testmod. with commenting out exclude in the fates-sp usermod vs default tag. Test with `FatesCold` and removing hist_fincl and empty_tapes showed no diffs.

*CTSM (or) E3SM (specify which) test hash-tag: ctsm5.4.038

*CTSM (or) E3SM (specify which) baseline hash-tag: ctsm5.4.038

*FATES baseline hash-tag: sci.1.92.4_api.45.0.0

*Test Output:*

### DIFF
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -8606,7 +8606,7 @@ contains
 
           call this%set_history_var(vname='FATES_VEGC_SZ', units = 'kg m-2',         &
                long='total biomass by size class in kg carbon per m2',              &
-               use_default=trim(drop_in_sp), avgflag='A', vtype=site_size_r8,             &
+               use_default='inactive', avgflag='A', vtype=site_size_r8,             &
                hlms='CLM:ALM', upfreq=group_dyna_complx, ivar=ivar,                                 &
                initialize=initialize_variables, index = ih_biomass_si_scls)
 
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -6571,7 +6571,7 @@ contains
     ! cohort size x crown damage       (site_cdsc_r8)     : SZCD
     ! cohort size x crown damage x pft (site_cdpf_r8)     : CDPF
 
-    if (hlm_use_sp) then
+    if (hlm_use_sp .eq. itrue) then
       drop_in_sp = 'inactive'
     else
       drop_in_sp = 'active'
--- a/main/FatesHistoryInterfaceMod.F90
+++ b/main/FatesHistoryInterfaceMod.F90
@@ -6537,6 +6537,7 @@ contains
 
     integer :: ivar
     character(len=10) :: tempstring
+    character(len=10) :: drop_in_sp
 
     ivar=0
 
@@ -6570,6 +6571,11 @@ contains
     ! cohort size x crown damage       (site_cdsc_r8)     : SZCD
     ! cohort size x crown damage x pft (site_cdpf_r8)     : CDPF
 
+    if (hlm_use_sp) then
+      drop_in_sp = 'inactive'
+    else
+      drop_in_sp = 'active'
+    endif
 
     if_dyn0: if(hlm_hist_level_dynam>0) then
 
@@ -6589,7 +6595,7 @@ contains
        ! Patch variables
        call this%set_history_var(vname='FATES_TRIMMING', units='1',               &
             long='degree to which canopy expansion is limited by leaf economics (0-1)', &
-            use_default='active', avgflag='A', vtype=site_r8, hlms='CLM:ALM',     &
+            use_default=trim(drop_in_sp), avgflag='A', vtype=site_r8, hlms='CLM:ALM',     &
             upfreq=group_dyna_simple, ivar=ivar, initialize=initialize_variables,                 &
             index=ih_trimming_si)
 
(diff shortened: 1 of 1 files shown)


rgknox:
Love this change

main/FatesHistoryInterfaceMod.F90
```diff
@@ -6907,29 +6913,29 @@ subroutine define_history_vars(this, initialize_variables)
 
        call this%set_history_var(vname='FATES_FROOTC', units='kg m-2',            &
             long='total biomass in live plant fine roots in kg carbon per m2',    &
-            use_default='active', avgflag='A', vtype=site_r8, hlms='CLM:ALM',     &
+            use_default=trim(drop_in_sp), avgflag='A', vtype=site_r8, hlms='CLM:ALM',     &
```
rosiealice, line 6916: These allometric C structure variables are, I think, relevant/used in SP mode. The cohort is defined based on the height and the other pools are set accordingly. So there should be valid values for VEGC, SAPWOODC, FROOTC, STRUCTC, NONSTRUCTC

main/FatesHistoryInterfaceMod.F90
```diff
@@ -7386,7 +7392,7 @@ subroutine define_history_vars(this, initialize_variables)
           
           call this%set_history_var(vname='FATES_VEGC_PF', units='kg m-2',           &
                long='total PFT-level biomass in kg of carbon per land area',         &
-               use_default='active', avgflag='A', vtype=site_pft_r8, hlms='CLM:ALM', &
+               use_default=trim(drop_in_sp), avgflag='A', vtype=site_pft_r8, hlms='CLM:ALM', &
```
rosiealice, line 7395: This should exist in SP mode.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -8627,7 +8633,7 @@ subroutine define_history_vars(this, initialize_variables)
 
           call this%set_history_var(vname='FATES_LAI_CANOPY_SZ', units = 'm2 m-2',   &
                long='leaf area index (LAI) of canopy plants by size class',         &
-               use_default='active', avgflag='A', vtype=site_size_r8,               &
+               use_default=trim(drop_in_sp), avgflag='A', vtype=site_size_r8,               &
```
rosiealice, line 8636: In principle the plants in SP mode do have a height and and LAi and so in theory this has some meaning...

main/FatesHistoryInterfaceMod.F90
mvdebolskiy: @rosiealice but it currently does no.
https://github.com/ESCOMP/CTSM/blob/e1f563b8345400d33421263f05b237f13e6b0e3d/cime_config/usermods_dirs/clm/fates_sp/user_nl_clm#L13

main/FatesHistoryInterfaceMod.F90
```diff
@@ -8687,48 +8693,48 @@ subroutine define_history_vars(this, initialize_variables)
                initialize=initialize_variables, index = ih_sai_understory_si_scls)
 
           call this%set_history_var(vname='FATES_NPLANT_SZ', units = 'm-2',          &
-               long='number of plants per m2 by size class', use_default='active',  &
+               long='number of plants per m2 by size class', use_default=trim(drop_in_sp),  &
```
rosiealice, line 8696: This might also be useful as a diagnostic to check that the plants are coming out in the right size class given HTOP...

rosiealice (commented):
Looks great @mvdebolskiy ... Thanks for pushing this back.

mvdebolskiy:
@rosiealice I don't quite get the comments. If the variables there "should be active" why are they exluded in the `usermod` that is getting picked up by FATES-SP compsets? Do you need this variables ouputted all the time you have an SP run, or can they just be requested by user whenever user needs them?

rosiealice:
> @rosiealice I don't quite get the comments. If the variables there "should be active" why are they exluded in the `usermod` that is getting picked up by FATES-SP compsets? Do you need this variables ouputted all the time you have an SP run, or can they just be requested by user whenever user needs them?

Fair enough. It's not critical either way. I guess i just didn't think about these reasons for keeping them last time I went through the variable list (I realize I'm critiquing my own earlier classification effort here, so should have been clearer about that :) ) .

main/FatesHistoryInterfaceMod.F90
rgknox: I don't think we calculate it though. 

WHen in SP or ST3 mode, we perform a bypass of the whole allocation algorithm:

https://github.com/NGEET/fates/blob/main/main/EDMainMod.F90#L1133

Without calling it, there is no VEGC.  We could certainly improve the bypass though to set VEGC to allometric targets though....

main/FatesHistoryInterfaceMod.F90
rgknox: But there is only one size ("SZ") right? I see this diagnostic as not wrong, just unneeded with SP.

main/FatesHistoryInterfaceMod.F90
rgknox: See comment below.

main/FatesHistoryInterfaceMod.F90
```diff
@@ -6570,6 +6571,11 @@ subroutine define_history_vars(this, initialize_variables)
     ! cohort size x crown damage       (site_cdsc_r8)     : SZCD
     ! cohort size x crown damage x pft (site_cdpf_r8)     : CDPF
 
+    if (hlm_use_sp) then
```
mvdebolskiy:
```suggestion
    if (hlm_use_sp .eq. itrue) then
```

main/FatesHistoryInterfaceMod.F90
ckoven: I actually think this might be useful, even in SP. since the height bins aren't the height of the plant, but the height of the leaf area distributions of the plants, so there could in principle be some information here

rosiealice:
Just for now, none of this is super critical and we can always add these variables back in later if necessary. I think we can pull this PR as it is...

rosiealice (commented):
I think this is really OK as it is, for the sake of time. We can always modify the details later, but this is necessary for the whole FATES compset workflow and so it's probably best to include it sooner rather than later (thus retracting my earlier nitpicking over the inclusion of certain fields :) )

rosiealice (approved):

ckoven:
> I think this is really OK as it is, for the sake of time. We can always modify the details later, but this is necessary for the whole FATES compset workflow and so it's probably best to include it sooner rather than later (thus retracting my earlier nitpicking over the inclusion of certain fields :) )

agreed

ckoven (approved):

glemieux:
Regression testing underway on `derecho`.

mvdebolskiy:
Btw, this needs an acompaning PR that removes the hist_fexl statement from fates-sp usermod or a PR with refactor for ctsm compsets NorESMhub/CTSM#224 the answers won't be b4b, since there are more fields removed than in the usermod.

wwieder:
Can you remind me, Is this something we want to integrate for the CLM-FATES freeze?  If so, is there a CLM side issue or PR that we can use for tracking?

glemieux:
> Can you remind me, Is this something we want to integrate for the CLM-FATES freeze? If so, is there a CLM side issue or PR that we can use for tracking?

@wwieder I don't think there is an ctsm-side issue yet.  Has there been discussion with the CTSM SE team about adopting the FATES SP mode build namelist BGC flag option that NorESM implemented in the PR https://github.com/NorESMhub/CTSM/pull/224 @mvdebolskiy mentioned above (i.e. moving away from the user_mod setup for FATES-SP complist).
https://github.com/mvdebolskiy/CTSM/blob/fb55d3470691726315d986a63fd9f950e85f59c9/bld/CLMBuildNamelist.pm#L88-L108

I've got a branch just update the user_mod per Matvey's reminder if that's the preference instead.

mvdebolskiy:
@wwieder I've tested my noresm-compsets PR that also reduces some fates usermod burden on derecho and it did not affect clm-test's namelists in aux_clm. The only issue with it is that I need a PR to put drydep and bvoc's into ngeet from NorESM fork, in order to have CTSM-side working. And the last question is how many fates_parameter files we need to maintain. Plus, the initial datasets/landuse forcing is not finalized yet, since we are fixing some bugs with tiny numbers causing balance errors.

glemieux:
> Btw, this needs an acompaning PR that removes the hist_fexl statement from fates-sp usermod 

This is waiting on approval of escomp/ctsm#4079.

Issue #1499, opened 2025-10-28
PAR fraction calculation causes model to crash when using Tree Recruitment Scheme with seedling dynamics


r-ward:
### Describe the issue

The `SeedlingParPatch` subroutine uses absolute canopy area (m2) instead of fractional area when calculating `par_high_frac` and `par_low_frac`, causing negative values that produce NaNs in seedling light calculations.

https://github.com/NGEET/fates/blob/3b58472a513e39fcb8d0dafc7bc0ae31b91f6e95/main/FatesInterfaceMod.F90#L2411-L2414

Solution: normalize by patch area 
```par_high_frac = 1._r8 - (cpatch%total_canopy_area / cpatch%area)```
```par_low_frac = cpatch%total_canopy_area / cpatch%area ``` 

### Relevant log output

```shell
--WARNING-- skipping CN balance check for first timesteps after startup or data assimilation
 --WARNING-- skipping CN balance check for first timesteps after startup or data assimilation
 --WARNING-- skipping CN balance check for first timesteps after startup or data assimilation
 mass balance error detected
 element type (see PRTGenericMod.F90):            1
 error fraction relative to biomass stock:    0.0000000000000000     
 absolut error (flux in - change):                        NaN
 call index:            1
 Element index (PARTEH global):           1
 net:                        NaN
 dstock:                        NaN
 seed_in:    2.7397260273972605E-002
 net_root_uptake:    0.0000000000000000     
 gpp_acc:    1.8857599694259481     
 flux_generic_in:    0.0000000000000000     
 wood_product_harvest:    0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000     
 wood_product_landusechange:    0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000     
 error from patch resizing:    0.0000000000000000     
 burn_flux_to_atm:    0.0000000000000000     
 seed_out:    0.0000000000000000     
[...]

Issue #1504, opened 2025-11-04
Seed pool can become negative when using TRS with seedling dynamics, eventually crashing fates


r-ward:
### Describe the issue

When running FATES with the Tree Recruitment Scheme (TRS) and seedling dynamics, the seed pool can become negative, eventually crashing the model. I think this is due to two problems in the calculation of `seed_germ_in` (the mass germinating out of the seed pool). TRS then uses `seed_germ` for its construction costs. 

**Background:** The seed pool is updated in the PreDisturbanceIntegrateLitter Mod: 
https://github.com/NGEET/fates/blob/75a465c0b650fad9e40059204903f6017f22a1ee/biogeochem/EDPhysiologyMod.F90#L544-L552

When TRS with seedling dynamics is enabled, `seed_decay` includes non-reproductive biomass from in-fluxing litter and decay from the existing seed bank.  

And `seed_germ_in` = an environmentally sensitive fraction of the seed pool: `seed` * `seedling_emerg_rate` (Step 4 below):

https://github.com/NGEET/fates/blob/75a465c0b650fad9e40059204903f6017f22a1ee/biogeochem/EDPhysiologyMod.F90#L2402-L2445

**There are two problems here:** 
**Problem 1.** Order of operations: when `seedling_emerg_rate` is high, the updated seed pool can become negative if (seed pool * seed_emerg_rate)  > (seed pool - seed_decay)   
- Potential fix: calculate in two steps to make sure that the material in seed_germ_in does not exceed the available seed material after decay e.g.
```
! Calculate available seed after inputs and decay
   net_seed_available = litt%seed(pft) + litt%seed_in_local(pft) + &
                       litt%seed_in_extern(pft) - litt%seed_decay(pft)
   
   ! Cap germination at available seed
   litt%seed_germ_in(pft) = min(litt%seed_germ_in(pft), net_seed_available))
   
   ! Update pools
   litt%seed(pft) = net_seed_available - litt%seed_germ_in(pft)
   litt%seed_germ(pft) = litt%seed_germ(pft) + litt%seed_germ_in(pft) - litt%seed_germ_decay(pft)
```
   
**Problem 2.** The `seedling_emerg_rate` is currently not capped, and can be > 1,  which contributes to problem 1 above.   
- Potential fix: cap seedling emergence at 1?

As an example, I'm pasting log output from a day 353 of single-point run at BCI with 2-pfts.
[...]

Issue #1501, opened 2025-10-28
Seedling mortality PAR initialized with wrong variable when using Tree Recruitment Scheme with seedling dynamics


r-ward:
### Describe the issue

The initial value for `sdlng_mort_par` is set to `temp_init_veg' rather than `init_seedling_par`. 

https://github.com/NGEET/fates/blob/3b58472a513e39fcb8d0dafc7bc0ae31b91f6e95/biogeochem/FatesPatchMod.F90#L642-L646

https://github.com/NGEET/fates/blob/3b58472a513e39fcb8d0dafc7bc0ae31b91f6e95/biogeochem/FatesPatchMod.F90#L665-L671

### Relevant log output

```shell

```

### FATES tag

sci.1.87.5_api.41.0.0-12-gf2c22336

### Host land model tag

ctsm5.3.077

### Machine

derecho

### Additional context

I have fixes for this and couple of additional bugs (issues #1500 , #1499 ), on my branch [debug_trs](https://github.com/r-ward/fates/tree/trs_debug)

Issue #1500, opened 2025-10-28
First-day host model SMP cause 100% seedling mortality when using the Tree Recruitment Scheme with seedling dynamics


r-ward:
### Describe the issue

Host model soil matric potential (SMP) spikes to ~negative 115 trillion on model day 1, which is remembered by the exponential moving averages that track accumulated moisture deficit `new_seedling_mdd` and `new_seedling_layer_smp`. The result is >100% moisture stress mortality rate, crashing fates when using the tree recruitment scheme with seedling dynamics.

Moisture deficit is accumulated in the UpdateFatesRMeansTStep subroutine. I'm pasting log output that shows the SPM spike (see log for c_index=2, by c_index=9 or 10 SMP returns to what seems to be a more reasonable value). Moisture deficit (`new_seedling_mdd`) is accumulated when deficit is beyond a critical threshold `seedling_psi_crit` and then added to the running mean `Current EMA c_mean`: 

https://github.com/NGEET/fates/blob/3b58472a513e39fcb8d0dafc7bc0ae31b91f6e95/main/FatesInterfaceMod.F90#L2300C1-L2344C1

The moisture stress mortality rate is calculated using the running mean of seedling_mdd here: 
https://github.com/NGEET/fates/blob/3b58472a513e39fcb8d0dafc7bc0ae31b91f6e95/biogeochem/EDPhysiologyMod.F90#L2316C1-L2344C32

On my branch I've avoided the problem by skipping the running mean updates on model day 1: 

https://github.com/r-ward/fates/blob/f2c2233661320102a9eb46aed331e33a2c12936f/main/FatesInterfaceMod.F90#L2338C1-L2343C23

...but I'm not sure if that's the best solution, or if the SMP spike is concerning in itself! 

### Relevant log output

```shell
------------------------------------------------------------
 Successfully initialized the land model
 begin initial run at: 
    nstep=            1  year=         1985  month=            1  day=            1  seconds=         1800

************************************************************

 === MDD DEBUG Day:   1.0208333333333333     
 PFT:           1 Soil layer:           2
 Raw SMP from host:   0.0000000000000000     
 new_seedling_layer_smp:   0.0000000000000000     
 seedling_psi_crit:  -251995.70000000001     
 Deficit check val for check   251995.70000000001     
 new_seedling_mdd (after zero check):   0.0000000000000000     
 Current EMA c_mean:   0.0000000000000000     
 Current c_index:           1
 n_mem:        6048
[...]

PR #1572, opened 2026-05-27
Fix bugs in the Tree Recruitment Scheme with seedling dynamics

r-ward:
### Description:
Fixes several bugs that cause the model to crash when running FATES with the Tree Recruitment Scheme (TRS; namelist option `fates_regeneration_model = 'trs'`). 

#### Bug fixes: 
#1499 - Fix PAR fraction calculations (normalization by patch area was previously missing, causing NaN errors)
#1500 - Skip updating seedling layer SMP running mean on first model day (host model SMP spike leads to 100% seedling mortality; the host model spike itself is a separate issue: https://github.com/ESCOMP/CTSM/issues/3584#issue-3591631905)
#1501 - Fix initial value for `sdlng_mort_par` (was set to `temp_init_veg`,  should be `init_seedling_par`)
#1504 - Add rate caps for TRS seedling emergence and decay rates so that they cannot exceed 1; rates > 1 unrealistically drive the TRS seedling pool negative, eventually crashing the model

### Collaborators:
@rgknox @glemieux 

Results using other regeneration models (`default`, `trs_no_seed_dyn`) are unchanged, results will differ when using `fates_regeneration_model='trs'` (model previously crashed)

image

### Description of generative AI usage (as necessary)
None

### Checklist

*Contributor*
- [x] Describe use of generative AI (if necessary)

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -2340,12 +2340,11 @@ contains
           
           ! Step 3. Sum modes of mortality (including background mortality) and send dead seedlings
           ! to litter        
-          ! Cap total mortality rate at 1 to prevent negative seed pool 
-           litt%seed_germ_decay(pft) = litt%seed_germ(pft) * &
-               min(1.0_r8, seedling_light_mort_rate + &
-                           seedling_h2o_mort_rate + &
-                           (EDPftvarcon_inst%background_seedling_mort(pft) * &
-                           years_per_day))
+          ! Cap total mortality rate at 1 to prevent negative seedling pool
+          litt%seed_germ_decay(pft) = litt%seed_germ(pft) * &
+            min(1.0_r8, seedling_light_mort_rate + &
+                        seedling_h2o_mort_rate + &
+                        (EDPftvarcon_inst%background_seedling_mort(pft) * years_per_day))
 
        else
           
@@ -2449,10 +2448,11 @@ contains
 
           ! If SMP is below a pft-specific value, then no germination occurs
           if ( seedling_layer_smp .GE. EDPftvarcon_inst%seedling_psi_emerg(pft) ) then
+             seedling_emerg_rate = photoblastic_germ_modifier * EDPftvarcon_inst%a_emerg(pft) * &
+                  wetness_index**EDPftvarcon_inst%b_emerg(pft)
              ! Cap emergence rate at 1, rate can exceed 1 for some parameter combinations,
              ! leading to negative seed bank
-             seedling_emerg_rate = min(1.0_r8, photoblastic_germ_modifier * EDPftvarcon_inst%a_emerg(pft) * &
-                  wetness_index**EDPftvarcon_inst%b_emerg(pft))
+             seedling_emerg_rate = min(1.0_r8, seedling_emerg_rate )
           else 
 
              seedling_emerg_rate = 0.0_r8
@@ -2683,7 +2683,7 @@ contains
                      mass_avail = currentPatch%area*                           &
                         currentPatch%litter(el)%seed_germ(ft)*                 & 
                         EDPftvarcon_inst%seedling_light_rec_a(ft)*             &
-                        sdlng2sap_par**EDPftvarcon_inst%seedling_light_rec_b(ft) 
+                        sdlng2sap_par**EDPftvarcon_inst%seedling_light_rec_b(ft)
                      
                      ! If soil moisture is below pft-specific seedling  moisture stress threshold the 
                      ! recruitment does not occur.
--- a/biogeochem/EDPhysiologyMod.F90
+++ b/biogeochem/EDPhysiologyMod.F90
@@ -541,6 +541,7 @@ contains
     integer :: nlevsoil    ! number of soil layers
     integer :: ilyr        ! soil layer loop counter
     integer :: dcmpy       ! decomposability index
+    real(r8) :: net_seed_available  ! Available seed after inputs and decay
 
     do el = 1, num_elements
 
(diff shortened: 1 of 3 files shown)


biogeochem/EDPhysiologyMod.F90
```diff
@@ -551,15 +551,20 @@ subroutine PreDisturbanceIntegrateLitter(currentPatch)
...
-          ! Note that the recruitment scheme will use seed_germ
-          ! for its construction costs.
-          litt%seed_germ(pft) = litt%seed_germ(pft) + &
+
+         ! Calculate net available seed after inputs and decay
+         net_seed_available = litt%seed(pft) + litt%seed_in_local(pft) + &
+            litt%seed_in_extern(pft) - litt%seed_decay(pft)
+
+         ! Cap germination at available seed to prevent negative seed pool
+         litt%seed_germ_in(pft) = min(litt%seed_germ_in(pft), net_seed_available)
```
rgknox, line 560: @r-ward , I'm wondering if this seed_germ_in could potentially be negative.  I don't think it can since seed decay is the only negative term in net_seed_available, and it is by definition a fraction of seed mass.  Can you verify this?

biogeochem/EDPhysiologyMod.F90
```diff
@@ -2279,79 +2287,96 @@ subroutine SeedDecay( litt , currentPatch )
...
              seedling_h2o_mort_rate = 0.0_r8
           else
              seedling_h2o_mort_rate = EDPftvarcon_inst%seedling_h2o_mort_a(pft) * seedling_mdds**2 + &
                   EDPftvarcon_inst%seedling_h2o_mort_b(pft) * seedling_mdds + &
                   EDPftvarcon_inst%seedling_h2o_mort_c(pft)
+             ! Cap h2o mortality rate at 1, quadratic can produce values >1 for large moisture
+             ! deficit days, if rate exceeds 1 write to fates log 
+
+             if (seedling_h2o_mort_rate > 1.0_r8) then
+               write(fates_log(), *) 'TRS seedling_h2o_mort_rate > 1', &
```
rgknox: logging messages that don't have terminations associated with them are most helpful in my experience while actively debugging. Another alternative is to use FatesWarn()  which will count the times this situation triggers and summarize at the end of a run. Another alternative would be create a history variable that tracks the mort rate in this situation.

biogeochem/EDPhysiologyMod.F90
```diff
@@ -2279,79 +2287,96 @@ subroutine SeedDecay( litt , currentPatch )
...
-               (litt%seed_germ(pft) * seedling_h2o_mort_rate) + &
-               (litt%seed_germ(pft) * EDPftvarcon_inst%background_seedling_mort(pft) &
-               * years_per_day)
+          ! to litter        
+          ! Cap total mortality rate at 1 to prevent negative seedling pool
+          total_seedling_mort_rate = seedling_light_mort_rate + &
+                        seedling_h2o_mort_rate + &
+                        (EDPftvarcon_inst%background_seedling_mort(pft) * years_per_day)
+          if (total_seedling_mort_rate > 1.0_r8) then 
+            write(fates_log(), *) 'TRS total seedling mortality rate > 1', &
```
rgknox: same comment as above

main/FatesInterfaceMod.F90
```diff
@@ -2361,8 +2335,12 @@ subroutine UpdateFatesRMeansTStep(sites,bc_in, bc_out)
                  endif
 
                  ! Update the seedling layer smp and mdd running means
-                 call cpatch%sdlng_emerg_smp(pft)%p%UpdateRMean(new_seedling_layer_smp)
-                 call cpatch%sdlng_mdd(pft)%p%UpdateRMean(new_seedling_mdd)
+                 ! Only update after first model day to avoid adding
+                 ! initialization spike in smp to running means
+                 if (hlm_model_day > 2.0_r8) then 
+                   call cpatch%sdlng_mdd(pft)%p%UpdateRMean(new_seedling_mdd)
+                   call cpatch%sdlng_emerg_smp(pft)%p%UpdateRMean(new_seedling_layer_smp)
```
rgknox, line 2342: We want to make sure that hlm_model_day is continuous across restarts. If for some reason someone is running two day restarts, and hlm_model_day is resetting each time, this running mean will never increment with meaningful data. 

While I don't think SMP spikes are unexpected at the beginning of a simulation, and your move to ignore them makes sense.  But I also wanted to check in regarding frozen soils.  I think there was some discussion of it here:

https://github.com/NGEET/fates/issues/1500

SHould we apply some filters to prevent calculations when soils are frozen?

rgknox (changes requested):

biogeochem/EDPhysiologyMod.F90
r-ward: Correct, seed_germ_in is never negative. Although, I guess it is possible for the flux to go negative if fates_recruit_seed_germination_rate or fates_trs_seedling_a_emerg were erroneously set as negative values. Maybe there should be a check on this in FatesCheckParams (I see for example a check against values of seed_dispersal_fraction < 0). I could add something similar. What do you think?

biogeochem/EDPhysiologyMod.F90
r-ward: Makes sense, will remove. I'm working on a separate PR to add history variables for some TRS-specific output; I'll add history variables or FatesWarn() messages there.

main/FatesInterfaceMod.F90
r-ward: Ok, I see. Re: frozen soils - I think you're right that we need some conditionality, but I don't think we should stop calculations when soils are frozen. We could do something analogous to the phenology code (https://github.com/r-ward/fates/blob/3b58472a513e39fcb8d0dafc7bc0ae31b91f6e95/biogeochem/EDPhysiologyMod.F90#L1222)) and use a lower bound value similar to `smp_lwr_bound` for cases where `check_layer_water` finds no liquid water.  Frozen conditions are physiologically dry, so I think that makes sense (although really we should implement behavior dependent on moisture _and_ chilling degree days for temperate/boreal forest uses of TRS).  

In the tropics, though, using an `smp_lwr_bound` during the SMP initialization spike suppresses germination and inflates seedling mortality for weeks/months, so this fix alone doesn't work.  

Can we differentiate real frozen soils from initialization/erroneous freezing by adding a check on soil temperature? If soil temperature is warm and there is no liquid water, skip the calculations. If soil temperature is freezing, and there is no liquid water, use an smp_lwr_bound.  @lmkueppers - what do you think?

biogeochem/EDPhysiologyMod.F90
rgknox: I think its fine then, resolved

biogeochem/EDPhysiologyMod.F90
rgknox: resolved, thanks

main/FatesInterfaceMod.F90
rgknox: @glemieux tracked the model day, and it is indeed a restarted variable. So we should be fine here. I think the frozen soil stuff is an issue that is outside the scope of this PR.  resolved

rgknox (approved):

PR #1577, opened 2026-06-14
Interstitial bareground

glemieux:
Changes how FATES handles "interstitial" bareground.  Interstitial bareground in this context means the fraction of a given FATES patch area that is not covered by vegetation.   

### Description:

CLM associated pull request is https://github.com/ESCOMP/CTSM/pull/4094

Relevant reference: [CTSM technical note on MOST](https://escomp.github.io/CTSM/tech_note/Fluxes/CLM50_Tech_Note_Fluxes.html#monin-obukhov-similarity-theory)  

### Collaborators:
@ckoven @rgknox 

Yes, but only when `use_fates_interstitial_bareground` is set to `false` (`true` is the default)

### Checklist

*Contributor*
- [ ] Describe use of generative AI (if necessary)

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -36,6 +36,7 @@ module EDCanopyStructureMod
   use FatesInterfaceTypesMod     , only : hlm_use_planthydro
   use FatesInterfaceTypesMod     , only : hlm_use_cohort_age_tracking
   use FatesInterfaceTypesMod     , only : hlm_use_sp
+  use FatesInterfaceTypesMod     , only : hlm_use_interstitial_bareground
   use FatesInterfaceTypesMod     , only : numpft
   use FatesInterfaceTypesMod, only : bc_in_type
   use FatesPlantHydraulicsMod, only : UpdateH2OVeg,InitHydrCohort, RecruitWaterStorage
@@ -1351,6 +1352,7 @@ contains
     integer :: s, ifp, c, p
     type (fates_patch_type)  , pointer :: currentPatch
     real(r8) :: bare_frac_area
+    real(r8) :: local_patch_fraction
     real(r8) :: total_patch_area
     real(r8) :: total_patch_leaf_stem_area
     real(r8) :: weight  ! Weighting for cohort variables in patch
@@ -1358,7 +1360,8 @@ contains
     do s = 1,nsites
 
        total_patch_area = 0._r8
-       bc_out(s)%canopy_fraction_pa(:) = 0._r8
+       local_patch_fraction = 0._r8
+       bc_out(s)%patch_fraction(:) = 0._r8
        bc_out(s)%dleaf_pa(:) = 0._r8
        bc_out(s)%z0m_pa(:) = 0._r8
        bc_out(s)%displa_pa(:) = 0._r8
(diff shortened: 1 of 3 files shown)


biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1445,19 +1445,25 @@ subroutine update_hlm_dynamics(nsites,sites,fcolumn,bc_out)
...
              else
-                bc_out(s)%canopy_fraction_pa(ifp) = 0.0_r8
+                bc_out(s)%patch_fraction(ifp) = 0.0_r8
              endif
 
-             bare_frac_area = (1.0_r8 - min(1.0_r8,currentPatch%total_canopy_area/currentPatch%area)) * &
+             bare_frac_area = (1.0_r8 - local_patch_fraction) * &
                   (currentPatch%area/AREA)
 
-             total_patch_area = total_patch_area + bc_out(s)%canopy_fraction_pa(ifp) + bare_frac_area
```
glemieux, line 1458: Note that this is dead and duplicate code being removed.  `total_canopy_area` wasn't being used downstream.

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1433,9 +1443,13 @@ subroutine update_hlm_dynamics(nsites,sites,fcolumn,bc_out)
...
                 ! if no canopy, then use dummy values (first PFT) of aerodynamic properties
-                bc_out(s)%z0m_pa(ifp)    = EDPftvarcon_inst%z0mr(1) * bc_out(s)%htop_pa(ifp)
-                bc_out(s)%displa_pa(ifp) = EDPftvarcon_inst%displar(1) * bc_out(s)%htop_pa(ifp)
-                bc_out(s)%dleaf_pa(ifp)  = EDPftvarcon_inst%dleaf(1)
+                if (hlm_use_interstitial_bareground .eq. itrue) then
+                   ! do something else?
+                else
+                   bc_out(s)%z0m_pa(ifp)    = EDPftvarcon_inst%z0mr(1) * bc_out(s)%htop_pa(ifp)
+                   bc_out(s)%displa_pa(ifp) = EDPftvarcon_inst%displar(1) * bc_out(s)%htop_pa(ifp)
+                   bc_out(s)%dleaf_pa(ifp)  = EDPftvarcon_inst%dleaf(1)
```
glemieux, line 1464: Q: Is this still valid regardless of the inclusion of interstitial bareground or not?  Or do we need a different calculation if there truly is no canopy?  What do HLM's use for pure bareground in this case?

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -1389,17 +1389,27 @@ subroutine update_hlm_dynamics(nsites,sites,fcolumn,bc_out)
...
+               else
+                 weighting_area = currentPatch%total_canopy_area
+               end if
                 currentCohort => currentPatch%shortest
                 do while(associated(currentCohort))
                    if (currentCohort%canopy_layer .eq. 1) then
-                      weight = min(1.0_r8,currentCohort%c_area/currentPatch%total_canopy_area)
+                      weight = min(1.0_r8,currentCohort%c_area/weighting_area)
                       bc_out(s)%z0m_pa(ifp) = bc_out(s)%z0m_pa(ifp) + &
                            EDPftvarcon_inst%z0mr(currentCohort%pft) * currentCohort%height * weight
```
glemieux, line 1421: Q: Does more of this calculation need to change if we are including the interstitial bareground?

biogeochem/EDCanopyStructureMod.F90
glemieux: Note that this needs to be changed as skipping this sets `z0m_pa` to zero in the host land model and causes a run failure in ch4mod due to (I think) a division by zero in the friction velocity calculation causing `h20soi_vol` to become  `NaN` downstream.  Does that sound likely to you @rgknox ?

glemieux: I confirmed that this is what's happening.  So for when we have no canopy on a patch, my initial thoughts are:
1. Use the column-level `z0mg` passed in from the HLM as a new `bc_in` value (this is what the BaregroundFluxes calculation uses in CLM).
2. Stick with the current method of calculating from `htop`, which I believe will be `0.1` for a patch without canopy.

I'll try and run some comparisons and see what the resulting friction velocity output looks like

glemieux: Per conversation at the SE meeting today, the consensus is leaning towards making fates aware of the bareground roughness length scale.  The next question, is can/should we, as a near-term development, pass the bareground roughness itself during this case?

glemieux: > The next question, is can/should we, as a near-term development, pass the bareground roughness itself during this case?

Consensus is yes.  This has been added here and in the relevant HLM PR.

PR #1579, opened 2026-06-22
Patch fusion crash in historic land only simulation fix

glemieux:
### Description:

Changing EDCanopyStructureMod.F90 to have relative rather than absolute error comparisons so that errors are not triggered by differences which are in the noise of r8 precisions numbers

Addresses https://github.com/NorESMhub/CTSM/issues/223.  Cherry-picked from NorESMhub/ctsm#66.

### Collaborators:

@maritsandstad 

### Checklist

*Contributor*
- [ ] Describe use of generative AI (if necessary)

*Integrator*

### Documentation

### Test Results:

*FATES baseline hash-tag:*

*Test Output:*

### DIFF
--- a/biogeochem/EDCanopyStructureMod.F90
+++ b/biogeochem/EDCanopyStructureMod.F90
@@ -85,6 +85,18 @@ module EDCanopyStructureMod
   
   real(r8), parameter :: co_area_target_precision = 1.0E-9_r8 
 
+  ! Relative precision target used alongside the absolute targets above.
+  !    Area conservation checks compare differences of two large, similarly
+  !    sized areas (e.g. cohort crown areas, layer areas, patch areas). For
+  !    such differences a fixed absolute threshold becomes meaningless once
+  !    the operands are large, because a single rounding of an operand of
+  !    magnitude x already produces a residual of order ulp(x) ~ 2.2e-16*x.
+  !    To avoid spuriously tripping error checks on rounding noise, the
+  !    comparisons below use max(absolute_target, rel_area_precision*|operand|).
+  !    rel_area_precision is set to relative machine precision for r8 so that
+  !    a difference below machine precision is never treated as a real error.
+  real(r8), parameter :: rel_area_precision = 1.0E-15_r8
+
   integer, parameter :: demotion_phase  = 1
   integer, parameter :: promotion_phase = 2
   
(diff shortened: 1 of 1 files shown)


biogeochem/EDCanopyStructureMod.F90
```diff
@@ -85,6 +85,18 @@ module EDCanopyStructureMod
...
+  !    Area conservation checks compare differences of two large, similarly
+  !    sized areas (e.g. cohort crown areas, layer areas, patch areas). For
+  !    such differences a fixed absolute threshold becomes meaningless once
+  !    the operands are large, because a single rounding of an operand of
+  !    magnitude x already produces a residual of order ulp(x) ~ 2.2e-16*x.
+  !    To avoid spuriously tripping error checks on rounding noise, the
+  !    comparisons below use max(absolute_target, rel_area_precision*|operand|).
+  !    rel_area_precision is set to relative machine precision for r8 so that
+  !    a difference below machine precision is never treated as a real error.
+  real(r8), parameter :: rel_area_precision = 1.0E-15_r8
```
rgknox, line 98: Since this is supposed to match the general precision of math on most compilers, maybe we should move this to FatesConstantsMod , give it a more general name, and use it in more places?

rgknox: Also, we should not be so strict, and expect machine precision exactly. I think we should drop this down at least an order of magnitude, if not 2...

cc @maritsandstad

biogeochem/EDCanopyStructureMod.F90
```diff
@@ -620,7 +634,8 @@ subroutine PromoteOrDemote(site,patch,target_layer,phase,target_area)
             call endrun(msg=errMsg(sourcefile, __LINE__))
 
          
-         elseif ( abs(layer_co(ic)%pd_area - cohort%c_area) < co_area_target_precision ) then
+         elseif ( abs(layer_co(ic)%pd_area - cohort%c_area) < &
```
rgknox, line 637: @glemieux noted this would be better as an "<=", I agree

biogeochem/EDCanopyStructureMod.F90
maritsandstad: Actually I did do it by putting it in FatesConstantsMod for later PRs, I agree it should move here also for this (see here: https://github.com/NorESMhub/fates/pull/74).

I'm not sure what you mean, do you want it to be lower as in 1.e-17 or 1.e-13. I would argue quite strongly against the former, asl that would allow you to let noise through, the point of that is to not do that. I would be fine with going to a larger number like 1e-13, but it is easiest to argue the machine precision numbers.

biogeochem/EDCanopyStructureMod.F90
maritsandstad: I mean,  for all practical purposes there is no difference between a < or <= comparison here. Chances that you are bang on to 15 digits are virtually nill

glemieux:
Closing in favor of merging via #1583 or similar direct pull request.

